PR #31, opened 2015-05-08
Add explicit Java 7 version requirement

hdietze:
Some maven clients assume Java 4 or 5 as default, this makes it Java 7 as default clear and prevents problematic compile errors.


jamesaoverton:
I dropped OpenJDK in commit ac64068, so this should work now...

PR #68, opened 2016-05-16
Added commands for relax and reduce

cmungall:


jamesaoverton:
I think the only problem is the integration test at the end. The test tool wants an `examples/reduced.owl` file to compare with the automatically generated `examples/results/reduced.owl` file. I can look at this more closely, but not today.

cmungall:
OK, there is a sample file in the test/resources, but I can come up with 
something that is more meaningful for a human

jamesaoverton:
Travis is failing on the integration test specified by your new example command in the `examples/README.md`. Run the integration tests with:

    mvn clean verify

The test tool looks through the README for Markdown code blocks that look like `robot` command-line calls, and executes them using a new ROBOT binary. Then it compares each `examples/results/X` file (not checked in) to the corresponding `examples/X` file (checked in).

So your `examples/results/nucleus.owl` should be moved to `examples/reduced.owl`. I tried that, but the example didn't match the output. I compared them using

    robot diff --left examples/reduced.owl --right examples/results/reduced.owl

but I didn't immediately see why the differences are important.

Stepping back: This system is definitely confusing. The idea is to ensure that the README examples *actually run*, but getting it to work is convoluted. Suggestions are welcome.

cmungall:
Got it working now. No burning desire to change the system now I know how it works.

jamesaoverton:
I messed with the formatting until CheckStyle was happy. Most of the changes are dumb, but it makes things consistent.

Can you please add a quick example of `relax` to the `examples/README.md`?

cmungall:
shall we just use a uniform example for reason, reduce and relax? edit.owl has a lot of extraneous guff

jamesaoverton:
Using the same example file as the source for several examples is a good idea. Yes, the edit.owl would be more informative if it was smaller (or just replaced by something smaller).

cmungall:
should all be done now. I keep edit.owl as it's a bit more informative for the annotate command and the convert command.

But the reason, relax and reduce commands are only relevant for logical axioms, so I use a more minimal ribosome.owl that mimics common situations in many ontologies.

PR is ready for merging if you think it looks ok

Issue #69, opened 2016-05-16
Add developer styleguide


cmungall:
I just did a PR with new code; just realized I used tabs not spaces...

jamesaoverton:
Yes. There are some things, but not yet all in one place:

- the pom.xml uses the maven-checkstyle-plugin, modified by checkstyle-suppressions.xml; you can run with `mvn site` and look in `[module]/target/site`
- the .editorconfig can give your editor hints, see http://EditorConfig.org

PR #71, opened 2016-05-20
Add note on code style, see #69

jamesaoverton:


jamesaoverton:
@cmungall: Is this sufficient? Anything to add?

Issue #59, opened 2016-02-18
Updgrade to owlapi v4


cmungall:
The main difference will be people who use robot to go from obo to owl, they will get the desired behavior here:

https://github.com/owlcs/owlapi/issues/317

jamesaoverton:
See #12, which was waiting for 4.1.0 and reasoner support. Might still require some API changes. I'll try updating to 4.1.4 when I get a chance.

cmungall:
@hdietze and I have some experience doing this

PR #72, opened 2016-05-20
READY TO MERGE Owlapi v4 - see #59

cmungall:


cmungall:
Looks like we may need to recreate some of the test outputs, may be non-semantic syntactic differences in output ordering

cmungall:
OK, I believe that the robot-command tests should now run. Just need to fix any of the core junit tests that do a diff on output

cmungall:
cannot figure this output from org.obolibrary.robot.DiffOperationTest

```
testCompareModified(org.obolibrary.robot.DiffOperationTest)  Time elapsed: 0.051 sec  <<< FAILURE!
org.junit.ComparisonFailure: expected:<...sources/simple.owl>)[)

2 axioms in Ontology 2 but not in Ontology 1:
+ AnnotationAssertion(rdfs:label <https://github.com/ontodev/robot/robot-core/src/test/resources/simple.owl#test1> "TEST #1"^^xsd:string)
+ OntologyID(OntologyIRI(<https://github.com/ontodev/robot/robot-core/src/test/resources/simple1.ow]l>))
> but was:<...sources/simple.owl>)[ VersionIRI(<null>))

2 axioms in Ontology 2 but not in Ontology 1:
+ AnnotationAssertion(rdfs:label <https://github.com/ontodev/robot/robot-core/src/test/resources/simple.owl#test1> "TEST #1"^^xsd:string)
+ OntologyID(OntologyIRI(<https://github.com/ontodev/robot/robot-core/src/test/resources/simple1.owl>) VersionIRI(<nul]l>))
>
 ```

jamesaoverton:
The `DiffOperationTest.java` is comparing diff output to a file `robot-core/src/test/resources/simple1.diff`. The new string version of the OntologyID axioms seems to include the VersionIRI now, but it didn't before. I think you can just change the `simple1.diff` until the test passes, like this:

    1 axioms in Ontology 1 but not in Ontology 2:
    - OntologyID(OntologyIRI(<https://github.com/ontodev/robot/robot-core/src/test/resources/simple.owl>) VersionIRI(<null>))

    2 axioms in Ontology 2 but not in Ontology 1:
    + AnnotationAssertion(rdfs:label <https://github.com/ontodev/robot/robot-core/src/test/resources/simple.owl#test1> "TEST #1"^^xsd:string)
    + OntologyID(OntologyIRI(<https://github.com/ontodev/robot/robot-core/src/test/resources/simple1.owl>) VersionIRI(<null>))

cmungall:
yep - @hdietze helped me figure that out

jamesaoverton:
I want to check this with a few of my workflows.

cmungall:
You may experience slight change in the output diffs due to different serialization order

You may want to pay particular attention to the cases where previous convenience methods on the OWLObject were removed from owlapi4, necessitating a different bit of code to do the same thing (e.g. EntitySearcher). In theory these should all be safe but it wouldn't hurt to add a few extra tests here.

cmungall:
I added some extract tests. The MIREOT one is a bit dumb, it uses a single class as upper and lower, you may want to improve this test

jamesaoverton:
I've tested this on a few things without any trouble.

PR #75, opened 2016-05-26
Fix copy AnnotationProperty superproperties

jamesaoverton:
@cmungall: I think this was accidentally broken in 95d5602f2aed3834adc5b1f9a6c0895ca98a8cd7. Now my AnnotationProperty hierarchies aren't coming out as expected when MIREOTing.


cmungall:
sorry about that

obvious checks to do: both have same semantics wrt

  - ontology import closure
  - reflexivity
  - direct vs subproperty closure

Issue #67, opened 2016-04-28
Add option to reason command to add axiom annotations to all added inferred axioms


cmungall:
GO pipeline depends on inferred axioms being tagged:

```
is_a: GO:0030155 {is_inferred="true"} ! regulation of cell adhesion
```

Add reason option `-a Annotation`

where Annotation is functional syntax for the annotation

alternatively `-a P V` with built in assumption about filler type

cmungall:
Consider extracting https://github.com/owlcollab/owltools/blob/master/OWLTools-Core/src/main/java/owltools/graph/AxiomAnnotationTools.java from OWLTools into its own library or robot

For now, we will only annotate subClassOf axioms

matentzn:
Was there a technical reason for this restriction? I think it would be great if it worked for all axiom generators; is this a limitation of the OWL API?

Issue #66, opened 2016-04-28
Add option to reason command to place inferences in separate ontology


cmungall:
Currently reasoning is done in place; new axioms are added to existing ontology (and redundant are removed if option is specified).

For GO and other ontologies we want inferences to be placed into a new ontology, that can be separately imported and dropped.

Add a new option `-n` that will place inferences in a new ontology. This ontology will become the new main ontology so that subsequent chain operations will operate on it.

Note: combining `-s` and `-n` should throw a warning.

cc @dosumis @hdietze

PR #81, opened 2016-06-09
Additional options for reason command to annotate inferences and place new inferences in a separate ontology

cmungall:
aded create-new-ontology option in ReasonerOperation for #66
TODO: Tests, command, #67


cmungall:
@jamesaoverton - any comments?

The use case here is the GO one where we want to place new inferences in a new file.

It somehow doesn't feel quite right to rewrite the input ontology, but as the operation doesn't return an ontology not sure how else to do that without a bigger refactor

jamesaoverton:
This is how I handled something similar for the `template` command: https://github.com/ontodev/robot/blob/master/docs/template.md#merging

It's not a great solution. Yours probably makes more sense for this command.

Can you set your editor to use four spaces instead of tabs, please?

cmungall:
Ready to merge pending approval from @jamesaoverton 

Note all tabs are gone

jamesaoverton:
Thanks. I'll take a closer look this afternoon and merge it unless I have big questions.

Issue #85, opened 2016-06-21
Fully specify intended behavior of remove-redundant-subclass-axioms option in reason command


cmungall:
## Background

A typical robot pipeline will compose two operations: `reason o reduce`

(sometimes interspersed with `reduce`)

The semantics of each independent operation is clearly specified. The "composition semantics" (ie running one operation on the output of another) is also simple and clearly defined.

Note that the `reason` command also provides a `--remove-redundant-subclass-axioms` option. Because this is interleaved with reason, and not a separate composed operation, it's more difficult to have simple semantics for it. We could simply avoid this and use compositions, but the option turns out to be useful for GO where we do hacks to support obo-edit, such as placing inferences in a new ontology.

## Current semantics

The current semantics of this option seems to be:

 * First perform reasoning and add direct inferred SubClassOf axioms
 * iterate through all classes in the root ontology for the reasoner
    * for each asserted superclass (which includes those added in the previous step):
       * test if this is a direct inferred superclass. If not remove the axiom, __unless it is annotated__

This is reasonably clear and should be formally documented.

However, the semantics may not always conform to what is expected. If the root ontology has annotated SubClassOf axioms, we will end up with asserted inferred 'shadows' of these.

Now things get more ad-hoc when we bring in other options. The recently added `-a` option, added for GO to mark inferred axioms, will automatically 'protect' every added inferred axiom, since we never remove annotated axioms, even if redundant. We can easily add some code that circumvents this, but we end up deviating from simplicity and clarity.

Furthermore, things get complex when we include the `-n` option (used for GO to place inferences in a new file). This compositioal semantics are not clear here.

Perhaps GO should be doing it's more abstruse operations outside Robot, and leave Robot for ontologies that do not have complex legacy systems to support.

cc @dosumis

PR #86, opened 2016-06-21
Ensure that the set of axioms used to test for addition of duplicates is the set present at the initial state

cmungall:
Relevant to #85

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/ReasonOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/ReasonOperation.java
@@ -173,6 +173,12 @@ public class ReasonOperation {
             }
         }
 
+        // cache the complete set of asserted axioms at the initial state.
+        // we will later use this if the -x option is passed, to avoid
+        // asserting inferred axioms that are duplicates of existing axioms
+        Set<OWLAxiom> existingAxioms = ontology.getAxioms(Imports.INCLUDED);
+
+
         // Make sure to add the axiom generators in this way!!!
         List<InferredAxiomGenerator<? extends OWLAxiom>> gens =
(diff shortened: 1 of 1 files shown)


cmungall:
Prior to this PR we had the issue that the go new inferences file (generated using the `-n` option, added in #81) included axioms that were trivially redundant with axioms in the main ontology. E.g. 

```
[Term]
id: GO:0030054 ! cell junction
is_a: GO:0005575 {is_inferred="true"} ! cellular_component
```

Note that this is both a true axiom, and a *direct* inferred axiom. While we were specifying `-s` to remove redundant axioms, this was not removed since by the current definition of removal of redundant axioms (see #85) this was not removed, as it was direct (no intermediate superclasses).

The `-x` option was intended to handle this case. It prohibits the addition of duplicate axioms during the reason operation. However, it was not working as intended in combination with `-n`: in the above case, the `SubClassOf(cell-junction cellular-component)` axiom is not formally a duplicate, since the new ontology is created as a standalone ontology, intended to be imported.

This PR fixes the situation by changing the semantics of `-x` to be do not add anything that duplicates *any axiom present at the initial state, prior to `-n`*.

jamesaoverton:
This seems straightforward. Please merge it when you're ready.

dosumis:
> do not add anything that duplicates any axiom present at the initial state, prior to -n.

This sounds like the correct behaviour to me.  Spot checking test file of GO inferences.

dosumis:
Spot checking looks good.

Number of tagged inferred GO annotations back to similar numbers to those before change:

* last tuesday's GO file had 88781
* using latest GO + test GO inferred annotations file => 88638.  Expected slightly less due to deletion of stale axioms, but probably OK.

Issue #84, opened 2016-06-16
Documentation of merge behavior


dosumis:
Where can I find documentation of default merge behaviour in the presence of imports statements?  In tests it seems to automatically merge down the import chain.  This can be useful behaviour in some situations, but I don't think it should be the default.

cmungall:
ideally here:

```
$ robot merge -h
usage: robot filter --input <file> --input <file> --output <file>
 -h,--help              print usage information
 -i,--input <arg>       merge ontology from a file
 -I,--input-iri <arg>   merge ontology from an IRI
 -noprefixes            do not use default prefixes
 -o,--output <arg>      save merged ontology to a file
 -p,--prefix <arg>      add a prefix 'foo: http://bar'
 -P,--prefixes <arg>    use prefixes from JSON-LD file
 -v,--version           print version information
```

but this lacks the docs. It also says "filter"...

To avoid docs getting out of sync, how about:

 1. minimal docs on command line, most auto-derived
 2. formal specification in javadocs for the <FOO>Operation.java object
 3. A link from 1 to HTML generated from 2

Alternatively, a markdown or md/wiki page for each operation?

jamesaoverton:
Yes, the merge command merges all imports. Anything else will get complicated fast.

The "filter" mistake is easy to fix.

There's also the examples/README.md documentation, which is automatically tested.

dosumis:
> Yes, the merge command merges all imports. Anything else will get complicated fast.

I think we need to deal with the complexity.

I was trying to merge in some auto-generated files to GO.  It is essential to preserve GO import structure when doing this. OWLtools allows this, so I've gone back to using it.  Merging down import chains is a separate command. 

Keeping these steps separate is also essential to the release pipelines on VFB and FlyBase ontologies.

cmungall:
Attempting to specify current behavior. It's important to distinguish between Operation and Command. There are actually two methods in MergeOperation it's important to distinguish between.

## merge ontologies into target (mergeInto method)

https://github.com/ontodev/robot/blob/6049ab427cc2455cb9c4d14281a254310be9e5b3/robot-core/src/main/java/org/obolibrary/robot/MergeOperation.java#L106-L133

This does not touch the import closure of targetOntology. It does not merge in its imports closure, nor does it remove the declarations themselves.

It does copy the axioms from the full imports closures of the list of other ontologies into target.

[...]

PR #92, opened 2016-07-26
DO NOT MERGE YET changes for merge behavior

cmungall:
See https://github.com/ontodev/robot/issues/84#issuecomment-233156319


jamesaoverton:
@cmungall Is anything blocking this?

Arguably it should be the default, but just having the option would solve a bunch of problems.

Issue #98, opened 2016-09-13
Robot silently loses axioms for some ontologies


cmungall:
test.owl:

```
<?xml version="1.0"?>
<rdf:RDF xmlns="http://purl.obolibrary.org/obo/ro.owl#"
     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
     xmlns:owl="http://www.w3.org/2002/07/owl#"
     xmlns:xml="http://www.w3.org/XML/1998/namespace"
     xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#">
    <owl:Ontology rdf:about="http://purl.obolibrary.org/obo/ro.owl"></owl:Ontology>

    <!-- dodgy URI -->
    <owl:AnnotationProperty rdf:about="http://www.geneontology.org/formats/oboInOwl#http://www.w3.org/2000/01/rdf-schema#seeAlso"/>

    <owl:ObjectProperty rdf:about="http://purl.obolibrary.org/obo/BFO_0000050">
        <rdfs:label rdf:datatype="http://www.w3.org/2001/XMLSchema#string">part of</rdfs:label>
    </owl:ObjectProperty>
</rdf:RDF>
```

```
robot convert -i test.owl -o z.owl && cat z.owl
WARN  http://www.oracle.com/xml/jaxp/properties/entityExpansionLimit not supported by parser type org.apache.xerces.jaxp.SAXParserImpl
WARN  entityExpansionLimit not supported by parser type org.apache.xerces.jaxp.SAXParserImpl
INFO  Notice: root element does not have an xml:base. Relative IRIs will be resolved against file:/Users/cjm/repos/porifera-ontology/src/ontology/test.owl
WARN  http://www.oracle.com/xml/jaxp/properties/entityExpansionLimit not supported by parser type org.apache.xerces.jaxp.SAXParserImpl
WARN  entityExpansionLimit not supported by parser type org.apache.xerces.jaxp.SAXParserImpl
<?xml version="1.0"?>
<rdf:RDF xmlns="http://purl.obolibrary.org/obo/ro.owl#"
     xml:base="http://purl.obolibrary.org/obo/ro.owl"
     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
     xmlns:owl="http://www.w3.org/2002/07/owl#"
[...]

PR #99, opened 2016-10-07
Added production of jar with dependencies using OneJar plugin.

balhoff:
Fix for #98. This puts the combined jar in `robot-command/target/robot-command-0.0.1-SNAPSHOT.one-jar.jar`. @jamesaoverton I just kind of muddle through with Maven. Do we want this to be copied to `bin/robot.jar`? If so, I am not sure the preferred way to configure that.

### DIFF
--- a/robot-command/pom.xml
+++ b/robot-command/pom.xml
@@ -1,74 +1,63 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    <modelVersion>4.0.0</modelVersion>
-    <parent>
-        <groupId>org.obolibrary</groupId>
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Thanks! I can't look at it today. I'm not sure why that integration test failed.

Can you use spaces instead of tabs, please? We're trying for a consistent style: https://github.com/ontodev/robot#development

balhoff:
Sorry about the tabs! I didn't realize the Eclipse indent would do that. I updated the PR. I'll try to figure out why Travis is failing. `mvn test` succeeds on my machine.

jamesaoverton:
I can reproduce the problem on my machine. The problem is with `mvn verify` which runs `CommandLineIT.java`. It looks at the `examples/README.md`, extracts example commands, and tries to run them as command-line invocations. In order to do this, it builds `bin/robot.jar` and calls the `bin/robot` script.

So I think the problem is that one-jar is not building the expected `bin/robot.jar`. Can you reconfigure one-jar to build that artifact?

I like having automated tests for the documentation. Maybe this system is too clever.

balhoff:
@jamesaoverton I think this is all fixed now.

PR #102, opened 2016-10-20
Adding materialize command

### LINKED ISSUE #101: Add a materialize command to material inferred superclasses (named and existential)
## Motivation

Consider the following class, together with its axiom in a source/editors file

```
[Term]
id: GO:0002080
name: acrosomal membrane
intersection_of: GO:0098805 ! whole membrane
intersection_of: bounding_layer_of GO:0001669 ! acrosomal vesicle
```

The `reason` command will produce an extra subClassOf axiom:

```
id: GO:0002080
name: acrosomal membrane
is_a: GO:0030667 ! secretory granule membrane
intersection_of: GO:0098805 ! whole membrane
intersection_of: bounding_layer_of GO:0001669 ! acrosomal vesicle
```

(assuming additional axioms, e.g. acrosomal vesicle is a secretory vesicle, and axiomatization of SGM)

This is valid, but is not sufficient for the case where the consumer of the ontology wishes to traverse existential graphs over properties such as part-of

The `relax` command (#7) partially addresses this, in that it will create an additional:

```
relationship: bounding_layer_of GO:0001669 ! acrosomal vesicle
```

However, if we then wish to filter object properties this loses information
## Materialize

This will use the [EMR](https://github.com/owlcollab/expression-materializing-reasoner/) to materialize direct superclasses, named and existential. The set of existentials can be restricted by property.

For example, it should be possible to do compose the materialize command with `reduce`:

```
    robot materialize --reasoner ELK --input emr_example.obo --term BFO:0000050 filter -t BFO:0000050 reduce --output results/emr_reduced.obo
```

This should write all direct inferred axioms `C SubClassOf part-of some D` (as well as C SubClassOf D), where D is a named class.

The resulting SubClassOf graph is then guaranteed to be complete for part-of

In the example above, we would have:

```
id: GO:0002080
name: acrosomal membrane
namespace: cellular_component
def: "The membrane that surrounds the acrosomal lumen. The acrosome is a special type of lysosome in the head of a spermatozoon that contains acid hydrolases and is concerned with the breakdown of the outer membrane of the ovum during fertilization." [GOC:dph]
is_a: GO:0030667 ! secretory granule membrane
relationship: part_of GO:0001669 ! acrosomal vesicle
```

cc @dosumis

cmungall:
Fixes #101

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: java
 jdk:
   - oraclejdk8
-  - oraclejdk7
-  - openjdk7
 script: "mvn verify"
--- a/examples/README.md
+++ b/examples/README.md
@@ -109,6 +109,18 @@ Robot can be used to 'reduce' (i.e. remove redundant subClassOf axioms), indepen
 
     robot reduce --reasoner ELK --input ribosome.owl --output results/reduced.owl
 
+## Materialization
+
+Robot can materialize all parent superclass and superclass expressions using the expression materializing reasoner, which wraps an existing OWL Reasoner
+
+    robot materialize --reasoner ELK --input emr_example.obo --term obo:BFO_0000050  --output results/emr_output.obo
+
+This operation is similar to the reason command, but will also assert parents of the form `P some D`, for all P in the set passed in via `--property`
+
+This can be combined with filter and reduce to create an ontology subset that is complete
+
+    robot materialize --reasoner ELK --input emr_example.obo --term BFO:0000050 filter -t BFO:0000050 reduce --output results/emr_reduced.obo
+
 ## Annotating
 
 It's important to add metadata to an ontology before releasing it, and to update the ontology version IRI.
--- a/examples/emr_example.obo
+++ b/examples/emr_example.obo
@@ -1,0 +1,69 @@
+ontology: emr_test
+
+[Term]
+id: GO:0030667
+name: secretory granule membrane
+namespace: cellular_component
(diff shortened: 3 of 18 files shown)


cmungall:
@dosumis - makes sense?

Note that a nice side effect is that this works as expected for GCIs

jamesaoverton:
Why drop Java 7 support in this PR?

cmungall:
Dependency on the ExpressionMaterializingReasoner - https://github.com/owlcollab/expression-materializing-reasoner/ - but maybe there is some java-fu I'm missing here, may be possible to cross-compile?

jamesaoverton:
If we're dropping Java 7, then that should be a separate PR.

I'm a bit worried about the consequences. Travis doesn't seem to support OpenJDK 8. So dropping support for OpenJDK 7 means only testing OracleJDK. I'd like people to be able to use ROBOT with OpenJDK.

The feature is fine. `mvn clean site` shows some style problems. I can fix them if you don't have time.

cmungall:
I'm not sure if the lack of travis support for openjdk8 prevents anyone from using openjdk8.

I could look at backporting EMR to be java7, but we are java8 for everything now, my vote would be to just make the jump

jamesaoverton:
Is this good to go?

cmungall:
Yes. If you're happy with the java8 switch.

jamesaoverton:
Ok. I updated Java8 in #103. I'll merge this now.

PR #107, opened 2016-11-02
Fix logging

jamesaoverton:
- make logging configurable with CLI flags
- make default level ERROR


jamesaoverton:
I meant this to be separate from #106 -- oops. Well, they're both about fixing logging at the command line.

Issue #110, opened 2016-11-20
New operation/command: unmerge


cmungall:
Sometimes it's useful to temporarily add axioms to an ontology; e.g. `merge o ... o unmerge`

An unmerge command would subtract commands from a previously merged ontology.

More generally, this could be conceived of as a simple subtract operation; should we call it subtract or unmerge?

jamesaoverton:
So given ontologies X and Y, if you merge X and Y then unmerge Y, the result has all the axioms of X that are not axioms of Y?

I think that "subtract" is a better name for the operation, although "unmerge" emphasizes the duality with "merge".

cmungall:
> So given ontologies X and Y, if you merge X and Y then unmerge Y, the result has all the axioms of X that are not axioms of Y?

Only if there is no overlap between the axioms. Currently it's `X-Y`

> I think that "subtract" is a better name for the operation, although "unmerge" emphasizes the duality with "merge".

Subtract is the better name for what is currently implemented (`X-Y`)

But the idea is we could add options for scenarios that are more like an unmerge. For example

 - anything that has been MIREOTed in (an 'unmireot') or is somehow tagged
 - an option to switch behavior to `X-(Y-(X^Y))` (ie remove everything from X that is uniquely in Y)

But arguably 'subtract' is also fine for these too?

jamesaoverton:
Ok. If we do a better job indicating where axioms come from, then that could work.

cmungall:
So shall we go with unmerge as the name then?

PR #111, opened 2016-11-20
Adding unmerge command

cmungall:
See #110


robot-core/src/main/resources/log4j.properties
```diff
@@ -3,4 +3,4 @@ log4j.appender.console.layout=org.apache.log4j.PatternLayout
 log4j.appender.console.layout.ConversionPattern=%-5p %m%n
 #log4j.appender.console.layout.ConversionPattern=%d [%t] %-5p %c - %m%n
 log4j.logger.user=DEBUG
-log4j.rootLogger=ERROR, console
+log4j.rootLogger=INFO, console
```
jamesaoverton: Please leave the default logging level at ERROR. In #107 I added verbose flags, if you need them.

jamesaoverton:
Should there be an entry in `examples/README.md` for this? Maybe later?

cmungall:
we have this in the README:

## Unmerging

A merge can be 'undone' using the unmerge command. This can be useful if we want to perform some operation (for example, reasoning) on a merged file, but later subtract the results.

This will merge in foo, and then subtract out foo from the merged ontology.

    robot merge --input edit.owl --input foo.owl unmerge --input foo.owl --output results/unmerged.owl

Issue #114, opened 2016-11-24
Is it possible for robot to enact entity substitution when outputting ontology?


Public-Health-Bioinformatics:
I have the robot working on my GenEpiO ontology, which has a number of XML entities defined up front, like: xmlns:obo="http://purl.obolibrary.org/obo/" .  I was wondering if, as an output file space saving measure, there was a way for robot to implement these entities wherever they exist? e.g.     
   
```
<owl:Class rdf:about="http://purl.obolibrary.org/obo/ERO_0001965">
   =>     
   <owl:Class rdf:about="&obo;ERO_0001965">
```
I think the entities get replaced with expanded URL in the original genepio-edit.owl file whenever Protege saves it.  (If this is beyond robot's purview, then I will use an old-fashioned search & replace approach for the job.)

jamesaoverton:
ROBOT does not currently have an option for output using DOCTYPE and XML entities. It looks like OWLAPI exposes this via [`setUseNamespaceEntities`](http://owlcs.github.io/owlapi/apidocs_4_2_3/org/semanticweb/owlapi/rdf/rdfxml/renderer/XMLWriterPreferences.html) on `XMLWriterPreferences`.

This would be a good feature to add to ROBOT's `IOHelper`. I'm very short on time right now. How are your Java skills? :^)

Public-Health-Bioinformatics:
I will see if I can give it a shot in the next few weeks but not holding breath (java for me is usually an archaeological experience).  Thanks for pointing out the solution flag directly!

d.

jamesaoverton:
@cmungall: Do you think it's better to use entities by default when writing XML? I'm leaning toward "No": more complex, less predictable, harder to process with other tools (e.g. `grep`), could break text-base diffs for no good reason.

It's nice to have the option, though!

Public-Health-Bioinformatics:
Righto, I left the entities option disabled by default so one must deliberately flip it on.

Public-Health-Bioinformatics:
I'm just about finished the test script except for one problem - I don't quite know how to properly reference creation of a temporary file for the ontology output (unless there's a simpler streaming solution).  Right now whatever I put for the ontologyOutputPath results in a "java.lang.NullPointerException: stream cannot be null" error in the saveOntology call.  Tips?

```
    public void testOntologyUseXMLEntities() throws IOException {
        String ontologyInputPath = "/mireot.owl";
        String ontologyOutputPath = "/mireot_xmlentities_test.owl";
[...]

PR #115, opened 2016-11-25
adding -e --entities flag

Public-Health-Bioinformatics:
Enables saving of XML ontology formats to encode the ontology’s
entities in the output.

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java
@@ -251,7 +251,7 @@ public class CommandLineHelper {
 
     /**
      * Given a command line, return an initialized IOHelper.
-     * The --prefix, --prefixes, --noprefixes and --entities options are handled.
+     * The --prefix, --prefixes, --noprefixes and --xml-entities options are handled.
      *
      * @param line the command line to use
      * @return an initialized IOHelper
@@ -269,7 +269,7 @@ public class CommandLineHelper {
             ioHelper.addPrefix(prefix);
         }
 
-        ioHelper.setEntityFlag(line.hasOption("entities"));
+        ioHelper.setXMLEntityFlag(line.hasOption("xml-entities"));
 
         return ioHelper;
     }
@@ -540,7 +540,7 @@ public class CommandLineHelper {
         o.addOption("p", "prefix",  true,  "add a prefix 'foo: http://bar'");
         o.addOption("P", "prefixes", true, "use prefixes from JSON-LD file");
         o.addOption("noprefixes", false, "do not use default prefixes");
-        o.addOption("e", "entities", false, "use entity substitution when saving XML ontology");
+        o.addOption("x", "xml-entities", false, "use entity substitution with ontology XML output");
         return o;
     }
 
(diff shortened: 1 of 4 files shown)


jamesaoverton:
Much appreciated! (Actually, it's hard to express just how much we appreciate PRs.)

I have two requests:

1. Can you please add a test, to help us avoid accidentally breaking this code. My suggestion would be a round-trip test in `IOHelperTest.java`: load one of the fancier test files (maybe `mireot.owl`), save as XML with entities, load that XML with entities, then diff against the original to make sure that they are the same.

2. I'd like to change some of the names. I'm worried about confusion between XML entity, `OWLEntity`, and BFO `entity`, and I'd prefer to leave the short `-e` option available for use by other commands. I'll make some line-comments about this.

(See #114)

robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java
```diff
@@ -538,6 +540,7 @@ public static Options getCommonOptions() {
         o.addOption("p", "prefix",  true,  "add a prefix 'foo: http://bar'");
         o.addOption("P", "prefixes", true, "use prefixes from JSON-LD file");
         o.addOption("noprefixes", false, "do not use default prefixes");
+        o.addOption("e", "entities", false, "use entity substitution when saving XML ontology");
```
jamesaoverton: I'd prefer to use a single long option for this: `--xml-entities`

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
```diff
@@ -683,6 +689,29 @@ public void setContext(Map<String, Object> map) {
     }
 
     /**
+     * Set whether or not XML entities will be swapped into URI's in saveOntology XML output formats.
+     *
+     * @param map a map of strings for the new JSON-LD context
+     */
+    public void setEntityFlag(Boolean entityFlag) {
```
jamesaoverton: Please rename to `setXMLEntityFlag`

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
```diff
@@ -683,6 +689,29 @@ public void setContext(Map<String, Object> map) {
...
+            logger.warn(e.getMessage());
+        }
+    }
+
+    /**
+     * Get the useXMLEntities flag.
+     *
+     * @return boolean useXMLEntities flag
+     */
+    public Boolean getEntityFlag() {
```
jamesaoverton: Please rename to `getXMLEntityFlag`

robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java
Public-Health-Bioinformatics: Good point.  Will implement all this shortly.

Public-Health-Bioinformatics:
Inside pull request I accidentally deleted an examples/uberon_mireot.owl file.  Its restored, and all should be working and tested.

jamesaoverton:
Looks good. Your tempFile solution is a good idea.

The only problem is that we should not include the binary files in our source tree. I'm not sure how to remove those from the pull request's history. So I'm going to start a new branch with your source changes but without the binary files, and merge that.

jamesaoverton:
Thanks again! This is a great option to have, and I'm really glad that you implemented it.

I merged these changes in 8983128670790a61d734eb32f289ab178d9ebc24, and in f54dc017c790a6de41ecaf91662135d848cb183b I updated the `.gitignore` to stop the binary files from being accidentally committed in the future.

Public-Health-Bioinformatics:
Great.  It was a little bit of a java / owllib learning experience, glad it proved fruitful in the end.  I much prefer reading .owl with entities in place.

PR #118, opened 2016-11-30
Added validate-profile command.

balhoff:
I added a command to validate an ontology against an OWL 2 profile. For example:

```
$ ./bin/robot validate-profile -profile EL --input pato.owl 
OWL 2 EL Profile Report: [Ontology and imports closure in profile]

```


jamesaoverton:
Very nice! I've never validated a profile before, but this seems to work as described. I added an example and integration test.

Issue #119, opened 2016-12-01
Add option to reason locally, to prevent asserting entailed axioms in the wrong ontology


cmungall:
Scenario:

 - Ontology O1 imports O2
 - O2 has not been reasoned, and entails axiom `X SubClassOf Y`
 - When reason is executed for O1, `X SubClassOf Y` is asserted in O1

This is often undesirable, as the axiom is inserted in the 'wrong' ontology here. Formalization of what is intuitively the 'wrong' ontology here is not straightforward. An example can be found where axioms connecting two GO classes are injected into the uberon releases: https://github.com/geneontology/go-ontology/issues/12843

## Simple approach

Here we only assert superclasses for `X` if `X` is in the signature of the base ontology (i.e. `Imports.EXCLUDED` in the OWLAPI)

However, this is imperfect as a class can be in the signature of multiple ontologies. For example, if O1=uberon and O2=go, and uberon uses a GO class to define an uberon class (e.g. via capable_of) then we have the potential of GO entailments being asserted in the Uberon release, even if `Imports.EXCLUDED` is used.

## Complete approach

A more complete approach would be to first iterate over the (proper, transitive) imports closure of `O1`, finding entailments. The set of entailed axioms would form a blacklist when as a final step we reason over `O2`

cmungall:
As an example of this phenomena, try

```
robot reason -i import-non-reasoned.owl -o /tmp/foo.owl
```

in robot-core/src/test/resources

cmungall:
Fixed by #120

cmungall:
## Alternate Approach

only assert superclasses for any class for which `O` asserts any axiom about.

rctauber:
Is this fixed by #131?

rctauber:
I believe this is resolved by #131. I'm going to close it, but please re-open if it needs more work.

PR #120, opened 2016-12-01
Materialize in local ontology only

cmungall:
Changes behavior of materialize in the presence of imports closure

First iterate through all imported ontologies (non-reflexive, transitive), performs materialization.

Then when materialization is performed on the main ontology, ensure that no axioms that exist (in full imports closure) are asserted.

See #119


jamesaoverton:
Is `robot-core/src/test/resources/catalog-v001.xml` needed?

I haven't tested this. You should merge when you're ready.

cmungall:
> Is robot-core/src/test/resources/catalog-v001.xml needed?

No way to test imports without it!

PR #138, opened 2017-01-08
Adding a obo graph json export capability

### LINKED ISSUE #94: Add an obograph-json export option
OWLAPI-level access to an ontology is generally not useful or appropriate for bioinformatics developers. To wean people off obo format we have created a json format that is graph-oriented

See:
https://github.com/geneontology/obographs

On mvncentral so easy to include. 

Note the owlapi isn't aware of this, so we'd either need a special purpose `convert` command, or an annoying switch in the code that checks the suffix and routes to this to the FromOwl converter

cmungall:
Fixes #94 

This will fail until obographs 0.0.8 percolates to mvncentral in ~24hrs...

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/ConvertCommand.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/ConvertCommand.java
@@ -104,6 +104,7 @@ public class ConvertCommand implements Command {
      * <li>OWLXML .owx
      * <li>Manchester .omn
      * <li>OWL Functional .ofn
+     * <li>OboGraphs JSON .json
      * </ul>
      *
      * @param state the state from the previous command, or null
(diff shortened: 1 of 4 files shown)


cmungall:
0.0.8 is on mvncentral, so this is ready to merge if it looks OK

jamesaoverton:
Sure, merge it.

My only concern is that a file ending in `.json` could have many flavours. The current ROBOT design is limited to relying on the file extension alone to determine format. If it becomes a real problem, we could add another output option that accepts two arguments: format name and file name.

cmungall:
An explicit format option would be good.

This is a common paradigm for me in makefiles:

```
foo2.owl: foo.owl
    robot OPERATION -i $< -o $@.tmp && mv $@.tmp $@
```

easy to get around like this, but somewhat awkward:

```
foo2.owl: foo.owl
    robot OPERATION -i $< -o $@.tmp.owl && mv $@.tmp.owl $@
```

PR #148, opened 2017-02-22
Initial dockerfile with "robot" entry point

joshmoore:
This dockerfile builds the robot jar and installs
it under /usr/local/bin along with the "robot"
script which is used as the entrypoint. Once built,
"docker run --rm robot ..." is equivalent to running
the command-line.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,8 @@
-FROM maven:3-jdk-8
+FROM maven:3-jdk-8-alpine
 
-RUN adduser robot
+RUN adduser -D robot
 RUN mkdir -p /usr/src/app
 RUN chown robot /usr/src/app
-
 
 COPY pom.xml /usr/src/app
 COPY robot-command /usr/src/app/robot-command
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM maven:3-jdk-8
+
+RUN adduser robot
+RUN mkdir -p /usr/src/app
+RUN chown robot /usr/src/app
+
+
+COPY pom.xml /usr/src/app
+COPY robot-command /usr/src/app/robot-command
+COPY robot-core /usr/src/app/robot-core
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Thanks for the PR. This worked for me, and I think it's a good idea.

My main concern is that `maven:3-jdk-8` is pretty big. I tried using `maven:3-jdk-8-alpine` but got an error `passwd: password for robot is unchanged`. Are there strong reasons to use the larger Debian one rather than the smaller Alpine one? (I don't really keep up with Dockers stuff.)

cmungall:
this looks really useful, thx

cc @kltm

joshmoore:
```
REPOSITORY                               TAG                       IMAGE ID            CREATED             SIZE
test-robot                               latest                    4458d93d511a        14 hours ago        776.9 MB
test-robot-alpine                        latest                    484aa97fe8c2        14 seconds ago      284.9 MB
```

Going smaller would likely require using the openjdk base image and downloading maven, building robot, and removing maven all within a single `RUN` step.

jamesaoverton:
I merged this and updated the README.md.

Much appreciated!

cmungall:
Should we be releasing to dockerhub?

Does this change now we have robot on mvnventral? Be good to sync.

I have also made a Dockerfile and released an image that collects robot and other ontology tools
https://github.com/INCATools/ontology-starter-kit/blob/docker/docker/Dockerfile

For now this pulls the jar from a jenkins job which is weird.. 

I'm not sure what best practice is for combing different command line scripts together. Perhaps https://docs.docker.com/develop/develop-images/multistage-build/

jamesaoverton:
While it's good to have this example of a Dockerfile for ROBOT alone, I see a lot more benefit from the image that you're building with aa variety of tools. I'd be happy if we just publish that, without worrying about publishing an image just for ROBOT.

PR #164, opened 2017-05-02
added a --construct option to query, to allow for CONSTRUCT type sparql queries

dougli1sqrd:
As an example to test, you can use the SPARQL:

    prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
    PREFIX part_of: <http://purl.obolibrary.org/obo/BFO_0000050>
    CONSTRUCT {
        ?part part_of: ?whole
    }
    WHERE {
        ?part rdfs:subClassOf [	rdf:type owl:Restriction ;
							owl:onProperty part_of: ;
							owl:someValuesFrom ?whole ]
    }

And use the command:
`bin/robot query -i examples/edit.owl --construct query.rq output -f ttl`

This produces a ttl file called output with direct "part of" assertions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 .classpath
 .project
 .settings
+.idea
+*.iml
 bin/original-robot.jar
 bin/owltools2.jar
 bin/robot.jar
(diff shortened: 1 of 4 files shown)


cmungall:
Can you add a test to QueryOperationTest

dougli1sqrd:
Will do

dougli1sqrd:
Simple test added.

PR #168, opened 2017-05-10
made reason command -e take one of 3 options for Chris' new equivalen…

dougli1sqrd:
…ce class detection method. Tests to go with it


dougli1sqrd:
I actually allowed false|true to still map to None or All so anyone using the older options are still okay

dougli1sqrd:
Should I merge this?

Issue #177, opened 2017-08-08
Add verify subcommand for SPARQL queries


cmungall:
See #159 for full details.

Use case: https://github.com/geneontology/go-ontology/pull/14033

we have an ugly perl wrapper over query to do this now, and it's not very informative:

```

../util/check-violations-sparql-output.pl reports/equivalent-classes-violation.csv reports/trailing-whitespace-violation.csv reports/owldef-self-reference-violation.csv
 // in reports/trailing-whitespace-violation.csv
Violations: 1
```

dougli1sqrd:
Currently the sparql queries run and the results are output in CSV in the reports dir. Then the perl script reads them. It doesn't seem like there is a limit on the number of results in terms of violations that can occur, is that true? So if there's millions of violations we end up with a huge report? Should we limit that in the sparql query side, or in the Robot end?

cmungall:
Let's default to 100, and have an option to add more

jamesaoverton:
Closed by #179

Issue #159, opened 2017-04-13
Add ability to do SPARQL construct


cmungall:
Following up from
https://github.com/ontodev/robot/issues/24#issuecomment-225994582

not sure if this should be new command or a `--construct` option on query

jamesaoverton:
My intention was to provide `--select` (implemented), `--ask`, `--verify`, and `--construct` options on query, because the semantics are very different. `--verify` would be like `--ask` but would exit on failure.

But I'm willing to be convinced of a better design.

cmungall:
sounds good to me

dougli1sqrd:
I'm a little confused. Why are there command line options for different sparql constructs? Can you not merely run the query, and let the query (which might have `CONSTRUCT`, `SELECT`, etc) results come back? A separation like between just querying and verifying could make sense, cause then you're drawing a line between actual ROBOT behavior: in the first case we just run the query, and the second we check if the query results are correct.

cmungall:
I'm OK with introspecting the behavior, but the main drawback is it 
leads to possible unpredictability. If the options are explicit there is 
less chance of unexpected behavior. What about both? If nothing 
specified, introspect. If option passed, use it. What do you think 
James?

balhoff:
I know you mean the other James, but I'll just make a comment. :-) Most of the RDF APIs I've used have separate methods for executing SELECT vs CONSTRUCT vs ASK, etc., I think primarily because the return types are completely different: result iterator vs graph vs boolean. That's probably more important in a type-safe programming language compared to a command line that is text in and text out. Not sure if that sways the decision either direction.

jamesaoverton:
Yes, just like  @balhoff said, Jena uses different methods and I was following their lead. Also, I only needed SELECT, so that's all I implemented.

I don't mind introspecting, but I haven't looked for a library method to do that.

It would be great to see some examples of the CLI calls everyone wants to make.

dougli1sqrd:
[...]

PR #179, opened 2017-08-10
Verify command

dougli1sqrd:
for https://github.com/ontodev/robot/issues/177

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -142,7 +142,9 @@ directory passed to `-O` must exist.
 
 For example:
 
-    robot verify --input asserted-equiv.owl --queries equivalent.sparql -O results/
+```
+robot verify --input asserted-equiv.owl --queries equivalent.sparql -O results/
+```
     
 Should output as a response: 
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -93,6 +93,68 @@ For more details see:
 - [SLME](http://owlapi.sourceforge.net/javadoc/uk/ac/manchester/cs/owlapi/modularity/SyntacticLocalityModuleExtractor.html)
 - [ModuleType](http://owlapi.sourceforge.net/javadoc/uk/ac/manchester/cs/owlapi/modularity/ModuleType.html)
 
+## Querying
+
+Robot can be used to submit [sparql](https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#bind) 
+queries against an ontology. Robot currently has two commands `query` and `verify` that use sparql to
+perform queries on ontology sets.
+
+### `query`
+
+The `query` command comes in two flavors, `select` and `construct`, which are passed in as options:
+`robot query --select` or `robot query --construct`. `select` corresponds to a sparql 
+_[SELECT](https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#select)_ query. This is used to 
+extract data matching a pattern. 
+
+    robot query --input nucleus.owl --select subclass_of_0044464.sparql output
+
+Resulting in:
+    
+    subs,names
+    http://purl.obolibrary.org/obo/GO_0044424,intracellular part
+    http://purl.obolibrary.org/obo/GO_0005622,intracellular
+    
+The `construct` option allows one to build triples based on a pattern using a 
+[CONSTRUCT](https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#construct) query. This will output
+triples in the format specified with the `--format` option, and should probably be set to "ttl" for this
+query mode. For example:
+
+    robot query --format ttl --input nucleus.owl --construct construct_some.sparql output
+    
+Has many results, but a portion could look like:
+
+    <http://purl.obolibrary.org/obo/GO_0044464>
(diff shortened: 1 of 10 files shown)


cmungall:
Logic looks fine once test is added...

...but I want to check about the decision to implement as a separate command rather than an option - we went through this discussion when implementing `construct`:

https://github.com/ontodev/robot/issues/159#issuecomment-294157848

I thought we decided there that construct and verify should be options for `query`?

Sorry I may have mislead by using the term 'subcommand' in the ticket.

What do you think @jamesaoverton

dougli1sqrd:
Ah I see. It's kind of nice this in it's own command, but it wouldn't be hard to move to query either.

jamesaoverton:
I had been thinking of it being part of `query`, but my reasons are mostly for easier implementation. From the user's perspective, this distinct `verify` subcommand seems to be better. The exit code behaviour is very different from `query`, for one thing.

If the output target is a directory, then I think `--report-dir` is a better name than `--report`.

jamesaoverton:
An advantage to putting verify inside `query` is that it can be slow and expensive to populate the triplestore, and currently we aren't putting the triplestore in the CommandState to pass to the next command in the chain. How likely are we to select then verify, or verify then select, and/or do we want to add the triplestore to the CommandState?

Thinking ahead, if we're constructing, then we probably want to do more work with the new triples, so we probably want to pass them down the chain. But then we have to worry about passing the triplestore or repacking it as an OWLOntology.

dougli1sqrd:
I agree that `report-dir` is probably better, I can change that. With verify as a command, I could pass forward the command state instead of ignoring it like I am now. Should the `--input, -i` parameter _replace_ the existing ontology in the `CommandState` or should it update somehow? 

And `construct` I guess doesn't put anything in the command state? We could pass it in to the command state.

jamesaoverton:
The thing is that `CommandState` only includes an `OWLOntology`, not the `DatasetGraph` for the triple store. You're currently just returning the `inputState` unchanged, which is fine because this command should not make any changes to the ontology. If the next command wants to do some SPARQL thing, it will have to build a new `DatasetGraph`, which is relatively expensive. On the other hand, throwing out the `DatasetGraph` saves memory. On the third hand, a construct operation would change the triples, then would we want to pass a new `OWLOntology` or the modified `DatasetGraph` to the next command?

The proposed implementation of verify is probably what we want. I think you should go ahead.

I just worry (like I always do) about regretting something later. The worrying never helps. :^)

dougli1sqrd:
Okay sounds good. I'll change to `report-dir`, and then we'll just leave this as is for now. We can always change the behavior to something else in case we end up developing a use case for construct -> verify. I think we don't have to worry in this case, because this solution is still pretty flexible.

jamesaoverton:
I'm trying this out with OBI. If there are no results, should a report file be generated?

cmungall:
I'd say yes. We need this functioning makefile.

jamesaoverton:
Doing some more work with OBI... The `--report-dir` and `--queries` options for this `verify` command are convenient, but they're totally different from the `query` command options, so it's annoying to switch the behaviour of a single query from "throw errors" to "just warn me". We can change or add to the `query` options, but I'm not sure how we would distinguish select and construct. We could replace `query` with `select` and `construct`. I dunno.

This might be an opportunity for "documentation driven development", where we write the docs first, then adjust the implementation to match.

jamesaoverton:
My proposal in #182 is a superset of this `verify` command, with one exception: I now think that `--report-dir` should be `--output-dir` with short option `-O`, which will be more consistent with other commands.

I think that we should make that small change and merge this PR today. I need this for OBI stuff ASAP.

cmungall:
+1

OK, we'll get this done!

dougli1sqrd:
@jamesaoverton Yeah, you should always get output when you run the verify. And a report file is generated even if there are no results in the results-dir. Is that not happening for you?

jamesaoverton:
@dougli1sqrd Yes, I always get output. If the test passes, then the output file only contains the header line. I was just asking if that was more useful to delete files with just a header, but I agree with @cmungall that deleting them would cause problems for Make. The current implementation is good.

dougli1sqrd:
What do we think of my change from last week, based on @jamesaoverton's suggestion? I'm also working on that broader sparql query robot proposal we discussed.

cmungall:
Ok, let's go ahead and merge, but we also need an example in examples/README

dougli1sqrd:
Cool. Here'a  ticket to make examples: https://github.com/ontodev/robot/issues/186

dougli1sqrd:
So the verify command exiting with 1 if it finds violations is the "spec" I was using since that's what the perl script does we were using previously. However, in examples/readme.md, since we literally run those examples in the unit tests (that fact should be documented probably since I wasn't expecting the examples to be test code also) and it assumes that all commands exit 0, we have a problem. Since the example I use shows a violation sparql rule. Any suggestions?

jamesaoverton:
Hmm. The code is very particular in looking for indented lines started with `robot`. You could use Markdown "code fences" instead: triple backticks or tildes.

James

PR #194, opened 2017-09-28
reason --create-new-ontology-with-annotations

simonjupp:
I want to create an inferred view, removing redundant subclasses, but keeping all the annotation properties. What do you think about supporting something like this? Or is there already another way to achieve this that I missed?


jamesaoverton:
I think this is good. @cmungall ?

jamesaoverton:
Thanks @simonjupp!

I should have merged this right away, but I was travelling and I dropped the ball. Sorry.

PR #196, opened 2017-10-17
Use getInputOntology, fixes #195

### LINKED ISSUE #195: robot verify does not use the catalog-xml
example:

```
~/repos/robot/robot-core/src/test/resources(master) $ robot verify -i import_test.owl --queries test.sparql -O reports/
```

where test.sparql is `SELECT * WHERE { ?x ?r ?y }`

fails with:

```
org.semanticweb.owlapi.model.UnloadableImportException: Could not load imported ontology: <https://github.com/ontodev/robot/robot-core/src/test/resources/simple.owl> Cause: https://github.com/ontodev/robot/robot-core/src/test/resources/simple.owl
```

This works as expected:

```
robot convert -i import_test.owl -o foo.obo
```

This is a bit of a blocker and additionally we have no way of passing in an alternate catalog: 
#41

jamesaoverton:


jamesaoverton:
@cmungall: Can you check that this addresses #195?

Issue #214, opened 2017-12-11
Proposed new command: repair


cmungall:
This would be a counterpart to `report` (#205) 

For any automatically repairable problem, fix it. This could include lexical operations (e.g. removing trailing whitespace or unwanted characters from synonyms, and other literal fields, a common problem)

My current main use case is replacing references to obsolete classes (see #1) using the IAO 'replaced by' property.

I envision the command/op starting with minimal functionality but gradually we can add stuff.

rctauber:
I spent awhile cleaning up duplicate synonyms recently - could this be another `repair` candidate?

cmungall:
Is this duplicate as in the axiom plus annotations are identical?

There are a few different scenarios

 1. same class+literal, different synonym scope (e.g broad/narrow). Typically unintended
 2. Identical triple, except for difference is casing (or minor lexical variation). This is getting into something you'd want a specialized too to normalize though
 3. Identical triples, no axiom annotations: not possible with OWLAPI, as they would be collapsed?
 4. Identical triple, but different axiom annotations. Valid, but sometimes unintentional

For 3: one operation we have in OWLTools is merge duplicate axioms, merging axiom annotations. This doesn't quite seem like a repair operation, as having distinct copies of the axiom with their own provenance is perfectly valid.

rctauber:
I suppose I was thinking the situation in 4, where you would just merge the axiom annotations together. If maintaining the distinct copies of the axiom is important, though, that wouldn't make sense.

rctauber:
Ah, that makes sense. Could we have an option, e.g. `--merge-duplicate`?

cmungall:
How about `--merge-axiom-annotations`?

rctauber:
I think `--merge-axiom-annotations` would be a good option to have. By default, they wouldn't merge. What are your thoughts on this @jamesaoverton ?

jamesaoverton:
Yeah, let's implement this.

rctauber:
Is there anything else that needs work for `repair`, or is this safe to close?

rctauber:
I believe this is completed by #347. Please re-open if it needs more work. Thanks!

Issue #205, opened 2017-11-15
New Operation: report


cmungall:
ROBOT performs logical checks using `reason` and can run a custom suite of SPARQL queries using `verify`. Some repos successfully use this in combination with Makefiles and a custom suite of queries to implement robust checks in their pipeline (e.g. [GO](https://github.com/geneontology/go-ontology/tree/master/src/sparql); and also OSK).

We should make it easier for groups to do this out of the box. While customization is good, ROBOT should still provide an OBO-ish core set of queries and reports that could be used by OBO central for evaluation purposes. The groups could still customize and say for example "allow some classes to lack definitions".

In GO we are still very reliant on this script, it depends on obo format but it is a good model for a set of potential standard checks: https://github.com/geneontology/go-ontology/blob/master/src/util/check-obo-for-standard-release.pl

This could be implemented by robot distributing some sparql in its resources folder (or a standard obo place to fetch these) and/or procedural code (sometimes easier than sparql)

cmungall:
Proposed name for this: `report`

the output product could be called a "report card"

cmungall:
Should this command also produce standardized exports of things like tables of all classes and their labels, or is this overloading? Would that belong in `convert`?

jamesaoverton:
I've used SPARQL for this. It's important to distinguish internal from external (imported) terms.

I'm fine with including this in `report`.

cmungall:
Some suggestions on potential output. For checks, these could be treated as informative for now. Working with obo-ops and the obo community we could come up with varying levels of stringency. For some we can imagine different profiles, with the profile (e.g. obo-basic) declared in header.

general reporting

 * aggregate statistics
    * number/proportion of classes with textual definitions
    * number/proportion of classes with textual definitions that have/lack provenance
    * number/proportion of classes with owl definitions
    * number/proportion of classes with synonyms

checks

 * cardinality violations on annotation properties
    * rdfs:label
    * text defs
 * uniqueness
    * rdfs:label
    * text defs
    * exact syns+labels
 * logical checks
    * coherent (we get this from reason already but nice to bundle together)
    * no equivalent class pairs in main ontology
 * lexical checks (many can be `repair`ed)
    * trailing whitespace in literals
    * text defs first character capitalized
    * text defs end in `.` [OR: none end in `.`]
    * no illegal characters
    * most APs (label, syn, def) should never have newlines or formatting characters
    * CURIEs (e.g value of has_dbxref) should have declared prefix and be valid CURIEs
[...]

PR #215, opened 2017-12-12
Repair and Report operations and commands

cmungall:
Adds

 - report #205 
 - repair #214 

Both are quite slim at the moment. The idea is that we can just keep adding new checks, and give the user leeway in what is considered failures vs warnings. Some may be simply lint, others may be things that OBO decides are of importance for ontology quality.


jamesaoverton:
This sounds good to me. Let me know if you want me to review it. Otherwise just merge when you're ready.

Issue #183, opened 2017-08-22
Add `remove` command


jamesaoverton:
User story: OBI has a "Core" subset, designed to be small and focused. I have some old code for extracting the Core using SLME STAR. I want to replace the old code with ROBOT, and the `extract` command should work, but there are some problems:

- I don't want to include all the individuals, which drag in their parent classes
- a few classes have complex SubClassOf expressions that I don't want to include, because they drag a bunch of other classes in with them

So I would like ROBOT to be able to remove logical axioms for all individuals, and strip anonymous superclasses from targeted classes.

This functionality doesn't exist in ROBOT, and I'm not sure where to put it. The closest command is `filter`, which specifies which ObjectProperties to keep, but I want to specify things to remove. So here's my idea for a `remove` command:

```
usage: robot remove --input <file> --class <arg> --output <file>

 [skip global options]
 --class <arg>                  remove a Class
 --individual <arg>             remove a NamedIndividual
 --object-property <arg>        remove an ObjectProperty
 --annotation-property <arg>    remove an AnnotationProperty
 --datatype-property <arg>      remove a DatatypeProperty

 --all-individuals              remove all NamedIndividuals
 --descendant-classes <arg>     remove all classes descended from a class
 --anonymous-superclasses <arg> remove all anonymous superclasses from a class
```

@cmungall: thoughts?

cmungall:
On the one hand, satisfying all the different kinds of custom removals without any kind of DSL will get hard (see #149 - ability to add in SPARQL may give some additional expressivity)

On the other hand, Robot is an OBO tool and we should perhaps try and do more to standardize what 'core' means in OBO and not worrying about having a release tool that is overly fit around that use case.

OBI core sounds like OBO basic http://owlcollab.github.io/oboformat/doc/obo-syntax.html#6.2 -- which itself is a composition of [other properties](http://owlcollab.github.io/oboformat/doc/obo-syntax.html#6.1). Although currently defined in the obo format doc, it's generalizable, and should perhaps be elevated to OBO-wide. Some of the constraints above are implicit in the limitations of OBO format full, e.g. no individuals.

cmungall:
go-basic http://geneontology.org/page/download-ontology

cmungall:
It would be really useful to have a standard basic/core exporter in robot

We go through some gymnastics at the moment for various ontologies, this is typical:
https://github.com/EnvironmentOntology/envo/blob/master/src/envo/Makefile#L51-L70

[...]

PR #221, opened 2018-01-03
DO NOT MERGE: Remove command

rctauber:
For #183 - see discussion in issue


jamesaoverton:
Let's wait on #224, because we might want to reuse `RelatedEntitiesHelper` (or something like it).

rctauber:
See #290

Issue #212, opened 2017-12-08
Better import configuration


jamesaoverton:
ROBOT has a range of `extract` features that take lists of term IRIs. In practise, we often have mixed import strategies. Currently that requires multiple `extract` calls in a `Makefile`, or something. It would be nice to have more control in a friendlier interface.

1. OntoFox has configuration files: http://ontofox.hegroup.org/tutorial/index.php#input_format
2. The original MIREOT implementation would accept a very basic OWL file and fill it out: https://github.com/ontodev/robot/issues/10#issuecomment-112109599
3. OntoPilot has very nice spreadsheets with terms and import strategies: https://github.com/stuckyb/ontopilot/wiki/Managing-imports

It would be nice to have at least one of these in ROBOT, and having all of them would be even better.

Whatever we do should be integrated with the ontology-starter-kit: https://github.com/INCATools/ontology-starter-kit/blob/master/template/src/ontology/imports/MY-IMPORTED_terms.txt

rctauber:
The OntoPilot CSVs for the terms include Related Entities and Exclude fields. Do we want to have these in an implementation for ROBOT?

Related entities
> Whether to retrieve entities that are related to the target entity for inclusion in the import module. This can either be empty, in which case no related entities are explicitly retrieved, or it can be a comma-separated list of one or more of the following values: ancestors, descendants, equivalents, disjoints, domains, ranges, inverses, types, or property assertions.

Exclude
> If the value of this field is “true”, “yes”, “t”, or “y” (not case sensitive), then this entity (and related entities, if specified in the Related entities column), will be explicitly excluded from the final import module.

I'm trying to think of a way to make it most user-friendly. I was thinking we could have a master source file like OntoPilot, where you specify the IRIs and entity files for each import... plus maybe a method (MIREOT, STAR, etc.). Then, for each entity file, just a simple list of CURIEs?

jamesaoverton:
I think we want these things, but the first implementation can focus on operations that we already support.

rctauber:
I got a basic implementation of this working. It definitely needs some refinement, but it gets the job done.

The user passes in a csv to `--source-file`, and does not need to add an `--input` ontology. Right now, I thought it would be best to keep it separate from merge. If we wanted to do a full extract & merge into an input ontology, that might make more sense as a new command? Just thinking out loud.

The source file looks something like this:
```
ID,IRI,method,term file,output
[...]

PR #224, opened 2018-01-10
DO NOT MERGE: Better import configuration

rctauber:
Opening this to discuss implementation from #212 

The `import` command expects `--source <file>` which is a CSV containing import specifications as follows:
```
ID,IRI,term file
BFO,http://purl.obolibrary.org/obo/bfo.owl,bfo_imports.csv
```

The term file for each import has specifications as follows:
```
ID,method,relation options separated by spaces (optional)
BFO:0000002,STAR,ancestors disjoints
```
The command doesn't expect headers like that, but I can add to ignore if they're included...
MIREOT is not implemented right now, it's marked as a TODO.


jamesaoverton:
This didn't work out as well as we hoped. This PR is a bit of a mess. If we decide to revisit this feature we'll start a new PR.

Issue #226, opened 2018-01-14
Support annotations on annotation assertions with templates


rctauber:
Many ontologies use `database_cross_reference` as an annotation on a definition to cite sources. Both ECO and DO rely heavily on templates to create new terms, but we have to do this part manually.

I would like to look in to adding a new type of template string to support this. Not sure the best way to implement it on the template, but I was thinking something like:
```
A rdfs:comment [A rdfs:comment]
This is a comment [This is a comment on the comment]
```
My other thought was to do it as the column directly to the right of what you want to annotate with `AA` to specify that it's an annotation on that:
```
A rdfs:comment       AA rdfs:comment
This is a comment    This is a comment on the comment
```
I'd appreciate any input! I'm going to start working on the operation first.

rctauber:
Got it working with the first method - that seems to be the easiest way. I changed the `get*Annotation` methods to return axioms instead of annotations. I pushed the changes to a new branch on my fork.

I did test it out briefly with this csv:
```
id,comment,other comment
ID,A rdfs:comment [rdfs:comment],A:rdfs:comment
ID:1,Comment 1 [Comment 2],Comment 3
```

And it seemed to work fine (yay!)
```
<Class rdf:about="ID:1">
    <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Comment 1</rdfs:comment>
    <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Comment 3</rdfs:comment>
</Class>
<Axiom>
    <annotatedSource rdf:resource="ID:1"/>
    <annotatedProperty rdf:resource="http://www.w3.org/2000/01/rdf-schema#comment"/>
    <annotatedTarget rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Comment 1</annotatedTarget>
    <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Comment 2</rdfs:comment>
</Axiom>
```
I'm going to play around with it more tomorrow, since I have a handful of new DO terms to add.

jamesaoverton:
That output looks right.

I prefer the second option above: using separate columns. I think it's more natural, won't get mixed up with content of the annotation (e.g. square brackets in the content), and should allow for multiple annotations.

rctauber:
I'll try and sort that out. That makes things a little more complicated (which I didn't think about before I dove into the code)... it could be done with another loop though.

[...]

PR #228, opened 2018-01-15
Add support for axiom annotations

rctauber:
See #226


jamesaoverton:
Good. We need:

- to make sure this also works to annotate both annotation axioms and logical axioms
- add/modify at least one unit test to cover those two cases
- update the documentation, keeping our terminology as clear as possible: https://github.com/ontodev/robot/issues/226#issuecomment-357828902

rctauber:
* Logical axiom annotations are supported with `>C`.
* I updated `test/resources/template.csv` to include `>*` template strings. The axiom count stays the same, though, since they're included in the axioms they're annotating. If we wanted to check these specifically, I think we'd have to loop through the axioms and make sure the `rdfs:comment` annotations are included. What are your thoughts on this?
* Documentation updated. I didn't include `>>A` even though it's supported... I think anybody who's new to the technical aspect of ontology development might be overwhelmed by that.

jamesaoverton:
Good. Two things: 

1. I'd feel better if the template was more thoroughly checked. I suggest saving a known-good `template.owl` output of `template.csv` into the repo, and making a unit test against to diff against that.
2. Somebody is going to try `>AL` and wonder why it doesn't work. Can you generalize to handle `AL`, `AT`, `AI`?

robot-core/src/main/java/org/obolibrary/robot/TemplateOperation.java
```diff
@@ -244,7 +241,7 @@ public static boolean validateTemplateString(String template) {
     if (template.equals("CLASS_TYPE")) {
       return true;
     }
-    if (template.matches("^(A|AT|AL|AI|C) .*")) {
+    if (template.matches("^(A|>A|>AT|>AL|>AI|>>A|>>AT|>>AL|>>AI|AT|AL|AI|C|>C) .*")) {
```
jamesaoverton: Please simplify to something like: `^>?C|^>{0,2}(A|AT|AL|AI)`

jamesaoverton:
I checked this on OBI and MRO, both of which use several templates, and it worked fine.

Please make that small regex simplification and I'll merge.

PR #234, opened 2018-01-17
Suppress reason output

dougli1sqrd:
This ensures that only 10 dangling reference violations are printed out after the reason command, in order to reduce logging output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: java
 jdk:
   - oraclejdk8
-script: "mvn verify"
+script: "mvn verify && ls bin/robot.jar"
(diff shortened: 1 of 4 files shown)


jamesaoverton:
Fine with me.

Completely optional suggestion: You could count all the dangling axioms and print the total at the end.

cmungall:
Either good with me. The main thing is the reason command points to the existence of a potential issue. People can use report to get the complete report.

dougli1sqrd:
There should be a count also at the top of the errors. 
`ERROR Reference violations found: 1278 - reasoning may be incomplete`

Issue #239, opened 2018-01-30
Improve error reporting


rctauber:
Some errors throw complex stack traces, non-informative messages, or show nothing.
For example, #153.

In addition to the issue above, I also found:
* If no `--method` is specified in `extract`, an `IndexOutOfBoundsException` is thrown
* Both `merge` and `unmerge` can be run with only one input (no feedback)
* If two arguments are not included with `annotate --annotation`, an `IndexOutOfBoundsException` is thrown (instead of specifying that both a property and value need to be provided)
* Similarly, if a valid property is not provided first, ROBOT throws `NullPointerException: iri cannot be null` with no other details

These are just a few examples, but I plan to go through the commands and see where feedback could be provided for users. Please update this thread if you know of any other cases where more detailed error reporting would be useful. Thanks!

rctauber:
Hey @cmungall - I was going over the commands that use reasoners and noticed a discrepancy between `reason` and the others (`reduce` and `materialize`).

`reason` supports ELK, HermiT, 'structural', EMR, and JFact. The others only support ELK, HermiT, and 'structural'. Is there a reason why they can't use EMR and JFact? I was hoping to make one reusable method to get the `reasonerFactory` from the `--reasoner` passed in.

cmungall:
Thanks for doing this, I have been meaning to report some of these for a while.

Reason: yes it would be good to unify this. The EMR is a bit of a special case. It supports inferred class expressions as parents, so `reason` detects if the EMR is present an will switch to this behavior. There is some discussion on the tracker about whether or not to merge materialize and reason.

But I don't see why we can't unify the command options and factory generation

rctauber:
I can always put a check that makes sure EMR isn't input for reduce and materialize, for now, if things need to be updated.

My thought was to add `getReasonerFactory(String reasonerName)` to `IOHelper`... I don't want IOHelper to be the catchall for any multi-purpose operation method but I also don't see the benefit of creating another helper class for just one method.

Issue #237, opened 2018-01-27
support wildcards for file input


wdduncan:
I have a directory of 10 turtle files I'd like to merge. Is it possible to allow wildcards in the input pattern?

e.g., robot merge --input *.ttl --output merge.ttl

jamesaoverton:
I can see how this would be helpful.

It can also be a little dangerous. If you run `robot merge --input *.ttl --output merge.ttl` twice, then on the second pass `merge.ttl` will be included in `*.ttl`.

wdduncan:
Yes. I see the danger. I would have to careful not to merge twice. Although, if merge.ttl was included in the merge process, is there anyway to check for this and throw an error?

jamesaoverton:
We'll look into adding wildcards.

I've never seen another command line tool that checks that the output is not in the input. I think this would add complexity and violate the [Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). You'll have to be careful, like with any other command line tool. I just wanted to point out the danger in the example you provided.

cmungall:
I agree. All unix tools allow you to shoot yourself in the foot this way. I think if you're using wildcards then you generally know this. Less unix aware users who are just following recipes will be less likely to use them.

We can still have a line in the example docs warning against this

And wildcards aside, it might be worth noting in the docs that

`robot OPERATION -i foo.owl -o foo.owl`

is an anti-pattern. IMO although technically safe transformations should never overwrite, bad practice.

jamesaoverton:
Should be fixed by #246

PR #246, opened 2018-02-12
Improve error reporting

rctauber:
See #239.

Error messages are now constants at the beginning of classes to keep track of number IDs. All known error messages are referenced in the appropriate doc page, including a new "errors" doc page. I tried to keep all error reporting in `robot-command`, but there are a few exceptions that had to stay in the core classes. `IOHelper` and `CommandLineHelper` also have detailed error messages for general operations.

I went through each specific command and tried to find areas where we could give better feedback.
Error messages returned in the console include a link to the error message in the documentation.

I enforced checks on `--input` (or `--input-iri`) options (typically one input allowed, unless merging or unmerging).

I added support for wildcards with an `--inputs` option in merge and unmerge (see #237). The patterns must include ? or * and be quoted. If they are not, the user will get an error message (which also has info in the docs). e.g.:
```
WILDCARD ERROR --inputs argument must be a quoted wildcard pattern
See http://robot.obolibrary.org/errors#input-9 for more details
```

Template-specific convenience methods have been moved from `IOHelper` to a new `TemplateHelper` class to reduce size and complexity.

Whew, I think that pretty much sums it up, but I would appreciate any review (sorry it's massive!).


jamesaoverton:
@cmungall Do you have any objections/suggestions for this PR? If not, I'll double-check it and merge ASAP.

cmungall:
Go ahead and merge, thx!

jamesaoverton:
Thanks. I’ll review and merge tomorrow morning.

PR #248, opened 2018-02-12
Update relax.md

cmungall:
Providing motivation for use of relax

### DIFF
--- a/docs/relax.md
+++ b/docs/relax.md
@@ -2,4 +2,27 @@
 
 Robot can be used to relax Equivalence Axioms to weaker SubClassOf axioms. The resulting axioms will be redundant with the stronger equivalence axioms, but may be useful for applications that only consume SubClassOf axioms
 
+Example:
+
     robot relax  --input ribosome.owl --output results/relaxed.owl
+    
+## Motivation
+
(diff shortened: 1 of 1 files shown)


cmungall:
@rctauber @jamesaoverton is this a reasonable level of detail? Or do we want to keep descriptions minimal and point to additional docs?

It seems when writing these we want to accidentally avoiding tripping unit tests. I believe these are indicated by 4 spaces. So quoting non-exec code blocks with backquote sections OK?

jamesaoverton:
I think that part of the goal of these individual command pages is to be able to add as much detail as we want.

Yes, indentation is used for the integration tests. Code fences (three backticks or tildes) are ignored.

rctauber:
That looks like good details to me. This sentence is a little awkward...
>> These are required for being able to use the [reason](reason.html) command to classify an ontology.

Maybe rephrase as
>> These are required to utilize the reason command...

Also, I like the details about using it in combination with other commands. Maybe we can start also putting together a doc page for releases with ROBOT? Or would that fall more under the ontology starter kit?

cmungall:
I like the idea of having a more high level view on the robot site, where we can walk people through gradually the concepts of chaining and providing the output of one command as input of another. Then at the end we can point people towards the osk as an example implementation.

Issue #64, opened 2016-04-07
When writing OBO format, allow the equivalent of owltools `--no-check`


cmungall:
From https://github.com/ontodev/ncit-obo/issues/12

The default OBO writer is very strict in enforcing closed-world tag cardinality rules http://owlcollab.github.io/oboformat/doc/obo-syntax.html#4

This is useful for linting, and also to make sure nothing odd happened in the obo->owl->obo. But sometimes you just want to say "pick any definition, I don't care". This should never be default, but useful option.

See https://github.com/owlcollab/owltools/blob/master/OWLTools-Core/src/main/java/owltools/io/ParserWrapper.java#L241

cmungall:
this is the relevant line from owltools

```
    oboWriter.setCheckStructure(isCheckOboDoc); 
```

jamesaoverton:
This seems to be the only reason that `owltools` is required for <https://github.com/INCATools/ontology-starter-kit>. (I'm not sure about `ontology-release-runner`.)

What would be required for ROBOT to do this properly? It seems like a small thing...

cmungall:
literally just an extra option on `-o` that calls `oboWriter.setCheckStructure(isCheckOboDoc)`.. or we could just make it the default

cmungall:
It works, thanks!

PR #256, opened 2018-02-23
Add --check option to convert

rctauber:
`--check false` allows the user to ignore the OBO document structure rules, see #64 

By default, `--check` is true.

Updated documentation included.


jamesaoverton:
@cmungall or @dougli1sqrd Can you see if this works for your use cases?

jamesaoverton:
@cmungall or @dougli1sqrd does this solve #64 for you?

jamesaoverton:
Testing this with `convert -i obi.owl -o obi.obo` (without `--check`) results in an empty `obi.obo` file. Failures should not generate empty files.

rctauber:
Merging, looking into fix for the empty output here: https://github.com/owlcs/owlapi/issues/741

PR #264, opened 2018-03-07
Robot command timing

dougli1sqrd:
This adds in a timing output statement for each subcommand in robot. 

I made the addition to `CommandState` in order to convey that a command had fatally erred. The existing way that we used to do this was just to `System.exit()`, but when that would happen (say with the verify command) the command wouldn't actually complete, and how long it took wouldn't be known. So the `CommandManager` now inspects if the `CommandState` has an error flag set, and if it does, then it calls System.exit, only after displaying the duration.

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/CommandManager.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/CommandManager.java
@@ -159,6 +159,9 @@ public class CommandManager implements Command {
 
     while (arguments.size() > 0) {
       state = executeCommand(state, globalOptionArgs, arguments);
+      if (state.hadAnError()) {
+        System.exit(1);
+      }
     }
 
     return state;
@@ -216,7 +219,14 @@ public class CommandManager implements Command {
     optionArgs.addAll(globalOptionArgs);
     optionArgs.addAll(localOptionArgs);
 
-    return command.execute(state, asArgs(optionArgs));
+    long start = System.currentTimeMillis();
+
+    CommandState s = command.execute(state, asArgs(optionArgs));
+
+    double duration = (System.currentTimeMillis() - start) / 1000.0;
+    System.out.println(commandName + " took " + duration + " seconds");
+
+    return s;
   }
 
   /** Print general help plus a list of available commands. */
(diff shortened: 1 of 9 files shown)


jamesaoverton:
I'm reluctant to change the return types on important methods such as `reason`.

Would it work to just wrap the try/catch block in `CommandManager.java` around 240? Like this:

```
long start = System.currentTimeMillis();
try {
    state = command.execute(state, asArgs(optionArgs));
} catch (Exception e) {
    double duration = (System.currentTimeMillis() - start) / 1000.0;
    System.out.println(commandName + " took " + duration + " seconds");
   ...
```

If that's good enough, then I don't see the need for `hadAnError`.

dougli1sqrd:
Previously though there wasn't a return. So this doesn't break anything previous, since not saving the return value is totally fine. But, if that's uncomfortable, we could make what you have above work if we threw exceptions for errors inside commands instead of exit().

jamesaoverton:
Yes, I'd prefer that approach, with `VerifyCommand` and `ReasonCommand` throwing exceptions rather than exiting. Exiting also makes them harder to test.

cmungall:
yes, it's standard in java to use exceptions rather than return codes.

jamesaoverton:
I can look more closely at this tomorrow.

The worry that I mentioned above is that other people might be using our code. We give our code away free, so we never know who might be using it or how they're using it. Whenever we change a signature that they depend on, we break their code. I want to avoid doing that as much as possible. So I want to avoid changing return types, adding new required arguments, and throwing new kinds of exceptions. I've never seen the benefit of adding new exception classes. I prefer to just throw existing exceptions.

I would prefer not to add `hasAnError` with this change.

Overall, adding this timing information to ROBOT is a good idea.

cmungall:
Just spoke to @dougli1sqrd 

Our main motivation here was finding out why the GO travis tests were taking up to 45 mins. On reflection it may have been easier to just change the log4j conf to show timings an all INFOs (maybe there could be a convenient shortcut options for this). We may want to look at logging in general as we're quite chatty...

I I think this will serve our needs for now. In future another way to get at the timings will be look in the header, when we start injecting this (#6).

It may be worth looking at our strategy for exceptions and state, but this will be done in a separate PR, probably with some F2F discussion in advance.

In this one particular case, there was no need to change the signature, as we already have `OntologyLogicException` being thrown. The equivalence check violation would be a new exception subclass of this. But anyway, that would be a separate issue from the timings

PR #267, opened 2018-03-12
Reporting

rctauber:
See #205 

Set of SPARQL queries and a YAML output. Currently, all violations are level 5 severity, but the reports can be run.


robot-core/src/main/java/org/obolibrary/robot/checks/Report.java
```diff
@@ -0,0 +1,204 @@
...
+    info = new HashMap<>();
+    warn = new HashMap<>();
+    error = new HashMap<>();
+
+    infoCount = 0;
+    warnCount = 0;
+    errorCount = 0;
+
+    // get ontology-specific labels
+    labels = OntologyHelper.getLabels(ontology);
```
jamesaoverton: This is brutally slow, even for a smallish ontology such as OBI. My code, my fault.

SPARQL would be faster. The most efficient approach would be to get only the labels that we need, once the violations have been collected.

cmungall:
I was able to build this locally, worked as expected, great!

I'm expecting that we will want 2 classes of queries: QC checks (>0 rows indicates an issue) and reports (>0 rows expected).

All the ones we have now are QC
https://github.com/rctauber/robot/tree/reporting/robot-core/src/main/resources/queries

How should we do this? Naming convention? E.g. `-qc.rq`. Different folders? Wrap the SPARQL in YAML, or have a metadata header in `#`s?

cmungall:
Note that if we have more metadata in the header we can do more intelligent reporting. E.g. multi-parentage check makes no sense in a release ontology, but does make sense in an edit module. Label checks make sense in modules intended for editors and end-users, but not in bridge modules. See:
https://github.com/information-artifact-ontology/ontology-metadata/pull/36

cmungall:
Appears to exit with non-zero error code even when `--fail-on none` is set

cmungall:
On 21 May 2018, at 3:22, Rebecca Tauber wrote:

> @cmungall I set the multi-parentage check to only return instances 
> where there were multiple parents, but none of them had `is_inferred 
> true`. Similarly, with equivalent named classes, report would only 
> catch instances of `is_inferred true`. I was hoping this could solve 
> some of the issues between reporting on edits and releases, as well as 
> the ability to provide different profiles... but it might be too 
> strict.

yeh, we don't consistently annotate these axioms this way

> I like the new specifications in 
> information-artifact-ontology/ontology-metadata#36 though. Are you 
> thinking that these could go in the ontology metadata? Then, if it is 
> specified, we could use a particular profile?

exactly!

> RE `--fail-on none`... I just tested it and I got a zero exit code. 
> Can you send me the command you're using so I can recreate it? Thank 
> you!

ok will try later may have been a user fail on my part

rctauber:
@cmungall That seems like a good idea. Maybe we can decide on a default profile to use if the ontology is not annotated as well, and continue to allow customization?

docs/report.md
```diff
@@ -0,0 +1,98 @@
...
+| `obsolete_label`         | WARN  | Label doesn't begin with 'obsolete' |
+| `lowercase_definition`   | INFO  | Definition starts with lowercase    |
+| `missing_superclass`     | INFO  | Class doesn't have asserted parent  |
+
+Each query retrieves a triple in the form of `?entity ?property ?value`. The `?entity` is the violating entity, the `?property` is the property that is being violated, and `?value` is what is causing the violation (which could be empty).
+
+For example, the query to retrieve references to deprecatd classes:
+```
+SELECT DISTINCT ?entity ?property ?value WHERE 
+  {?value owl:deprecated true .
```
jamesaoverton: `true` has to be quoted.

docs/report.md
```diff
@@ -0,0 +1,98 @@
...
+If the report fails, the command will exit with a status of `1`, but a report will still be generated and written to the output file. To always pass (even with errors), change the level to `--fail-on none`.
+
+## Queries
+
+`report` includes a set of 15 QC queries, each with a default logging level.
+
+| Query Name               | Level | Violation                           |
+|--------------------------|-------|-------------------------------------|
+| `definition_cardinality` | ERROR | Entity doesn't have ONE definition  |
+| `deprecated_class`       | ERROR | Deprecated class used in axiom      |
```
jamesaoverton: Looks like it's called `deprecated_class_reference`

PR #268, opened 2018-03-12
Robot timing 2

dougli1sqrd:
Okay,
This is a smaller, dumb simple version of what I had before. This just prints out the timing, and I made a log file change to also print date/time on a log message.

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/CommandManager.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/CommandManager.java
@@ -237,11 +237,15 @@ public class CommandManager implements Command {
       }
     }
 
+    long start = System.currentTimeMillis();
     try {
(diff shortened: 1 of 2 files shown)


jamesaoverton:
I'm happy with this approach. However I don't think we should print timing information every time ROBOT is run. Instead of `System.out.println` can we use the logger, or watch for an option such as `--verbose`? And I guess `reason` and `verify` are still exiting on failure, so we won't see the timing information.

dougli1sqrd:
I can do that. The only hesitancy I feel is even one level of verbosity to WARNING has a huge log output. I thought about a specific option for displaying the duration? And it's true, reason and verify, if they fail we won't see duration times, but that won't be a big deal I guess.

jamesaoverton:
The simplest thing would be to log the timing using a distinct string that you can grep for.

Issue #269, opened 2018-03-12
Improve efficiency of ReduceOperation


cmungall:
This takes 6m on go-plus: https://github.com/geneontology/go-ontology/pull/15284

oddly, it only takes <1m on go.obo

It may be worth exploring an approach this is not directly coupled to the OWLAPI - e.g. using JGraphT

yy20716:
@cmungall @dougli1sqrd If you don't mind, could you please clarify how I can re-produce this issue?, i.e. I'd like to know how I can test the robot with go.owl and go-plus.owl. Do you think I just need to download go.owl and go-plus.owl from [this svn site](http://viewvc.geneontology.org/viewvc/GO-SVN/ontology-releases/2018-03-12/), run them as follows, and compare the performance to reproduce your issue (1m vs. 6m)?
~~~~
robot reduce --reasoner ELK --input go.owl --output go_output.owl
robot reduce --reasoner ELK --input go_plus.owl --output go_plus_output.owl
~~~~
If this is not the way I can use, could you please let me know how to re-produce this issue? I also saw Eric's comment in your post, but it seems that this robot command is not accepting go.owl file but enhanced.owl. It also seems that later reasoned.owl is changed into go-plus.owl, but not exactly sure about my understanding.
~~~~
ROBOT_JAVA_ARGS=-Xmx12G robot reason -i enhanced.owl -r ELK -e asserted-only relax reduce annotate -V http://purl.obolibrary.org/obo/go/releases/date +%Y-%m-%d/go.owl -o reasoned.owl
~~~~
Any suggestions would be greatly appreciated.

dougli1sqrd:
In the `go-ontology` repository, in the `src/ontology/` directory there is a makefile. If you put your local robot and some owltools on your PATH, and run `make travis_test` you can duplicate. The make target that I noticed take a long time is the `reasoned.owl` target (which runs during `travis_test`)

yy20716:
Eric, thanks for letting me know. I will run "make travis_test" as you suggested and see what I can do.

PR #270, opened 2018-03-14
Optimized codes for reduce operation

yy20716:
This patch is built for handling the issue discussed in [this post](https://github.com/ontodev/robot/issues/269). 
 - I found that some codes used for debugging were causing the performance issue. [The previous codes](https://github.com/ontodev/robot/blob/fab33b94297d1df91646426808c85cddf767f5c6/robot-core/src/main/java/org/obolibrary/robot/ReduceOperation.java#L174) were built in a way that they always iterate and compare all members in `exprToNamedClassMap`, which can be time-consuming if the size of `exprToNamedClassMap` is very large. Instead of using the iteration, the codes added below now use an inverse map of `exprToNamedClassMap` to quickly find out relevant members, if we need to print out explanations for the redundancy. 

 - I confirmed that the codes in this branch passed all testcases. I also observed that the current master version took approx. 8 mins but this patch reduced the execution time up to approx. 2.5 mins on my laptop. I specifically used the following option for testing the codes, which was extracted from `Makefile` for go ontology.
~~~~
time ROBOT_JAVA_ARGS=-Xmx12G robot reason -i enhanced.owl -r ELK -e asserted-only relax reduce -o test.owl
~~~~

 - @dougli1sqrd If you don't mind, could you please test this branch on your machine when you are available? I guess you might just need to run `make travis_test` as you suggested before. Thank you so much for your time.

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/ReduceOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/ReduceOperation.java
@@ -1,5 +1,8 @@
 package org.obolibrary.robot;
 
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


dougli1sqrd:
Yeah this is much faster!

PR #276, opened 2018-03-30
[WIP] Assert additional axiom types in ReasonOperation

### LINKED ISSUE #273: reason command doesn't infer types on individuals
I've been testing out robot on the VFB pipeline. This requires instantiation of inferred SubClassOf and Type axioms to named classes, followed by redundancy stripping for both.  'reason' appears to work fine for SubClassOf axioms but not for Type axioms.  Can this be fixed?

cmungall:
Fixes #273


rctauber:
I would expect all to be the default, and then have options to choose not to assert certain things if the user wants.

cmungall:
cc @mcourtot we can add an option for materializing indirect as well as indirect here, which would solve your CL use case

jamesaoverton:
@cmungall Sorry that this has languished. @matentzn made a bunch of good points on the issue that I don't fully understand. I'm still worried about backwards compatibility. I think I'd prefer just adding the option to select your axiom generators.

cmungall:
I agree.

Issue #275, opened 2018-03-27
behavior of 'merge' with imports


balhoff:
When `merge` is used with multiple `--input` arguments, it seems like imports are merged in and import declarations are lost. I know I have requested in the past that imports declarations are removed when a single `--input` argument is merged with its imports. But when multiple arguments are being merged, I would not like the imports to be merged, and also not be removed.

It seems to be working as the docs say (I just don't want it to):

>You don’t need owl:import statements: you can merge any number of ontologies by using multiple --input arguments. All the ontologies and their imports are merged into the first ontology.

balhoff:
I just edited my original description—it does work as it says, I just want to be able to not merge imports.

jamesaoverton:
We have to maintain backward compatibility, but it would be great to have option(s) that give control over when import statements are kept and dropped. It's clearly causing a range of problems for people. I don't have a specific solution in mind.

balhoff:
Could be as simple as `robot merge -i ont1.owl -i ont2.owl --merge-imports false --remove-imports false`

A single `-i` combined with `--merge-imports false` could produce a no-op warning.

More fine-grained combinations of merging imports for different inputs could probably be accomplished via command chaining.

rctauber:
There is a collapse-import-closure option for the merge command, but there is no code in ROBOT to support it. I can make this fix.

This ignores the imports, but does not remove the import statements. I was hoping to add a way to remove import statements in the “remove” command, but I haven’t quite thought about the best way to implement it. I made a comment about this in #272.

In the end, you should be able to merge with collapse-import-closure as false, then chain with remove (import statements) and trim dangling entities (to remove the empty entities that may be left over from the removal of import statements).

balhoff:
So I think you're saying if `--collapse-import-closure false` then imports would not be removed. That sounds good. Currently `merge` seems to always remove import statements.

rctauber:
Correct. If you merge normally, all imports get merged in and the import
statement is no longer needed, since all the external terms are now
declared in the resulting ontology.

With this as false, you’re not merging the imports into your resulting
ontology, those terms are still external, and the import statement remains.
Hopefully that’s what you’re looking for?

rctauber:
This seems to be resolved, and included in 1.1.0. Please reopen if it requires more discussion.

PR #277, opened 2018-04-03
Support collapse-import-closure & include-annotations in merge

rctauber:
See #275

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/TemplateCommand.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/TemplateCommand.java
@@ -34,15 +34,19 @@ public class TemplateCommand implements Command {
   /** Initialize the command. */
   public TemplateCommand() {
     Options o = CommandLineHelper.getCommonOptions();
-    o.addOption("i", "input", true, "load ontology from a file");
-    o.addOption("I", "input-iri", true, "load ontology from an IRI");
+    o.addOption("i", "input", true, "load ontology to merge from a file");
+    o.addOption("I", "input-iri", true, "load ontology to merge from an IRI");
     o.addOption("o", "output", true, "save ontology to a file");
     o.addOption("O", "ontology-iri", true, "set the output ontology IRI");
     o.addOption("V", "version-iri", true, "set the output version IRI");
     o.addOption("t", "template", true, "read template from a file");
-    o.addOption("a", "ancestors", false, "MIREOT ancestors into results");
+    o.addOption("A", "ancestors", false, "MIREOT ancestors into results");
     o.addOption("m", "merge-before", false, "merge into input ontology before any output");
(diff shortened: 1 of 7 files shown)


jamesaoverton:
This looks good, but I have two concerns about backwards compatibility:

1. I believe that in `merge(List<OWLOntology> ontologies, Map<String, String> mergeOptions)` the `mergeOptions` was never used. That's not great, but I don't see a reason to remove a method we've published

2. I think that this changes the default behaviour of `--collapse-import-signature` from `false` to `true`. I prefer not to change behaviour, even if the old behaviour was wrong. In this case I might be convinced to break that rule. I dunno.

Are there other breaking changes that I missed?

rctauber:
1. Instead of having booleans for `collapseImports` and `includeAnnotations`, I could change this to keep `mergeOptions` and have it be a map (eg. `collapseImports` or `COLLAPSE_IMPORT_CLOSURE`, `true` as string).
2. I don't feel strongly either way on the default, so whatever you think is best works for me.

The only other big change was including the option for the `template` command.

robot-command/src/main/java/org/obolibrary/robot/TemplateCommand.java
```diff
@@ -34,15 +34,19 @@
...
-    o.addOption("i", "input", true, "load ontology from a file");
-    o.addOption("I", "input-iri", true, "load ontology from an IRI");
+    o.addOption("i", "input", true, "load ontology to merge from a file");
+    o.addOption("I", "input-iri", true, "load ontology to merge from an IRI");
     o.addOption("o", "output", true, "save ontology to a file");
     o.addOption("O", "ontology-iri", true, "set the output ontology IRI");
     o.addOption("V", "version-iri", true, "set the output version IRI");
     o.addOption("t", "template", true, "read template from a file");
-    o.addOption("a", "ancestors", false, "MIREOT ancestors into results");
+    o.addOption("A", "ancestors", false, "MIREOT ancestors into results");
```
jamesaoverton: We can't change options that we've published, it could break something downstream.

jamesaoverton:
1. I like the explicit arguments that you have. Just leave `merge(List<OWLOntology> ontologies, Map<String, String> mergeOptions)` as an available method with the options unhandled for now (like they were before). Maybe add a warning to the docstring.

2. Let's keep the old default behaviour, even if it's annoying.

3. We can't change the meaning of the `-a` option for TemplateCommand.

rctauber:
1. Should I tag `merge(List<OWLOntology> ontologies, Map<String, String> mergeOptions)` with `@Deprecated`?
2. `collapse-import-closure` will default to `false`.
3.  I'll set `a` = `ancestors` and `A` = `include-annotations`

jamesaoverton:
I dunno. Let’s not deprecate.

rctauber:
Alright, I’ll just put in a note in the docstring then.

PR #286, opened 2018-04-25
Update for Mac OS X 10.12

safisher:
The following changes were required to run the robot script on the Mac (OS X 10.12).

I needed to change this:
  ROBOT_SCRIPT="$0"
  DIR="$(dirname $ROBOT_SCRIPT)"
to this:
  ROBOT_SCRIPT=$0
  DIR=$(dirname "$ROBOT_SCRIPT")

Note that I did not test this on any other systems but I added a test to the script to see if it's running on a Mac and if so then the following changes are implemented. If not running on a Mac then there should be no change in behavior.

### DIFF
--- a/bin/robot
+++ b/bin/robot
@@ -6,9 +6,20 @@ if uname | grep -iq cygwin; then
     IS_CYGWIN="TRUE"
 fi
 
+# Check if Mac OS X
+IS_MAC="FALSE"
+if uname | grep -iq Darwin; then
+	IS_MAC="TRUE"
+fi
+	
 # Variable to hold path to this script
 # Start by assuming it was the path invoked.
-ROBOT_SCRIPT="$0"
+if [ $IS_MAC = "TRUE" ]
+then
+	ROBOT_SCRIPT=$0
+else
+	ROBOT_SCRIPT="$0"
+fi
 
 # Handle resolving symlinks to this script.
 # Using ls instead of readlink, because bsd and gnu flavors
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Thanks for the PR.

We haven't seen this problem on macOS before. I'm a little worried about `ROBOT_SCRIPT=$0` because quoting is usually the safer thing to do. Any idea why this isn't working with the quotes?

safisher:
Ok I dug into this a bit more. Apparently the problem occurs when the robot path contains a space. On Linux it's obviously pretty uncommon to create directory structures using spaces in directory names. I happen to have done just that on my Mac. When I remove the space in the directory name then robot functions just fine as originally written.

jamesaoverton:
Ok. We definitely want to support spaces in file paths. I would think that quoting is the safe way to support spaces, and that removing the quotes would cause problems. I'm confused because your patch is doing the opposite. But maybe we're quoting too many times? Does this work for you?

```
ROBOT_SCRIPT="$0"
DIR=$(dirname "$ROBOT_SCRIPT")
```

safisher:
Yes that works.

safisher:
FYI, it's also the same issue with my owltools PR.

jamesaoverton:
Good. That should work cross-platform. So let's just make that small change, without the platform detection stuff.

If you have time to update the PR @safisher, that would be great. Otherwise I'll ask @rctauber to do it. We'll test on Linux before merging.

safisher:
Great. I'll update the PR. I'm somewhat of a GitHub novice, so if I bork it I'll submit a new PR.

jamesaoverton:
Looks good. When I get a minute I'll test under a few different systems, then merge.

Thanks again!

jamesaoverton:
I checked this on a Linux system an macOS 10.11. I think it's a solid improvement. Thanks again!

Issue #272, opened 2018-03-19
Proposal for remove and filter commands


jamesaoverton:
ROBOT has not been focused on fine-grained editing of ontologies, but we've accumulated a list of tasks that "power users" want to accomplish.

@rctauber and I have discussed and prototyped a number of options. Here's my current thinking in the form of documentation for the proposed commands. This includes the whole kitchen sink, but we can start with with the basics. The 'select' system is inspired by Brian Stucky's OntoPilot import system.

Feedback appreciated!

## Remove

The ROBOT `remove` command allows you to remove selected axioms from an ontology. The `filter` command is the opposite of `remove`, allowing you to keep only selected axioms. `remove` works in three steps:

1. `--entity`: specify a set of entities (default: all entities)
2. `--select`: select a new set of entities using one or more relations (zero or more)
3. `--axioms`: specify the axiom types to remove from those entities (default: all axioms)

For example, to remove all descendants of 'assay' from OBI:

```
robot remove --input obi.owl --entity OBI:0000070 --select descendants
```

1. `--entity` is a single IRI or CURIE, `--entities` is a file with a list of IRIs or CURIEs. These options can be used multiple times. Order is not significant: ROBOT takes the union of all specified entities before proceeding to the next step.
2. `--select` can be: `self`, `children`, `parents`, `descendants`,  `ancestors`, `equivalents`, `classes`, `properties`, `individuals`, `object-properties`, `annotation-properties`, `data-properties`, `ontology`, `anonymous`, `named`, `complement`. To match patterns you can specify the `CURIE` of a property together with an IRI value (`CURIE=<IRI>`), CURIE value (`CURIE=CURIE`), a lexical value (`CURIE='value'`), or a regular expression for matching lexical values (`CURIE~='pattern'`). ROBOT will take the current set of entities and apply the selector to generate a new set of entities. The value can be a quoted string with one or more selectors, separated by spaces. In this case, ROBOT applies each selector to the current set of entities then takes the union of the results. If more than one `--select` option is specified, then they are processed in order, modifying the  set of entities step-by-step.
[...]

PR #290, opened 2018-05-03
Filter and Remove

rctauber:
See #272 

If this looks good, the only thing I have left to do is add in documentation for the site!


rctauber:
I need to update this command in the integration tests as well:
```
robot materialize --reasoner ELK --input emr_example.obo --term BFO:0000050 \
    filter -t BFO:0000050 \
    reduce --output results/emr_reduced.obo
```

I didn't see this before, and since we changed the behavior of filter this is now wrong. I'll fix this as I update the docs for `remove` and `filter`.

docs/remove.md
```diff
@@ -1,23 +1,120 @@
...
+6. `individuals`
+
+### Patterns
+
+Entities can also be selected from the set based on axioms. This can be helpful if you want to remove only entities with a specific annotation. When selecting with axioms, the `--select` option must always be quoted.
+
+1. `CURIE=CURIE`
+2. `CURIE=<IRI>`
+3. `CURIE='literal'`
+4. `CURIE='literal'^^datatype`*
```
jamesaoverton: It's more work, but we shouldn't limit the available datatypes.

jamesaoverton:
We should not drop `FilterOperation.java`, even if we don't use it anymore. We should just point to the new approach in the docs for that file. And we might as well keep its test file.

How much does the new Guava dependency buy us?

A few examples still use `--entity`. They should use `--term`.

The options are now called `--term` and `--term-file` but we say 'entities' a lot in the documentation. I'm not sure what to do about that. We can leave it for now.

This is a major feature. I think it's good, but I want feedback from other developers and users.

rctauber:
> We should not drop FilterOperation.java, even if we don't use it anymore. We should just point to the new approach in the docs for that file. And we might as well keep its test file.

Added those back in, with a note on `FilterOperation.java`

> How much does the new Guava dependency buy us?

This was leftover from earlier, I think... I removed it and everything still runs fine.

> A few examples still use --entity. They should use --term.

Fixed.

> The options are now called --term and --term-file but we say 'entities' a lot in the documentation. I'm not sure what to do about that. We can leave it for now.

I changed a few of them to make more sense.

cmungall:
@rctauber can you resolve the conflicts in the imports, thanks!

PR #297, opened 2018-05-20
Additional options for reason and reduce

### LINKED ISSUE #295: change reduce / remove redundant subclass behavior for annotated subClassOf axioms
When I implemented this, I added code to preserve redundant axioms if they were annotated. The rationale was that removing this would result in a loss of overall information.

I think this is potentially confusing. By the principle of least surprise, a command to remove logically redundant axioms should do precisely that. However, it should be possible to override this behavior, but the default should be to do the removal.

I propose a `--preserve-annotated-axioms` boolean option on both `reduce` and `reason` commands. The default should be false.

While changing the default could surprise existing users, I am fairly sure that at least in the OBO world I made the pipelines for all ontologies that annotated logical axioms in this way.

### LINKED ISSUE #296: reason command should not inject class axioms for external classes
Currently when running `reason` on an ontology A that imports B, it is possible for axioms of the form `B1 SubClassOf B2` to be injected into A.

This can sometimes be desirable, but usually not. We should add a boolean option `--exclude-external-entities`

The precise logic is not necessarily straightforward. How do we determine what is external? Should this be based on explicit declaration axioms? Occasionally these leak into the main ontology. But this still seems more reliable that making any additional assumptions (e.g. but URI structure, or presence of other axioms).

What about edge cases, where we may infer an axiom spanning A and B?

I propose a simple test whereby for any class axiom, the class signature of the inferred axiom must overlap with the declarations of the main ontology by at least one class. If `--exclude-external-entities` is true, then anything that does not pass this is dropped.

cmungall:
Added option for reduced and reason commands, to preserve annotated axioms when removing redundant classes.

```
 -p,--preserve-annotated-axioms <arg>              preserve annotated
                                                   axioms when removing
                                                   redundant subclass
                                                   axioms
```

Fixes #295 

Added option for reason operation and command:

```
 -X,--exclude-external-entities <arg>              if true, do not add an
                                                   axiom if it is about
                                                   classes in external
                                                   ontologies
```

Fixes #296


cmungall:
@balhoff @dosumis seem OK to you?

jamesaoverton:
@cmungall Is this ready to review/merge?

rctauber:
@cmungall - do you still want this branch or can we remove it? I saw you restored it a bit ago.

PR #298, opened 2018-05-22
Search full import closure for entity labels

rctauber:
See https://github.com/obi-ontology/obi/pull/932#event-1636013540


robot-core/src/main/java/org/obolibrary/robot/QuotedEntityChecker.java
```diff
@@ -145,6 +148,7 @@ public void addAll(OWLOntology ontology) {
     Set<OWLOntology> ontologies = new HashSet<OWLOntology>();
     ontologies.add(ontology);
     ReferencedEntitySetProvider resp = new ReferencedEntitySetProvider(ontologies);
+    OntologyHelper.getEntities(ontology);
```
jamesaoverton: This line isn't doing any work.

jamesaoverton:
@rctauber Please add a unit test for this.

PR #299, opened 2018-05-24
Documentation Updates

rctauber:
Documentation updates (adding examples, clarification, etc.) and harmonizing boolean commands.

All options now accept args (e.g. true or false), but for backwards compatibility, there are a few options that also work as flags:
* `annotate --remove-annotations`
* `template --ancestors`
* `template --merge-before`
* `template --merge-after`


rctauber:
Tutorial removed in place of [new tutorial](https://github.com/rctauber/robot-tutorial)

Issue #304, opened 2018-06-07
ROBOT should show stack trace if ontology is unparseable


cmungall:
sample ontology:

```
Ontology: <http://x.org/>
Prefix: : <http://x.org/>
#
Class: A SubClassOf: B
```

this has a deliberate error, `B` is undeclared.

```
$ robot convert -i foo.omn -o foo.obo
[Fatal Error] :1:1: Content is not allowed in prolog.
File foo.omn is not a valid ontology
For details see: http://robot.obolibrary.org/

usage: robot convert --input <file> --format <format> --output <file>
 -c,--check <arg>           if false, ignore OBO document structure checks
 -f,--format <arg>          the format: obo, owl, ttl, owx, omn, ofn, json
 -h,--help                  print usage information
 -i,--input <arg>           convert ontology from a file
 -I,--input-iri <arg>       convert ontology from an IRI
 -noprefixes                do not use default prefixes
 -o,--output <arg>          save ontology to a file
 -p,--prefix <arg>          add a prefix 'foo: http://bar'
 -P,--prefixes <arg>        use prefixes from JSON-LD file
 -V,--version               print version information
 -v,--verbose               increased logging
 -vv,--very-verbose         high logging
 -vvv,--very-very-verbose   maximum logging, including stack traces
 -x,--xml-entities          use entity substitution with ontology XML
                            output
```

the output isn't informative. The first line 
```
[Fatal Error] :1:1: Content is not allowed in prolog.
```

comes from the owlapi as it cycles through a stdout-leaky jena parser: https://github.com/owlcs/owlapi/issues/550

it isn't super helpful to list the usage here since the user's command line options were not at fault.

In contrast, `owltools foo.omn -o -f obo foo.obo` shows an enormous stack trace from trying every parser - it's ugly but at least allows us to find lune number

rctauber:
Hey Chris - I will add this into the error message.

I wonder if there are other cases where we shouldn’t show the command line
options after an exception? Right now, the exception handler always returns
the CL options.

cmungall:
If there is a way to only show the usage info if incorrect options are provided (or help is explicitly asked for) that would be great

jamesaoverton:
@cmungall: Good idea. Let’s do that.

rctauber:
Just pushed some changes to a new branch, I'll make a PR. The OBO error now prints the exception message. Usage information is only printed if the exception comes from a Command class.

PR #305, opened 2018-06-11
Update exception handling

rctauber:
See #304

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java
@@ -776,7 +776,8 @@ public class CommandLineHelper {
    */
   public static void handleException(String usage, Options options, Exception exception) {
     ExceptionHelper.handleException(exception);
-    if (exception.getStackTrace()[0].toString().contains("Command")) {
+    String msg = exception.getStackTrace()[0].toString();
+    if (msg.contains("Command") && !msg.contains("updateInputOntology")) {
       printHelp(usage, options);
     }
     System.exit(1);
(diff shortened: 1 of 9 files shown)


robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java
```diff
@@ -613,7 +613,7 @@ public static IRI maybeCreateIRI(IOHelper ioHelper, String term, String field) {
    */
   public static Set<IRI> getTerms(IOHelper ioHelper, CommandLine line, String singles, String paths)
       throws IllegalArgumentException, IOException {
-    Set<String> termStrings = new HashSet<String>();
+    Set<String> termStrings = new HashSet<>();
```
jamesaoverton: Why change `HashSet<String>` to `HashSet<>` (and a few others)?

rctauber: It's less verbose - https://docs.oracle.com/javase/8/docs/technotes/guides/language/type-inference-generic-instance-creation.html

I can change it back if you prefer, it's the same in terms of performance, just makes the code a little easier to read.

jamesaoverton: Unless we're going to change it across the whole code base, I'd prefer to stick with the more explicit style for consistency.

jamesaoverton: I guess it's not consistent. Just leave it, it doesn't matter.

jamesaoverton:
Closing without merge in favour of #307.

PR #307, opened 2018-06-15
Improve error messages and CLI output

jamesaoverton:
This PR improves CLI output on exceptions by:

1. not printing usage information or stack traces by default
2. always printing an error ID, brief friendly message, and link to our online docs
3. reminding the user that `-vvv` will show the stack trace; ensuring that the full stack trace is printed; adding more information to stack traces
4. reminding the user that `--help` will show usage information

Example (from #304):

```
$ bin/robot convert --input foo.omn
[Fatal Error] :1:1: Content is not allowed in prolog.
INVALID ONTOLOGY FILE ERROR Could not load a valid ontology from file: foo.omn
For details see: http://robot.obolibrary.org/errors#invalid-ontology-file-error
Use the -vvv option to show the stack trace.
Use the --help option to see usage information.
```

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -16,9 +16,25 @@ The command line option provided only accepts `true` or `false` as input.
 
 [Chained commands](/chaining) take the output of the last command as the input of the next. Because of this, `--input` should only be used with the first command in the chain. This excludes [merge](/merge) and [unmerge](/unmerge), which allow multiple inputs.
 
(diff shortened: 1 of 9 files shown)


jamesaoverton:
Whenever we catch an exception and throw a new exception, I think we should include the original exception as a cause. If you think this is a bad idea, please reply!

PR #315, opened 2018-07-23
More consistent query

jamesaoverton:
Add `execVerify` and `runVerify` methods, simplify `VerifyCommand`.


jamesaoverton:
Unfortunately this is writing empty files when verify tests pass.

jamesaoverton:
This is still not as consistent as I would like, but it's better.

PR #318, opened 2018-07-24
Test against multiple Java versions

rctauber:


jamesaoverton:
ROBOT passed tests on all six JVM variants. That's good. On the other hand, this didn't catch the problem reported in #308, which was our goal.

I don't want to wait for six JVMs on every commit. I'm going to close this without merging.

PR #324, opened 2018-07-31
Added a feature to ReasonOperation where an unsatisfiable module can be dumped (174)

### LINKED ISSUE #174: Reasoner: Optionally create an unsatisfiable module if ontology is incoherent
owltools does this, v useful, like a core dump

cmungall:
Fixes #174

 - Added documentation to reason.md
 - Added command (-D option)
 - Function is isolated n ReasonerHelper static class


cmungall:
The logic here is a little squirrely. The "core dump" only happens if the ontology is incoherent, which triggers an exception. Before the exception is thrown, the module is (optionally) dumped. See if it makes sense...

cmungall:
@jamesaoverton I think you meant tbox not xbox? I usually often to add test files but I think I got them all this time.

@rctauber - I'm not sure how to best make a module for inconsistent ontologies. I think we would need to use the explanation feature. Even if there is unsatisfiability at the root of the inconsistency, Elk will not let you ask for unsats if the ontology is inconsistent. But yes the reporting should be better. I just pushed a change that (a) reports if the ontology is inconsistent in all cases (b) apologizes for not being able to make an unsat dump if this was asked for

jamesaoverton:
Ok. I love my Xbox.

Issue #323, opened 2018-07-28
False positives in robot report


cmungall:
```
[Term]
id: X:1
name: obsolete foo
is_obsolete: true
```

```
$ robot report -i obs.obo 
Violations: 4
-----------------
ERROR:      3
WARN:       1
INFO:       0
Level   Rule Name       Subject Property        Value
ERROR   deprecated_class_reference      obo:X_1 rdf:type
ERROR   missing_label   oboInOwl:id     rdfs:label      
ERROR   ontology_metadata       obo:TEMP        dc:license      
ERROR   ontology_metadata       obo:TEMP        dc11:title      
ERROR   ontology_metadata       obo:TEMP        dc11:description
```

The first two are false positives

 - we still have rdf:type owl:Class assertions on obsolete classes
 - the id property assertion is auto-generated in the obo2owl conversion (perhaps not the best idea but that's another discussion). We want to treat this as a builtin property, no need to declare a label

cmungall:
Actually, looks like missing_label is reported for other entities too: anything in oboInOwl, dc:date, IAO_0000231, subsets, ...

I'm not sure what the best solution is here. We could hardcode a list of prefixes of vocabs considered "builtin". Or maybe we really should be declaring labels for all these via an iao import...

The most pragmatic may be to only insist on a label if it already has a type declaration?

cmungall:
Another odd exception that must be added if this is to work for GO and some other ontologies:

duplicated_labels are allowed so long as they are for deprecated classes

rctauber:
I like the idea to only check labels if it has a type declaration. I think
that will be the best way to leave out those annotation properties, or any
other annotation properties people are using and not defining.

As for the duplicate labels for deprecated classes, that should be easy to
add in. I’ll take care of these changes on Monday.

jamesaoverton:
Close in favour of #391

PR #325, opened 2018-07-31
Fix false positive for deprecated class references, see #323

jamesaoverton:


jamesaoverton:
Partially addresses #323.

Issue #158, opened 2017-04-12
Imported ontologies should be queryable as separate graphs for robot query


cmungall:
Robot converts an OWL ontology to Turtle for presentation to Jena for SPARQL queries. It appears the ontology goes into an (unnamed?) graph.

Would it not make sense to make the whole import chain queryable, and to preserve each ontology as its own graph?

As an aside, it would be great to have more published standards here. For example, OntoBee puts each ontology in its own graph which its nice, but it does its own renaming of the URI for the graph.

I thought to implement this in Robot it would be a straightforward switch of [this line](https://github.com/ontodev/robot/blob/b599e4773158517ca32e8f24eeafcea34d5117b2/robot-core/src/main/java/org/obolibrary/robot/QueryOperation.java#L54) to use Trig...

...unfortunately saving Trig from the OWLAPI does not have the effect I would expect. It only seems to save the parent ontology, not the closure... ...and it seems to place each class in its own unnamed graph, hmm.

It would be relatively straightforward to iterate through the imports closure and add each separately, thought there may be a cleaner way.

And finally (this may deserve a separate ticket) but it is common to store inferences in a separate NG. It would be straightforward to do this stepwise in robot (`reason`, save results, and then combine this into the source ontology as an import). There may be a more elegant way to do this?

cc @balhoff @dougli1sqrd

jamesaoverton:
Separate named graphs for imports and inferences would be very nice, as long as it's still easy to query everything at once. Many (but not all?) systems make the default graph the union of the named graphs, and I think that would be good behaviour in this case. I don't remember what Fuseki does off-hand.

The graph for each import can use the import IRI as its name. The graph for the uninferred ontology can use the ontology IRI as its name. The graph with inferences would need a new IRI, which could be ROBOT-specific.

Off the top of my head, I can't think of a better method than iterating through the imports, converting to Turtle, and inserting into a named graph.

Maybe the query command can accept a `--reasoner` option to indicate that reasoning should be done.

cmungall:
On 12 Apr 2017, at 14:46, James A. Overton wrote:

> Maybe the query command can accept a `--reasoner` option to indicate 
> that reasoning should be done.

We may end up replicating a lot of options; or options may be different. 
The standard use case for the reason command is to make a new ontology 
with direct inferred links materialized. For this use case, we may want 
indirect too. We may also want to do things like shadow the Tbox in the 
Abox (ie make triples from subclass-of-some-values-from). But then maybe 
[...]

PR #328, opened 2018-08-01
Query imports as named graphs

rctauber:
See #158


robot-core/src/main/java/org/obolibrary/robot/QueryOperation.java
```diff
@@ -33,60 +39,170 @@
...
    *
-   * @param ontology The ontology to load into the graph.
-   * @return A new DatasetGraph with the ontology loaded into the default graph.
-   * @throws OWLOntologyStorageException rarely
+   * @param ontology ontology to query
+   * @return dataset to query
+   * @throws IOException on issue creating temp files
+   * @throws OWLOntologyStorageException on issue writing ontology to TTL format
    */
-  public static DatasetGraph loadOntology(OWLOntology ontology) throws OWLOntologyStorageException {
```
jamesaoverton: We can't drop this signature. I think the best option is to create a new method `loadOntologyDataset`.

Issue #332, opened 2018-08-07
Running robot with no parameters should return the help?


bjonnh:
Running robot with no parameters I get:
Exception in thread "main" java.lang.StringIndexOutOfBoundsException: String index out of range: 4
	at java.lang.String.substring(String.java:1963)
	at org.obolibrary.robot.ExceptionHelper.getExceptionID(ExceptionHelper.java:50)
	at org.obolibrary.robot.ExceptionHelper.handleException(ExceptionHelper.java:26)
	at org.obolibrary.robot.CommandManager.main(CommandManager.java:141)
	at org.obolibrary.robot.CommandLineInterface.main(CommandLineInterface.java:53)

Maybe that would be more useful if it returned the help?

jamesaoverton:
Thanks, that's definitely a bug!

Fortunately the exception points right to the problem. We'll get that fixed.

bjonnh:
Please see #333 for a proposition of fix.

rctauber:
Resolved by #333

PR #333, opened 2018-08-07
Display the help when no argument is given on the command line.

bjonnh:
It is still a bit redundant to make if/else if with return statements, but it solves the issue. We could make it cleaner by first checking arg length and then doing the case by case.

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/CommandManager.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/CommandManager.java
@@ -154,11 +154,15 @@ public class CommandManager implements Command {
    */
   public CommandState execute(CommandState state, String[] args) throws Exception {
     // Maybe print general help or version info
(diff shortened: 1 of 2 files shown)


jamesaoverton:
Much appreciated! And good to meet you this week.

@rctauber Please add a fix for the underlying problem to this PR:

```
Exception in thread "main" java.lang.StringIndexOutOfBoundsException: String index out of range: 4
at java.lang.String.substring(String.java:1963)
at org.obolibrary.robot.ExceptionHelper.getExceptionID(ExceptionHelper.java:50)
```

rctauber:
Just added a fix for that to the exception helper. Should be good to go!

jamesaoverton:
Fixed #332. Thanks again @bjonnh !

bjonnh:
Glad to meet both of you this week too. Keep up with the good work! This is a great tool.

PR #344, opened 2018-08-15
IOException for JSON-LD Parse Error

### LINKED ISSUE #153: Parse errors on jsonld prefix files are silent
E.g.

```
robot -P test.jsonld extract -m BOT -i examples/uberon_fragment.owl -t ZOG:0000483 -o test.owl
```

This gives the expected result when test.jsonld is:

```
{
     "@context": {
         "ZOG": "http://purl.obolibrary.org/obo/UBERON_"
    }
}
```

However, if a syntax error is introduced to the jsonld file, robot does not complain, the prefix is not loaded, the CURIE is not recognized, and the module outputted is empty

rctauber:
IOHelper throws IOException on issue parsing JSON-LD context. Resolves #153

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -86,6 +86,13 @@ The ontology could not be saved to the specified IRI. The most common reasons ar
 
 Each command requires a set of options, although it is more common to get a `MISSING INPUT ERROR`. See the command-specific documentation for more details.
 
+### Prefix Load Error
+
+If a prefix is incorrectly formatted, or if the prefix target does not point to an absolute IRI, the prefix cannot be loaded. Prefixes should be formatted as follows:
+```
+robot -p "robot: http://purl.obolibrary.org/robot/"
+```
+
 ### Unknown Arg Error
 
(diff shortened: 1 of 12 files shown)


jamesaoverton:
@cmungall Adding this exception to the signatures breaks backwards compatibility, which is why the automated tests are failing. It might still be a good idea. The alternative is to just complain more loudly using `log.error`. What do you think?

cmungall:
I strongly favor exceptions/fails over logging.

I realize the importance of backwards compatibility at the java level as well as command line level, but failing on errors is important to prevent bad stuff happening in release pipelines.

Maybe we should do a one time analysis of operations to ensure we future proof ourselves for exception completeness?

jamesaoverton:
To make `japicmp` happy we'll make an exception for `IOHelper` here: https://github.com/ontodev/robot/blob/master/robot-core/pom.xml#L56 After 1.2.0 we'll remove these exceptions.

PR #345, opened 2018-08-15
Improve `extract` docs, fixes #339

### LINKED ISSUE #339: Better explanation for "module"
Our explanation and links for SLME on <http://robot.obolibrary.org/extract> aren't very helpful. Can somebody who knows more about this than I do add a paragraph saying what "module" really means, and the difference between STAR, TOP, BOT.

This site has too much information: http://owl.cs.manchester.ac.uk/research/modularity/

jamesaoverton:


balhoff:
The expansion of MIREOT from the paper is slightly different: "MIREOT: The minimum information to reference an external ontology term".

rctauber:
This looks good! It makes a lot more sense than what we had before. 😃

PR #350, opened 2018-08-17
Report Updates

rctauber:
Changes:
* use labels instead of CURIEs when possible
* properly display values in output
* sort by value for duplicate values (e.g. definitions)
* add CSV format
* separate `ontology_metadata` check into 3 different queries (would only return if *all* were missing)


jamesaoverton:
After looking at the report run on OBI, I'm thinking that it's better to include both CURIEs and labels. There were several cases where I wanted to know which ontology a term came from.

rctauber:
Output now looks like:

Level | Rule Name | Subject | Property | Value
-- | -- | -- | -- | --
ERROR | duplicate_definition | eBioscience [OBI:0001213] | definition [IAO:0000115] | A supplier of flow cytometry analyzers
WARN | multiple_asserted _superclasses | experimentally modified cell in vitro   [CL:0000578] | rdfs:subClassOf | cell [CL:0000000]

Blank nodes are represented as: `blank node <0560eacbae19f8ae6a14f7d9e3a28d8e>` (although would it be better to say `anonymous`?)

jamesaoverton:
When I was testing this I forgot the `--output` and the result was just `null`:

```
$ java -jar build/robot.jar report --input obi.owl --labels true
null
Use the -vvv option to show the stack trace.
Use the --help option to see usage information.
```

rctauber:
It was a `NullPointerException` in `ReportCommand` - I wanted to auto detect the format based on the output, but if output is null that's a problem. So it should be fixed now!

Issue #349, opened 2018-08-16
Review signatures, add exceptions as required


jamesaoverton:
See #344. We want to maintain backwards compatibility as much as possible, but there are some mistakes that we should fix. It would be better to do this all at once for the 1.2.0 release (which will already break backwards compatibility in `QueryOperation` because of Jena updates #314).

rctauber:
I was looking at `DiffOperation.java` and it looks like the original implementation was supposed to include labels and short forms of IRIs (the `addLabels` method), but it doesn't execute properly.

Updating it changes the output of diff to look something like this:
```
+ AnnotationAssertion(<obo:IAO_0000117>[term editor] <obo:OBI_0000921>[labeled RNA extract] "Group: OBI group"^^xsd:string)
+ AnnotationAssertion(<obo:IAO_0000117>[term editor] <obo:OBI_0000924>[labeled specimen] "OBI group"^^xsd:string)
+ AnnotationAssertion(<obo:IAO_0000117>[term editor] <obo:OBI_0000925>[infectious agent] "IEDB"@en)
+ AnnotationAssertion(<obo:IAO_0000117>[term editor] <obo:OBI_0000930>[calorimeter] "PERSON:Bjoern Peters"^^xsd:string)
```

As opposed to the original:
```
+ AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000117> <http://purl.obolibrary.org/obo/OBI_0000921> "Group: OBI group"^^xsd:string)
+ AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000117> <http://purl.obolibrary.org/obo/OBI_0000924> "OBI group"^^xsd:string)
+ AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000117> <http://purl.obolibrary.org/obo/OBI_0000925> "IEDB"@en)
+ AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000117> <http://purl.obolibrary.org/obo/OBI_0000930> "PERSON:Bjoern Peters"^^xsd:string)
```

Is this something that should be fixed? I can always do it in a separate issue tracker/PR, but it does add exceptions to some of the methods in `DiffOperation.java`.

jamesaoverton:
I changed the default to not look up labels, because of performance problems: https://github.com/ontodev/robot/issues/47

I'd rather fix the docs than change this (non-label default) behaviour.

rctauber:
[...]

PR #351, opened 2018-08-21
Add new Exceptions, optimize codebase, see #349

rctauber:
**Exceptions**:
- QueryParseException no longer fails silently
- If an exception does not have a pre-configured exception message, the Exception message is printed, instead of failing with "use -vvv to see stack trace"
- If a report query does not include `?entity`, it cannot be properly parsed - fail with message
- If an invalid `--catalog-file` is passed in, fail instead of logging a warning
- Check if something is `null` instead of waiting for a `NullPointerException`

With all these changes, I changes the `breakOn` statements for japicmp to false. We still get the reports, but the build won't fail because of these. I can change this to just ignore changed classes, if you'd prefer.

```
<breakBuildOnBinaryIncompatibleModifications>false</breakBuildOnBinaryIncompatibleModifications
<breakBuildOnSourceIncompatibleModifications>false</breakBuildOnSourceIncompatibleModifications>
```

**Optimization**:
- Collection types are inferred (`Map<String, String> myNewMap = new HashMap<>();`)
- Removed use of Guava `Optional`
- Removed redundant class casts
- Removed `if` statements that were always true (e.g. `if (state != null)` when it has already been checked to be not-null previously)
- Changed lambda expressions to `forEach` with Consumer
- Simplified boolean return statements
- Replaced instances of `x.indexOf(y) < 0` with `!x.contains(y)`
- Handle imports as sets of `ImportDeclaration` objects instead of sets of `OWLOntology` objects

I'm sorry if I missed something in this list, but this should pretty much cover the changes. This is just a proposal, so I'm happy to revert any of these changes if they don't make sense.

In `ReduceOperation`, I also found the variable `exprs` (a set of OWLClassExpressions) that is updated but it's never used. I left a `TODO` comment to look into this, but this could potentially just be removed before we merge these changes in.

Finally, as we've changed how many of the operations work, we include a lot of methods that are no longer used. I know we maintain these for backwards compatibility, but it may be nice to mark them as `@Deprecated` at some point as a warning to people who are still using them (this might be a `v2.0.0` release milestone)... I'd love to get thoughts on this.

### DIFF
--- a/docs/query.md
+++ b/docs/query.md
@@ -48,6 +48,10 @@ The names of the graphs correspond to the ontology IRIs of the imports. If the i
 
 You must specify a query to execute with `--query` or `--queries`.
 
+### Query Parse Error
+
+The query was not able to be parsed. Often, this is as a result of an undefined prefix in the query. See the error message for more details.
+
 ### Query Type Error
 
 Each SPARQL query should be a SELECT, ASK, DESCRIBE, or CONSTRUCT.
--- a/docs/report.md
+++ b/docs/report.md
@@ -75,6 +75,14 @@ INFO	file:///absolute/path/to/other_query.rq
 
 ## Error Messages
 
+### Fail On Error
+
+Only `info`, `warn`, and `error` are valid inputs for `--fail-on`.
+
+### Missing Entity Binding
+
+All queries must bind `?entity ?property ?value` for correct formatting. If `?entity` is ever `null`, the query cannot be reported on.
+
 ### Report Level Error
 
 The logging level defined in a profile must be `ERROR`, `WARN`, or `INFO`.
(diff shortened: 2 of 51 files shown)


jamesaoverton:
I created an issue for deprecation: #353

rctauber:
Japicmp now excludes the modified classes, instead of ignoring all:
```
<!-- Temporarily exclude QueryOperation for Jena upgrade #314. -->
<exclude>org.obolibrary.robot.QueryOperation</exclude>
<!-- Exceptions added -->
<exclude>org.obolibrary.robot.IOHelper</exclude>
<exclude>org.obolibrary.robot.OntologyHelper</exclude>
<exclude>org.obolibrary.robot.ReportOperation</exclude>
<exclude>org.obolibrary.robot.ReasonerHelper</exclude>
<!-- Exceptions removed -->
<exclude>org.obolibrary.robot.CatalogElementHandler</exclude>
<exclude>org.obolibrary.robot.TemplateOperation</exclude>
```

Issue #227, opened 2018-01-15
Add SPARQL Update


jamesaoverton:
It would be good to be able to modify ontologies using SPARQL DELETE/INSERT. See #182.

rctauber:
Implemented by #352

PR #352, opened 2018-08-22
SPARQL Update

rctauber:
See #227

Supports SPARQL update with 
```
robot query --input <ontology> --update <file> --output <ontology>
```


rctauber:
@ontodev/robot-team I would appreciate some feedback on this PR:

When loading triples into Jena, `xsd:string` literals are treated the same as plain literals. Therefore, the datatype is removed from the string. When round-tripping back to OWLAPI, we end up with all the `xsd:string` statements removed.

I didn't realize this until I added the new check which does an update and then performs another update to revert the first one. The output should be exactly the same as the input, which it is... except the strings no longer have `^^xsd:string`.

Is this a problem? Will it cause issues for any pipelines? Any input would be appreciated.

cmungall:
If it doesn't fully roundtrip then it could cause issues. But I think the best path here is to document the behavior.

I don't think we should be using xsd:strings as much as we do; this was an early obo2owl decision. We may be able to change this but may take a while to propagate. However, I don't think it will cause any issues for people who maintain their source in obo, the xsd:strings will be reconstituted next roundtrip

jamesaoverton:
@rctauber have been using this for a while, and it really comes in handy. I'd like this to be the major feature in ROBOT 1.4.0.

The key problem is that Jena makes `xsd:string` implicit while OWLAPI does not. This only applies when you use Jena to render the output, so it doesn't affect backward compatibility. And I can't think of anything that ROBOT can do to change the behaviour. So I think the only solution is good documentation.

I'm going to merge this now. If you run into special problems, or have suggestions, please let us know soon, before we make the next release.

Issue #329, opened 2018-08-01
Create an option to dump a list of explanations for unsatisfiable classes


cmungall:
A complement to #174 

Analogous owltools option: `-e` or `run-reasoner` command. owltools dumps a flat list of axioms in the explanation rendering IRIs as IRI-label pairs.

For robot, it would be super-awesome to have something similar, but more user friendly

- [X] ability to render as markdown, for easy pasting into tickets (I have made probably 100s of github tickets with protege UI explanation screenshots, which are not very searchable)
 - [X] use of nesting, as in the Protege explanation UI
 - [x] option to render source ontology(s) of each axiom [v useful for debugging where an errant import is often to blame]

Protege will find all explanations but will allow you to interrupt its search. I personally think the choice of explanation is abitrary and would be happy for the tool to include just one. 

- [X] Maybe an option to find max N.

- [X] Bonus: in an ontology with multiple unsats, it can be useful to start with the one with the most minimal explanation, typically the MRCAs of all unsats in the asserted hierarchy. It may even be possible to combine explanations to find the most likely offending axiom. 

- [ ] Bonus bonus: plug in kboom to use probabilistic OWL inference to find the most likely overall resolution...

I believe @matentzn is working on something like this. I am neutral as to whether this is part of robot or a separate dependency on maven central that robot can include.

balhoff:
> ability to render as markdown, for easy pasting into tickets

This would be awesome.

> option to render source ontology(s) of each axiom

Would be nice as a `title` attribute to show as tooltip (not sure how to do that in markdown). Protege is supposed to show this but it seems to have broken a while back (the tooltips claim all the axioms have source "inferred").

rctauber:
Does the `explain` command meet this request?

balhoff:
I think this could reuse the functionality in ExplainOperation, but it would be a little more specific. It would need to get the list of unsatisfiable classes "`Xs`" and then ask for the explanation for each axiom of the form `X EquivalentTo owl:Nothing`. This could potentially be thousands of computations and very slow, so adding an algorithm to guess at the most general problem class would probably be important. So, I think there is more work to do before closing this.

cmungall:
not sure what you mean, I do this manually in Protege all the time, the first explanation is always fine

matentzn:
FYI The algorithm for computing the most general classes is called root unsatisfiable classes; https://www.sciencedirect.com/science/article/pii/S1570826805000260
[...]

PR #354, opened 2018-08-29
Implemented 'explain' command with Markdown output.

balhoff:
This is the beginning of an `explain` command. I'm opening a pull request to get feedback; it's not quite ready to merge. Example output:

`robot explain -I 'http://purl.obolibrary.org/obo/pato.owl' -r ELK --axiom "shape SubClassOf quality" -o shape.md`

## [shape](http://purl.obolibrary.org/obo/PATO_0000052) SubClassOf [quality](http://purl.obolibrary.org/obo/PATO_0000001) ##

  - [shape](http://purl.obolibrary.org/obo/PATO_0000052) SubClassOf [morphology](http://purl.obolibrary.org/obo/PATO_0000051)
    - [morphology](http://purl.obolibrary.org/obo/PATO_0000051) SubClassOf [physical object quality](http://purl.obolibrary.org/obo/PATO_0001241)
      - [physical object quality](http://purl.obolibrary.org/obo/PATO_0001241) SubClassOf [quality](http://purl.obolibrary.org/obo/PATO_0000001)

`robot explain -I 'http://purl.obolibrary.org/obo/uberon/ext.owl' -r ELK --axiom "'uvular muscle' SubClassOf 'muscle organ' and 'part of' some 'jaw region'" -o uvular_muscle.md`

## [uvular muscle](http://purl.obolibrary.org/obo/UBERON_0010235) SubClassOf [muscle organ](http://purl.obolibrary.org/obo/UBERON_0001630) and ([part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [jaw region](http://purl.obolibrary.org/obo/UBERON_0011595)) ##

  - [uvular muscle](http://purl.obolibrary.org/obo/UBERON_0010235) SubClassOf [palatal muscle](http://purl.obolibrary.org/obo/UBERON_0003682)
    - [palatal muscle](http://purl.obolibrary.org/obo/UBERON_0003682) EquivalentTo [skeletal muscle organ](http://purl.obolibrary.org/obo/UBERON_0014892) and ([part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [soft palate](http://purl.obolibrary.org/obo/UBERON_0001733))
      - [soft palate](http://purl.obolibrary.org/obo/UBERON_0001733) SubClassOf [part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [secondary palate](http://purl.obolibrary.org/obo/UBERON_0001716)
        - [secondary palate](http://purl.obolibrary.org/obo/UBERON_0001716) SubClassOf [part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [upper jaw region](http://purl.obolibrary.org/obo/UBERON_0001709)
          - [upper jaw region](http://purl.obolibrary.org/obo/UBERON_0001709) SubClassOf [part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [jaw region](http://purl.obolibrary.org/obo/UBERON_0011595)
      -  Transitive: [part_of](http://purl.obolibrary.org/obo/BFO_0000050)
      - [skeletal muscle organ](http://purl.obolibrary.org/obo/UBERON_0014892) EquivalentTo [muscle organ](http://purl.obolibrary.org/obo/UBERON_0001630) and ([synapsed by](http://purl.obolibrary.org/obo/RO_0002103) some [motor neuron](http://purl.obolibrary.org/obo/CL_0000100)) and ([develops from](http://purl.obolibrary.org/obo/RO_0002202) some [myotome](http://purl.obolibrary.org/obo/UBERON_0003082)) and ([surrounded by](http://purl.obolibrary.org/obo/RO_0002219) some [epimysium](http://purl.obolibrary.org/obo/UBERON_0011899)) and ([composed primarily of](http://purl.obolibrary.org/obo/RO_0002473) some [skeletal muscle tissue](http://purl.obolibrary.org/obo/UBERON_0001134))


balhoff:
I copied a class out of Protege in order to sort and indent the axioms (https://github.com/protegeproject/explanation-workbench/blob/master/src/main/java/uk/ac/manchester/cs/owl/explanation/ProtegeExplanationOrderer.java). There is an ordering class included in OWL API, but it seems to have some bugs which cause it to fail with particular inputs. If we depend on the explanation-workbench directly, I think it will bring in all of Protege as a dependency, which doesn't seem desirable. Any issue with this?

matentzn:
Ah great Jim! Love it. It would be nice to also get the InconsistentOntologyExplanation into this command. I wouldn't worry too much about the ordering code at the moment - this can be worked on once we have a simple printout of all the axioms in place; I would also like to be able to dump out the explanation as a little ontology so we can open it in Protege.

jamesaoverton:
Very nice Jim! It would be great to have this in ROBOT.

Thinking ahead to how we'll document this: Do you have a use case where this is better/easier than using Protégé for explanations?

That Protégé class a larger chunk than we usually borrow from other projects. I couldn't find the original file in a quick shirt -- can you point me to it?

Protégé is also BSD licensed, and we can always ask Matt for his blessing. Does anybody have any thoughts or concerns about how we should go about including it in ROBOT?

balhoff:
> I would also like to be able to dump out the explanation as a little ontology so we can open it in Protege

@matentzn I had thought about making this possible. Ideally this would just provide an ontology that could be carried forward into further `robot` commands. So it seems like you need to be able to provide a separate argument to save the explanation text vs. the ontology. So maybe something like this (see `--dump` and `-o`:

`robot explain -I 'http://purl.obolibrary.org/obo/pato.owl' -r ELK --axiom "shape SubClassOf quality" --dump shape.md -o shape.ofn`
 
You could then do this (pass to `annotate`):

`robot explain -I 'http://purl.obolibrary.org/obo/pato.owl' -r ELK --axiom "shape SubClassOf quality" --dump shape.md annotate --ontology-iri 'http://example.org/my_explanation' -o shape.ofn`

@jamesaoverton what do you think about this interface?

balhoff:
@jamesaoverton the link to the original file was in my comment: 
https://github.com/protegeproject/explanation-workbench/blob/master/src/main/java/uk/ac/manchester/cs/owl/explanation/ProtegeExplanationOrderer.java

> Do you have a use case where this is better/easier than using Protégé for explanations?

My main motivation is to be able to easily obtain explanations as Markdown, for pasting into GitHub tickets. @cmungall also wanted to add an option to the `reason` command to dump an explanation when there are unsatisfiable classes (see #329). We can call into ExplainOperation for this.

jamesaoverton:
@balhoff `--dump` makes sense. I don't love the name. The `template` command has `--merge-before` and `--merge-after` options to do something similar, but I don't like those names any better. http://robot.obolibrary.org/template#merging

balhoff:
And I added [a documentation page](https://github.com/ontodev/robot/blob/a4f1fda8bb5f11029706c066e99ac3b17f1ac83c/docs/explain.md).

@jamesaoverton @rctauber can we merge? :smiley:

jamesaoverton:
Thanks @balhoff. Becky and I have been busy with some deadlines. We'll look at this as soon as we can!

docs/explain.md
```diff
@@ -0,0 +1,46 @@
+# Explain
+
+When reviewing inferred axioms, such as after applying the [`reason`](/reason) command,
+you may like more information about why a particular axiom was inferred:
+
+    robot explain -i uvula.ofn -r ELK --axiom "'uvular muscle' SubClassOf 'muscle organ' and 'part of' some 'jaw region'" -o uvular_muscle.md
```
rctauber: It would look better on the doc page if this was split as so:
```
robot explain --input uvula.ofn --reasoner ELK \
  --axiom "'uvular muscle' SubClassOf 'muscle organ' and 'part of' some 'jaw region'" \
  --output uvular_muscle.md
```

Would it be beneficial to also have an `--axiom-file` input option to provide a text file for the axiom? Multiple lines for multiple axioms that you'd like to explain?

docs/explain.md
```diff
@@ -0,0 +1,46 @@
...
+
+Here we provide an axiom in Manchester syntax. If that relationship is entailed by the 
+axioms asserted in [uvula.ofn](/examples/uvula.ofn), we can use the selected reasoner (here, ELK)
+to compute an _explanation_—a minimal set of axioms required to derive that inference. Reviewing 
+inference explanations is an extremely useful tool for debugging ontology classification issues.  
+The explanation is printed in Markdown format, which can be easily pasted and rendered within 
+GitHub issues:
+
+    ## [uvular muscle](http://purl.obolibrary.org/obo/UBERON_0010235) SubClassOf [muscle organ](http://purl.obolibrary.org/obo/UBERON_0001630) and ([part of](http://purl.obolibrary.org/obo/BFO_0000050) some [jaw region](http://purl.obolibrary.org/obo/UBERON_0011595)) ##
```
rctauber: I don't know if we need both the markdown code and the rendered markdown... I think the rendered markdown would be sufficient for the docs. @jamesaoverton - what do you think?

docs/explain.md
```diff
@@ -0,0 +1,46 @@
...
+      -  Transitive: [part of](http://purl.obolibrary.org/obo/BFO_0000050)
+      - [skeletal muscle organ](http://purl.obolibrary.org/obo/UBERON_0014892) EquivalentTo [muscle organ](http://purl.obolibrary.org/obo/UBERON_0001630) and ([develops_from](http://purl.obolibrary.org/obo/RO_0002202) some [myotome](http://purl.obolibrary.org/obo/UBERON_0003082)) and ([surrounded_by](http://purl.obolibrary.org/obo/RO_0002219) some [epimysium](http://purl.obolibrary.org/obo/UBERON_0011899)) and ([composed primarily of](http://purl.obolibrary.org/obo/RO_0002473) some [skeletal muscle tissue](http://purl.obolibrary.org/obo/UBERON_0001134)) and ([synapsed by](http://purl.obolibrary.org/obo/uberon/core#synapsed_by) some [motor neuron](http://purl.obolibrary.org/obo/CL_0000100))
+
+In addition to outputting the explanation axioms in Markdown, an ontology containing just 
+those axioms is provided to any subsequent chained commands:
+
+    robot explain -i uvula.ofn -r ELK --axiom "'uvular muscle' SubClassOf 'muscle organ' and 'part of' some 'jaw region'" convert -o uvular_muscle.ofn
```
rctauber:
```
robot explain --input uvula.ofn --reasoner ELK \
  --axiom "'uvular muscle' SubClassOf 'muscle organ' and 'part of' some 'jaw region'" \
  convert --output uvular_muscle.ofn
```

docs/explain.md
```diff
@@ -0,0 +1,46 @@
...
+
+You can still output a Markdown file, if desired, before the chained command.
```
rctauber: An example would be good (it wouldn't have to be part of the test cases if you surround it with ticks instead of the indentation).

robot-command/src/main/java/org/obolibrary/robot/ExplainCommand.java
```diff
@@ -0,0 +1,146 @@
...
+  private Options options;
+
+  public ExplainCommand() {
+    Options o = CommandLineHelper.getCommonOptions();
+    o.addOption("i", "input", true, "load ontology from a file");
+    o.addOption("I", "input-iri", true, "load ontology from an IRI");
+    o.addOption("r", "reasoner", true, "reasoner to use: ELK, HermiT, JFact");
+    o.addOption("a", "axiom", true, "the axiom to explain");
+    o.addOption("m", "max", true, "the maximum number of explanations to retrieve");
+    o.addOption("f", "format", true, "the explanation result format: MD, OFN, etc.");
```
rctauber: Could you add info in the docs on the valid output formats?

robot-command/src/main/java/org/obolibrary/robot/ExplainCommand.java
```diff
@@ -0,0 +1,146 @@
...
+    o.addOption("o", "output", true, "save explanation to a file");
```
rctauber: Just a thought...

Instead of having to chain to `convert` after using `explain`, the `--output` option could output the ontology just containing explanation options in any valid ontology format. And you could have an `--explanation` option for the markdown or text file, e.g.:
```
o.addOption("o", "output", true, "save explanation axioms to an ontology");
o.addOption("e", "explanation", true, "save explanation to a file");
```
For the other ROBOT commands, the `--output` is an ontology. For example, `query` does not use the `--output` flag for the generated results.

robot-command/src/main/java/org/obolibrary/robot/ExplainCommand.java
```diff
@@ -0,0 +1,146 @@
...
+    state = CommandLineHelper.updateInputOntology(ioHelper, state, line);
+    OWLOntology ontology = state.getOntology();
+    QuotedEntityChecker checker = new QuotedEntityChecker();
+    checker.setIOHelper(ioHelper);
+    checker.addProvider(new SimpleShortFormProvider());
+    checker.addProperty(OWLManager.getOWLDataFactory().getRDFSLabel());
+    if (ontology != null) {
+      checker.addAll(ontology);
+    }
+    int max = 1;
```
rctauber: This could be a bit more robust for catching errors. We also have a `getDefaultValue` method.
```
String maxString = CommandLineHelper.getDefaultValue(line, "max", "1");
int max;
try {
  max = Integer.parseInt(maxString);
} catch (NumberFormatException e) {
  throw new Exception(String.format(maxTypeError, maxString));
}
```

Where the error messages at the top are something like...
```
private static final String NS = "explain#";
private static final String maxTypeError = NS + "MAX TYPE ERROR --max ('%s') must be an integer"
```

This allows our `ExceptionHelper` to parse the error message and display a nice output in the console. See error message examples in our current docs: http://robot.obolibrary.org/convert#check-arg-error

robot-command/src/main/java/org/obolibrary/robot/ExplainCommand.java
```diff
@@ -0,0 +1,146 @@
...
+      state = new CommandState();
+    }
+    IOHelper ioHelper = CommandLineHelper.getIOHelper(line);
+    state = CommandLineHelper.updateInputOntology(ioHelper, state, line);
+    OWLOntology ontology = state.getOntology();
+    QuotedEntityChecker checker = new QuotedEntityChecker();
+    checker.setIOHelper(ioHelper);
+    checker.addProvider(new SimpleShortFormProvider());
+    checker.addProperty(OWLManager.getOWLDataFactory().getRDFSLabel());
+    if (ontology != null) {
```
rctauber: Minor thing - this check is unnecessary; `ontology` will not be null as `CommandLineHelper.updateInputOntology(...)` throws an exception if `--input` is empty.

robot-command/src/main/java/org/obolibrary/robot/ExplainCommand.java
```diff
@@ -0,0 +1,146 @@
...
+      checker.addAll(ontology);
+    }
+    int max = 1;
+    if (line.hasOption("max")) {
+      max = Integer.parseInt(line.getOptionValue("max"));
+    }
+    OWLReasonerFactory reasonerFactory = CommandLineHelper.getReasonerFactory(line, true);
+    ManchesterOWLSyntaxInlineAxiomParser parser =
+        new ManchesterOWLSyntaxInlineAxiomParser(OWLManager.getOWLDataFactory(), checker);
+    String expression = line.getOptionValue("axiom");
```
rctauber: If you use the `CommandLineHelper` method for this, it will throw an error if the `--axiom` is missing (error message being the third arg).
```
String expression = CommandLineHelper.getRequiredValue(line, "axiom" "an --axiom is required");
```

robot-command/src/main/java/org/obolibrary/robot/ExplainCommand.java
```diff
@@ -0,0 +1,146 @@
...
+    OWLAxiom axiom = parser.parse(expression);
+
+    Set<Explanation<OWLAxiom>> explanations =
+        ExplainOperation.explain(axiom, ontology, reasonerFactory, max);
+
+    if (line.hasOption("output")) {
```
rctauber: Could you output the explanations as plain text to the console if no `--output` is provided, or should this be required? Alternatively, this could be `--explanations` instead and the output could be the subset of the ontology created below.

robot-core/src/main/java/org/obolibrary/robot/ExplainOperation.java
```diff
@@ -0,0 +1,103 @@
...
+import org.semanticweb.owlapi.util.AnnotationValueShortFormProvider;
+import org.semanticweb.owlapi.util.ShortFormProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.ac.manchester.cs.owl.explanation.ProtegeExplanationOrderer;
+import uk.ac.manchester.cs.owl.explanation.ordering.ExplanationOrderer;
+import uk.ac.manchester.cs.owl.explanation.ordering.ExplanationTree;
+import uk.ac.manchester.cs.owl.explanation.ordering.Tree;
+
+public class ExplainOperation {
```
rctauber: If you get a change, could you add javadoc comments for this class & its methods?

matentzn:
For future, two suggestions. 
1. It would perhaps be good to be able to flip the url target with an option such as --urltarget "ols" so you dont have to rely on the automatic purl resolution.
2. Formatting explanations in this way is hard; and I think we might tweak it here and there for the more difficult cases. I particularly do not like the fact (not with your approach, but with justification displays in general) that the nesting in the explanation looks like a class hierarchy, but of course turns out to be not. It would be so cool to have something slightly more visual, an ASCII tree or something displaying subsumptions and existential restrictions plus a, hopefully super short, list of additional axioms (disjointness and characteristics). What can also be very helpful is additional markdown: For example, unsatisfiable classes could be marked up with bold or italic. Nothing to act on now, just for the future as a suggestion :)

robot-command/src/main/java/org/obolibrary/robot/ExplainCommand.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  @Override
+  public String getDescription() {
+    return "explain derivation of an inferred axiom";
+  }
+
+  @Override
+  public String getUsage() {
+    return "robot explain --input <file> --reasoner <reasoner> --axiom <axiom> <output>";
```
rctauber:
```
"robot explain --input <file> --axiom <axiom> --output <output>"
```
Default reasoner should is ELK (returned by `CommandLineHelper.getReasonerFactory(...)`) so the user wouldn't need to specify that. Maybe make a note in the docs? And if you link to the `reason` docs, they can see all available reasoners.

matentzn:
+1 on the Beckys --explanation suggestion.

robot-command/src/main/java/org/obolibrary/robot/ExplainCommand.java
balhoff: This is kind of a mistake—I didn't meant to combine Markdown and OWL here (I think this line is from when I first started working on this). I will update to reflect the fact that the format is just the ordinary format choice all the other commands use.

robot-command/src/main/java/org/obolibrary/robot/ExplainCommand.java
balhoff: Agree.

balhoff:
@rctauber I agree with `--explanation`—I think that will eliminate a potential source of confusion. I will update.

docs/explain.md
balhoff: Re: `--axiom-file`, I would be fine with this, but don't have a need for it now. We could always add it down the road in a backward-compatible way.

docs/explain.md
balhoff: I removed the plaintext version.

balhoff:
Thanks for the great feedback @rctauber. I think I've addressed all your comments.

jamesaoverton:
I'm happy with this.

I'm trying to get a release out the door, so I'll leave this as a PR a bit longer until I figure that out.

Issue #185, opened 2017-08-29
Consider excluding assertions when extracting a module


jamesaoverton:
When using SLME to extract a module from an ontology that includes individuals, all the individuals and their parent classes are included in the result. More generally (correct me if I'm phrasing this wrong), all the ABox axioms and everything that they depend on from the TBox are included in the extracted module. See <https://github.com/owlcs/owlapi/issues/462>.

Often this is not what you want. In response to that issue, an `excludeAssertions` option was added to SLME (available in OWLAPI 4.3.2): https://github.com/owlcs/owlapi/blob/version4/tools/src/main/java/uk/ac/manchester/cs/owlapi/modularity/SyntacticLocalityModuleExtractor.java#L242

Setting `excludeAssertions` to `true` means that all ABox axioms are stripped before extraction, and so nothing from the ABox is included in the result. Often this is not what you want, either.

When extracting OBI Core I would like to end up with all-and-only those individuals that belong to the classes in the extracted module. If I `excludeAssertions` then run this code, I get what I want:

```
// Restore individual axioms for those individuals still in the outputOntology.
Set<OWLNamedIndividual> individuals = outputOntology.getIndividualsInSignature(true);
for (OWLNamedIndividual individual: individuals) {
    Set<OWLIndividualAxiom> individualAxioms = inputOntology.getAxioms(individual);
    outputOntology.getOWLOntologyManager().addAxioms(outputOntology, individualAxioms);
}
```

I think this is a reasonable option to have. Am I crazy?

jamesaoverton:
Correction/clarification: For OBI Core I want to include all-and-only those individuals that are used to define classes. Case in point, 'curation status specification' <http://purl.obolibrary.org/obo/IAO_0000078> is defined as equivalent to the set of individual specifications. See also 'obsolescence reason specification' <http://purl.obolibrary.org/obo/IAO_0000225>.

These might be extreme/weird cases. If so, we can consider making some changes to IAO.

rctauber:
Did we determine if `filter` was able to accomplish this?

jamesaoverton:
I'm pretty sure that `filter` can't do it. You need to keep the individuals available but tell SLME to ignore them, then copy the individuals into the extracted module.

rctauber:
Not quite sure what to call this option, but I was thinking:

`--assertions include` all ABox axioms and everything they depend on from TBox (default?)
`--assertions minimal` only the ABox axioms that depend on something in the TBox axioms
[...]

PR #355, opened 2018-09-03
Update to OWLAPI 4.5.5

rctauber:
See #308

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -166,7 +166,7 @@
     <dependency>
       <groupId>net.sourceforge.owlapi</groupId>
       <artifactId>owlapi-distribution</artifactId>
-      <version>4.2.6</version>
+      <version>4.5.5</version>
     </dependency>
     <dependency>
       <groupId>org.semanticweb.elk</groupId>


jamesaoverton:
@cmungall This should fix #308 and enables a feature I would like #185.

Protégé is still on 4.2.8 https://github.com/protegeproject/protege/blob/master/pom.xml#L131

Do you think we should update?

cmungall:
cc @matthewhorridge, I made a PR on the Protege repo but I'm guessing there are other considerations besides passing travis, all plugins need to be compatible.

I feel that there are some subtle obo format reordering changes going between 4.2.x and 4.3.x and above, when there are multiple axiom annotations. I will investigate.

cmungall:
Thanks everyone. I'm guessing this won't make it into Protege 5.5?

Even so I think it's OK for you to go ahead if you need this @rctauber, on projects that need parity between owlapi versions, I think we can manage this by pinning to the current robot release/docker for an interim period.

matthewhorridge:
> I'm guessing this won't make it into Protege 5.5?

It will.  Still in the beta period and, although this is a major change, I think it's important.

matthewhorridge:
We'll do a few more tests here and then I'll merge in the owlapi-4.5.5 branch

jamesaoverton:
Protege moved to 4.5.6 (https://github.com/protegeproject/protege/issues/800), so let's match that.

rctauber:
Switched to v4.5.6

Issue #263, opened 2018-03-02
support for slims


simonjupp:
Has there been any discussion about how you might use ROBOT to extract a slim? I'd like something like 

`robot subset -t terms.txt` where the terms.txt file contains a list of IRIs for terms in the slim. I wouldn't want this tied to the OBO subset pattern as you could easily do a query operation first to generate the terms file from some APs in the ontology.

rctauber:
We are working on `remove` and `filter` commands - check out #272. I think `filter` will do exactly what you need.

simonjupp:
I'm not sure filter is enough to extract a slim. The owl-tools --extract-ontology-subset will fill in subclassOf axioms between terms in the slim which is a crucial part of the slimming process. 

So if I have a simple ontology with a hierarchy like
```
A
  - B
      - C
```
If A and C are in a subset and I try and extract it I should get a new ontology with 
```
A
  - C
```

It's not clear if the `filter` command would implement this.

jamesaoverton:
@simonjupp You're right that the current `filter` proposal doesn't cover that. We'll figure something out.

simonjupp:
I think this issue duplicates #302 and #254

rctauber:
I think the best way to handle this is through `filter`. I made some updates on a new branch: https://github.com/rctauber/robot/tree/gaps

This includes a new `--preserve-structure <true/false>` option. By default, this is true. Gaps are automatically spanned to maintain the structure in both `remove` and `filter`. If you don't want to maintain that structure, you can include `--preserve-structure false`.

So, going off your above example, if your input term set includes classes `A` and `C`, you will get:
```
owl:Thing
  - A
    - C
```

But with `--preserve-structure false`, you would get:
```
owl:Thing
  - A
  - C
```

Filling in the entire hierarchy (fill gaps, instead of span gaps) can be achieved with a two-step filter:
```
robot filter --input foo.owl \
  --term-file subset_terms.txt \
  --select "self ancestors annotations" \
filter --term-file subset_terms.txt \
  --select "self descendants annotations" \
  --output bar.owl
```

@simonjupp - I would appreciate if you could test this out and let me know if it works for your use case!

cmungall:
does spanning only cover named classes?

E.g. if I wanted to make a slim with just {nucleus, cell} from

 * nucleus subClassOf 'cell part'
 * 'cell part' subClassOf part_of some cell

Would there be a gap?

[...]

PR #357, opened 2018-09-05
Fill gaps in filter/remove

rctauber:
See #263


jamesaoverton:
Does this have any tests? It's not clear to me whether `remove_class.owl` is being used for an integration test since it seems like a new file, but it's mentioned in the previous `remove.md`.

rctauber:
That shouldn't be in there - sorry - it was part of the `results`. I'll swap that out and add an integration test for this.

jamesaoverton:
@rctauber Can you please resolve this merge conflict?

rctauber:
Do we want to change the default behavior of `--trim` to `true`? I know we talked about it a bit but I wasn't sure if I should implement that in this branch.

jamesaoverton:
Let's do that in a new PR, after this is merged.

jamesaoverton:
@ontodev/robot-team and maybe @balhoff and @matentzn in particular: This is an important improvement to remove/filter, but I'm worried that it will be a big performance hit. Can you please test this branch on any remove/filter tasks that you have set up?

I'd like to push a new release soon, so I'd really appreciate some quick feedback. Thanks!

jamesaoverton:
@ontodev/robot-team One more ping on this PR. This is the last big thing we're waiting on for 1.2.0, so I'd really appreciate some quick feedback.

balhoff:
I'm testing one of my targets.

balhoff:
On the targets where I use filter, I don't see any performance impact. However I don't think the options I use require any gap filling.

Issue #361, opened 2018-09-18
Add diff option to force output of axioms with labels


dosumis:
Diffs that present logical axioms with IRIs are hard to interpret.  Given that ROBOT is an OBO tool and that OBO ontologies lack readable short_forms, diff output would be much interpret it presented axioms rendered using labels.  IIRC, the OWLAPI Manchester Syntax renderer can already produce MS axioms with labels.  Not sure if there is a Functional Syntax alternative.

jamesaoverton:
Yes, there should be an option for labels.

The history of labels in diffs is a bit confusing, unfortunately. See #47, #351.

rctauber:
I came up with a couple of options for the format... which do you think would work best?

1. `- SubClassOf(<GO:0051674>[localization of cell] <GO:0051179>[localization])`
2. `- SubClassOf([localization of cell]<GO:0051674> [localization]<GO:0051179>)`
3. `- SubClassOf('localization of cell'[GO:0051674] localization[GO:0051179])`
4. `- SubClassOf('localization of cell'<GO:0051674> localization<GO:0051179>)`

jamesaoverton:
I think we used (1) before it mysteriously broke.

dosumis:
Should avoid single quote as some labels have them.  Happy with the other two.

rctauber:
Closed by #363 - please reopen if this needs more discussion

PR #363, opened 2018-09-18
Add labels option to diff

rctauber:
See #361


jamesaoverton:
Looks good. I'd like a test, please.

rctauber:
I changed `--use-labels` to `--labels` to match what we do with `report`. I can change it back if you prefer the first version, though.

jamesaoverton:
Yeah, I think `--labels` is better.

jamesaoverton:
Closes #361

Issue #365, opened 2018-09-19
'remove' not working


balhoff:
I can't seem to get `remove` to work. I must be using it incorrectly:

`robot remove -I 'http://purl.obolibrary.org/obo/uberon/uberon-base.owl' --term 'http://purl.obolibrary.org/obo/UBERON_6110811' -o uberon-filtered.ttl`

This doesn't seem to remove `http://purl.obolibrary.org/obo/UBERON_6110811` at all. What am I missing?

rctauber:
Try adding `--trim true`. By default, this option is false. So it only removes axioms in which *all* objects of the axiom are in the removal set. We should probably expand on this in the docs...

balhoff:
That's it, thanks! I didn't get that from the phrasing in the docs.

PR #366, opened 2018-09-19
Clarify --trim behavior

rctauber:


jamesaoverton:
I dunno. It still needs work.

1. All the examples that are integration tests include `--trim true`. It's suspicious that we don't have a `--trim false` example.
2. The OBI example (which is not an integration test) does not include `--trim`, and I bet it won't work as described.
3. We're treating "term" and "entity" as synonymous, which bugs me.

rctauber:
I always end up using `remove --trim true`, never `--trim false`. I think it's useful, but I'm wondering if we should switch the default to `true` so we don't end up with unexpected behavior like in #365?

I can change references to entity to "term" instead.

rctauber:
Do you have any ideas for a `--trim false` example? I can't think of a simple one to demonstrate it.

PR #371, opened 2018-09-25
Add support for .gz ontologies

rctauber:


robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
```diff
@@ -335,34 +360,40 @@ public OWLOntology loadOntology(InputStream ontologyStream) throws IOException {
...
     try {
       OWLOntologyManager manager = OWLManager.createOWLOntologyManager();
-      manager.setIRIMappers(Sets.newHashSet(new CatalogXmlIRIMapper(catalogFile)));
-      ontology = manager.loadOntologyFromOntologyDocument(ontologyIRI);
+      // If a catalog file was loaded, set IRI mappers
+      if (catalogFile != null) {
+        manager.setIRIMappers(Sets.newHashSet(new CatalogXmlIRIMapper(catalogFile)));
+      }
+      // Maybe load a zipped ontology
+      if (ontologyIRI.toString().endsWith("gz")) {
```
jamesaoverton: I think this should look for `.gz`

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
```diff
@@ -524,39 +559,19 @@ public OWLOntology saveOntology(
...
-      oboWriter.write(oboOntology, bw);
-    } else {
-      // use native save functionality
-      try {
-        ontology.getOWLOntologyManager().saveOntology(ontology, format, ontologyIRI);
-      } catch (OWLOntologyStorageException e) {
-        throw new IOException(String.format(ontologyStorageError, ontologyIRI.toString()), e);
-      }
+    // If saving in compressed format, get byte data then save to gzip
+    if (ontologyIRI.toString().endsWith("gz")) {
```
jamesaoverton: Should be `.gz`

jamesaoverton:
I took a quick look. I'd like some sort of automated test, please.

cmungall:
Will gz via http be supported?

rctauber:
@cmungall - the current code includes support for `--input-iri` gz files
@jamesaoverton - I'll add a test for this.

jamesaoverton:
This still needs a unit test, and there are two small changes that I commented on.

Please clarify in the docs that you can also save to `.gz`.

rctauber:
Test added. There's info in `convert.md`, but I wasn't sure where else to put it. Any ideas?

jamesaoverton:
Ok, the docs are fine.

I think the file extension checks should be `.endsWith(".gz")` throughout and not just `"gz"`. There are variants such as `".tgz"` that might not work properly with the current `"gz"` check.

rctauber:
Makes sense - changed to `.gz`.

PR #377, opened 2018-10-12
Axioms

Shalsh23:
Added functionality to extract abox, tbox, and rbox axioms using ` --axioms` option. This is the option to specify the axioms to filter/remove with the `filter` and `remove` commands respectively.

### DIFF
--- a/docs/remove.md
+++ b/docs/remove.md
@@ -77,9 +77,10 @@ The `--axioms` option allows you to specify the type of OWLAxiom to remove. More
 6. `equivalent` (classes and properties)
 7. `disjoint` (classes and properties)
 8. `type` (class assertions)
-9. `abox`
-10. `tbox`
-11. `rbox`
+9. `tbox` (classes and properties)
+10. `abox` (individuals)
+11. `rbox` (properties)
+
 ## Examples
 
 Remove a class ('organ') and and all its descendants:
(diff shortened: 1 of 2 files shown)


balhoff:
Thanks, this is going to be very useful! 

@jamesaoverton @rctauber: @Shalsh23 is working with me at RENCI.

Shalsh23:
Thanks. I have updated the documentation as well.

docs/remove.md
```diff
@@ -77,9 +77,10 @@ The `--axioms` option allows you to specify the type of OWLAxiom to remove. More
 6. `equivalent` (classes and properties)
 7. `disjoint` (classes and properties)
 8. `type` (class assertions)
-9. `abox`
-10. `tbox`
-11. `rbox`
+9. `tbox` (classes and properties)
```
balhoff: @jamesaoverton in the OWL API definition of `tbox`, I don't think you get any property axioms.

jamesaoverton: Ok. I was looking here https://github.com/owlcs/owlapi/blob/version4/api/src/main/java/org/semanticweb/owlapi/model/AxiomType.java#L314 which has a bunch of property stuff, but I guess that's just what's needed for class expressions.

balhoff: Oh I see, that make sense. It looks like it includes interactions between classes and properties (domain/range). I just noticed this because a while back I thought I was getting the property hierarchy when I was asking for TBOX, and it wasn't working until I realized they had an RBOX constant as well.

jamesaoverton: I'm happy to use whatever description you like. I just wanted a bit of a description for these options, since I've never been comfortable with these terms, and I expect there are other people like me out there.

balhoff: I don't have strong feelings. If we found an informative article to link to that would be nice. I checked at Ontogenesis, but I didn't see anything.

Issue #372, opened 2018-09-25
Document ROBOT_JAVA_ARGS


balhoff:
I couldn't find a reference to the `ROBOT_JAVA_ARGS` environment variable anywhere in the documentation. I nearly always need to use this, to increase the JVM heap size, so I expect it will be a common requirement.

cc @Shalsh23

rctauber:
Hi @balhoff - please see #378

balhoff:
Looks good!

PR #378, opened 2018-10-13
Add Java Options to docs

rctauber:
See #372

Also add Java options for the Windows script. I don't have access to a Windows machine, so it would be good if someone could verify that `robot.bat` works with the environment variable `ROBOT_JAVA_ARGS`. Thanks!


jamesaoverton:
I guess we'll find out about Windows when someone complains.

PR #385, opened 2018-10-24
Exclude assertions

rctauber:
See #185


rctauber:
`mvn verify` passes - could you point me to what you were worried about?

jamesaoverton:
Locally I tried copying the latest `robot-core/pom.xml` (with JAPICMP configuration updates) to this branch and running `mvn verify`. It failed with some JAPICMP compatility problems.

rctauber:
OK, I went back to ee2aec6 and it failed, but if you merge master into this branch, it passes... I think it's because new stuff got added in for `1.2.0`, and this branch just wasn't up-to-date so it thinks stuff got removed.

jamesaoverton:
For OBI I just want the individuals that are used to define terms such as ‘curation status specification’. Could we add an option for `--individuals definitions` that includes only the individuals used in logical axioms?

jamesaoverton:
I'm happy with the feature and the code. Please resolve the conflicts and add some sort of unit test.

Issue #390, opened 2018-10-26
Add option to extract command to stamp classes with rdfs:isDefinedBy


cmungall:
If selected, each class in a generated module would have an isDefinedBy annotation to the URI of the source ontology, unless isDefinedBy is already present.

balhoff:
It seems to me that for `extract` it would make more sense to have the subject be the extracted axiom rather than the class.

In some of my code I do have a utility that adds isDefinedBy annotations terms that have OBO-style IRIs to the canonical ontology IRI for that OBO namespace (much faster to query for them this way than filtering on IRI text).

cmungall:
It would be good to support any combination of C+A.

A rough start at some conventions here:

- on the class means the URI is managed by the group. By default, they 
should be group that assert axioms with the class as subject
- on the axioms means the axiom is managed by the group. usually this is 
the same group that 'owns' the class URI, but in some cases axiom 
injection from outside is encouraged

Note that 'subject' doesn't make sense at the OWL level for some axiom 
types like DisjointClasses, EquivalentClasses between two named classes.

rctauber:
We now have this feature with #392. Is this OK to close or does it require more discussion?

cmungall:
Yes we can add new tickets for enhanced functionality later

PR #392, opened 2018-10-31
Add option to annotate extracted entities with source

rctauber:
See #390 

`--annotate-with-source true` will add `rdfs:isDefinedBy` to the term. Does not include annotations on the axiom, but that can be added as well. I'm not sure if this should be a different option though?

`--sources <arg>` allows you to define a term-to-source map file (as CSV or TSV) to redirect adopted entities, e.g.:
```
term,source
BFO:0000050,RO
```

If the term to extract isn't in that mapping file, the source is created from the term's IRI.

You can also do full IRIs for the sources (or terms) in the map.
```
term,source
http://purl.obolibrary.org/obo/BFO_0000050,http://purl.obolibrary.org/obo/ro.owl
```


rctauber:
@cmungall - I forgot to add in a check for if the `rdfs:isDefinedBy` annotation already exists. I just updated the code with that.

I know there has been some talk surrounding term "adoption" recently, especially with the OBO Core work. As this is something that impacts the OBO Foundry on a wider scale, this definitely requires more thought & input, but I wanted to have some sort of mock-up to demonstrate how it would work within ROBOT. Eventually, ROBOT could potentially have a default term-to-source map, like the prefix map, which would work behind the scenes. Again, these are just some first thoughts.

jamesaoverton:
@cmungall I'd like to get this PR merged. The core functionality to add `isDefinedBy` is something we should have had from the beginning. The override feature is nice-to-have, regardless of what we settle on as an adoption stretegy, since we already have a few OBO terms that have moved from BFO to RO.

cmungall:
agreed, merge away, conflict looks simple to fix

Issue #391, opened 2018-10-30
Improvements to `report` command


dougli1sqrd:
Hello,
I'm attempting to replace the set of robot commands and queries in the geneontology/go-ontology repo that we use to validate changes to the ontology in travis checks and the like with `robot report`. These are run on the go-edit.obo file in that repository. go-edit.obo imports chebi. 

In `go-ontology/src/ontology/` I run `robot report --input go-edit.obo --output report.tsv --profile profile.txt`, and the profile.txt looks like:
```
ERROR	annotation_whitespace
ERROR	deprecated_class_reference
ERROR	duplicate_definition
WARN	duplicate_exact_synonym
ERROR	duplicate_label_synonym
ERROR	duplicate_label
WARN	duplicate_scoped_synonym
ERROR	equivalent_pair
WARN	invalid_xref
ERROR	label_formatting
ERROR	label_whitespace
INFO	lowercase_definition
ERROR	missing_definition
ERROR	missing_label
WARN	missing_obsolete_label
ERROR	missing_ontology_description
ERROR	missing_ontology_license
ERROR	missing_ontology_title
INFO	missing_superclass
ERROR	misused_obsolete_label
INFO	multiple_asserted_superclasses
ERROR	multiple_definitions
ERROR	multiple_labels
ERROR	file:./../sparql/never-in-taxon-non-IRI-value-violation.sparql
```
The report.tsv indicates about 2000 `missing_label` errors on CHEBI terms implying they are in this ontology, but these CHEBI terms are not in the go-edit.obo file, they're imported. 

So am I doing something wrong in the way I invoke the report command, or on the file that I'm reporting on? Or is this a bug? I see in the code it's supposed to not pull in all the imports I think? Thanks for your help!

cmungall:
Good point, this is underspecified at the moment.

There was an analogous issue with the `query` command: #158
see the fix: #328 and the instructions http://robot.obolibrary.org/query

Should we have equivalent options for report?

However, the defaults may have to be different, because as you point out any ontology that has imports and references classes in the import will produce false positive errors. @jamesaoverton @rctauber how do you feel about having the defaults be different here?

rctauber:
I would think that you’re only reporting on your base ontology. That way,
issues with other ontologies won’t break the report.

That said, it might make sense to have an option for including imports
anyway, which would make it easier to report on the edit versions of
ontologies who use templates like OBI.

cmungall:
[...]

Issue #383, opened 2018-10-22
Deprecated entities in ROBOT reports


rctauber:
After getting some feedback about the `report` command, I have some proposed changes for how deprecated entities are handled. I think this might be part of a wider discussion of the OBO Foundry standards for deprecated entities, though. 

[Branch located here](https://github.com/rctauber/robot/tree/report-query-fix). Below are the current changes in the branch.

**EXCLUDE deprecated entities from checks for:**
* [duplicate definition](http://robot.obolibrary.org/report_queries/duplicate_definition)
* [duplicate exact synonym](http://robot.obolibrary.org/report_queries/duplicate_exact_synonym)

(Deprecated entities are *partially* excluded from the checks above already. A deprecated entity will not show up as a violation, but if a non-deprecated entity shares a synonym/definition with a deprecated entity, it *will* show up. The change totally excludes deprecated entities and allows a non-deprecated entity to share a synonym/definition with a deprecated one.)

**INCLUDE deprecated entities in checks for:**
* [multiple labels](http://robot.obolibrary.org/report_queries/multiple_labels)
* [multiple definitions](http://robot.obolibrary.org/report_queries/multiple_definitions)
* [missing label](http://robot.obolibrary.org/report_queries/missing_label)

I'd like to get some feedback before creating a PR:
* Is it OK for a deprecated entity to share a definition with another non-deprecated entity? What about an exact synonym?
* Should we worry about deprecated entities with multiple definitions?
* Are there any other checks that should exclude deprecated entities? Such as:
    * [duplicate label-synonym](http://robot.obolibrary.org/report_queries/duplicate_label_synonym)
    * [duplicate scoped synonym](http://robot.obolibrary.org/report_queries/duplicate_scoped_synonym)

jamesaoverton:
I don't see why a deprecated term should have the same textual definition as another term. If the textual definition was good, why was the term deprecated?

I'm a little uncomfortable about the exact synonyms, but I could be convinced that ignoring this check for deprecated terms is ok.

PR #394, opened 2018-11-05
Report Query Fixes

rctauber:
Updates:
* Exclude entities with ONLY a type declaration from label & def checks (#391)
* Exclude subset definitions (children of `oboInOwl:SubsetProperty`) from label & def checks (#391)
* Allow deprecated classes to use OP 'has obsolescence reason' (#391)
* Allow dashes in CURIE prefixes (#395)
* Fix processing of comment lines in queries (#400)
* Weaken rules for deprecated entities (#383)
* Check for multiple equivalent classes (#402)


rctauber:
@cmungall @dougli1sqrd - I've added this in to missing definitions.

dougli1sqrd:
Okay, I think I need some SPARQL advice now :). 

So far, I have:
```
PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
PREFIX obo: <http://purl.obolibrary.org/obo/>

  VALUES ?property { obo:IAO_0000115 }
  ?entity ?any ?o .
  FILTER NOT EXISTS { ?entity ?property ?value }
  FILTER NOT EXISTS { ?entity a owl:Ontology }
  FILTER NOT EXISTS { ?entity a owl:NamedIndividual }
  FILTER NOT EXISTS { ?entity owl:deprecated true }
  FILTER NOT EXISTS {
    ?entity rdfs:subPropertyOf oboInOwl:SubsetProperty .
  }
  FILTER EXISTS {
      ?entity ?any2 ?object
      FILTER(?any2 != rdf:type)
  }
  FILTER (!isBlank(?entity))
}
ORDER BY ?entity
```
Which basically mirrors the missing label query. But here we pick up on (ironically) IAO:0000115 "definition", in addition to other annotation properties, like oboInOwl:hasAlternativeId, etc. These work for label since they have one. These properties don't have a definition though. 

```
<http://www.geneontology.org/formats/oboInOwl#hasExactSynonym> rdf:type owl:AnnotationProperty ;
                                                               rdfs:label "has_exact_synonym"^^xsd:string .
```
These all only have a type and a label. Any ideas? Thanks.

rctauber:
This portion is including the label triple in the filter:
```
  FILTER EXISTS {
      ?entity ?any2 ?object
      FILTER(?any2 != rdf:type)
  }
```

This was intended to exclude entities that ONLY have a type declaration. I already pushed this change to this branch. You could change it to this to exclude entities that have only a type and label:
```
  FILTER EXISTS {
      ?entity ?any2 ?object
      FILTER(?any2 != rdf:type)
      FILTER(?any2 != rdfs:label)
  }
```

At this point, you might be better off just filtering out annotation properties. But, why shouldn't an annotation property have a definition? I guess this is another OBO-to-OWL issue, as the default `oboInOwl` annotation properties will be added with a label but not a definition.

PR #396, opened 2018-11-06
Restoring previous version

cmungall:
Accidentally committed experimental changes to master previously

### DIFF
--- a/robot-core/src/main/resources/report_queries/deprecated_class_reference.rq
+++ b/robot-core/src/main/resources/report_queries/deprecated_class_reference.rq
@@ -14,16 +14,17 @@ SELECT DISTINCT ?entity ?property ?value WHERE {
      owl:equivalentClass
      rdfs:subClassOf
    }
-   ?entity owl:deprecated true .
-   FILTER ?dep .
+   ?entity owl:deprecated ?dep .
+   FILTER regex(str(?dep), "true") .
    ?value ?property ?entity .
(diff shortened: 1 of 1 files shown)


rctauber:
The `FILTER (?property != owl:deprecated))` should probably stay, as that's how the query should be... I'm not totally sure why I did a string match on the IRI there.

I was doing a string match on `owl:deprecated true` to make sure to catch any instances where people might accidentally use a string instead of a boolean.

dougli1sqrd:
This PR in particular was to restore what was there originally since Chris' changes got messed up. But I think we're probably not in much danger from accidentally using string true, since won't this run on obo files most of the time?

cmungall:
@dougli1sqrd I don't understand the relevance of obo files here.

All: apologies for the confusion. I accidentally committed my original changes to master. This PR is to restore things to how they were so we can then properly evaluate my changes as a PR with travis checks etc. But probably best to abandon this PR and work from master...

dougli1sqrd:
@cmungall I just meant that in robot obo will convert to correct OWL which will use actual true/false values, rather than string "true".

jamesaoverton:
@cmungall No worries. We can lock `master` on this repo if you prefer.

Sorry, but I don't understand what to do with this PR at this point. Please either merge or close, as you see fit.

cmungall:
Let's just work off of what we have in master

PR #403, opened 2018-11-15
Template Rework

rctauber:
This PR completely reworks the logic of `template`. Instead of using a static `TemplateOperation` class, it now uses a `Template` class. This allows for a cleaner way to add more features to template.

This rework also supports new template strings. I've included the documentation and examples below. I apologize in advance about the length 😅 TLDR:
- Property template strings
    - property logic (super properties, equivalents, disjoints, inverses)
    - extra property types (functional, inverse functional, symmetric, etc.)
- Individual template strings
    - property assertions (object and data)
    - same and different individual assertions
- Aliases for entity TYPE declarations
    - e.g. `object property` can be used instead of `owl:ObjectProperty`

@ontodev/robot-team This is an experimental update, so I would appreciate tests from anyone. The old template strings should produce the same results as before, even though the logic is different.

### Property Template Strings

- `PROPERTY_TYPE`: ROBOT creates a property for each row of data that has a `TYPE` of either an object or data property. The property type can be (any type followed by a \* can ONLY be used for object properties):
    - **logical types**: these types link the created property to other properties (annotation properties can *only* use `subproperty`)
      - `subproperty`: the created property will be a subproperty of each templated property expression (default)
      - `equivalent`: the created property will be equivalent to all of the templated property expressions
      - `disjoint`: the created property will be disjoint from each templated property expression and the values cannot be the same
      - `inverse`\*: the created object property will be the inverse of each templated property expression
    - **property types**: these types define the type of the created property, and will not work with annotation properties
      - `functional`: the created property will be functional, meaning each entity (subject) can have at most one value
      - `inverse functional`\*: the created object property will be inverse functional, meaning each value can have at most one subject
      - `irreflexive`\*: the created object property will be irreflexive, meaning the subject cannot also be the value
      - `reflexive`\*: the created object property will be reflexive, meaning each subject is also a value
      - `symmetric`\*: the created object property will be symmetric, meaning the subject and value can be reversed
      - `asymmetric`\*: the created object property will be asymmetric, meaning the subject and value cannot be reversed
      - `transitive`\*: the created object property will be transitive, meaning the property can be chained
- `P` **property expression**: If the template string starts with a `P` and a space then it will be interpreted as a property expression. The value of the current cell will be substituted into the template, replacing all occurrences of the `%` character. Then the result will be parsed into an OWL property expression. ROBOT uses the same syntax for property expressions as Protégé: [Manchester Syntax](http://www.w3.org/2007/OWL/wiki/ManchesterSyntax). If it does not recognize a name, ROBOT will assume that you're trying to refer to an entity by its IRI or CURIE. This can lead to unexpected behavior, but it allows you to refer to entities without loading them into the input ontology.
    - **object properties**: the only supported object property expression is the inverse object property expression. The template string is `P inverse(%)`. A single object property for a value can be specified by `P %`.
    - **data properties**: data property expressions are not yet supported by OWL. A data property for a value (e.g. for a parent property) can be specified by `P %`.
    - **annotation properties**: annotation property expressions are not possible. An annotation property for a value (e.g. for a parent property) can be specified by `P %`.
- `DOMAIN`: The domain to a property is a class expression in [Manchester Syntax](http://www.w3.org/2007/OWL/wiki/ManchesterSyntax) (for object and data properties). For annotation properties, the domain must be a single class specified by label, CURIE, or IRI.
- `RANGE`: The range to a property is either a class expression in [Manchester Syntax](http://www.w3.org/2007/OWL/wiki/ManchesterSyntax) (for object properties) or the name, CURIE, or IRI of a datatype (for annotation and data properties).

#### Example of Property Template Strings 

| **TYPE** | **PROPERTY_TYPE** | **P %** | **DOMAIN** | **RANGE** |
| --- | --- | --- | --- | --- |
| owl:ObjectProperty | subproperty | Property 1 | Class 1 | Class 2 |
| owl:DataProperty | functional | Property 2 | Class 2 | xsd:string |

The `functional` data property will still default to a `subproperty` logical axiom for the `P %` template string, unless a different logical property type (`equivalent`, `disjoint`) is provided. Property type can be split, e.g. `PROPERTY_TYPE SPLIT=|`.

### Individual Template Strings

- `INDIVIDUAL_TYPE`: ROBOT creates an individual for each or of data that has a `TYPE` of another class. The individual type can be:
    - `named`: the created individual will be a default named individual. When the `INDIVIDUAL_TYPE` is left blank, this is the default. This should be used when adding object property or data property assertions
    - `same`: the created individual will be asserted to be the same individual as each templated individual in the row
    - `different`: the created individual will be asserted to be a different individual than any of the templated individuals in the row
- `I` **individual assertion**:
    - `I property %`: when creating a `named` individual, replace property with an object property or data property to add assertions. The `%` will be replaced by the template cell value or values. For object property assertions, this is another individual. For data property assertions, this is a literal value.
    - `I %`: when creating a `same` or `different` individual, this template string is used to specify which individual will be the value of the same or different individual axiom.

#### Example of Individual Template Strings

| **INDIVIDUAL_TYPE** | **I 'Property 1' some %** | **I %** |
| --- | --- | --- |
| named | Individual 2 | |
| different | | Individual 1 | 

### Datatype Template Strings

*Datatypes can currently be added (`TYPE` = `owl:Datatype`), but I have not yet added support for [datatype definitions](https://www.w3.org/TR/owl2-syntax/#Datatype_Definitions). I'm still figuring out the best way to do this that wouldn't make the template strings too complex. There's nothing in the docs for this yet. Suggestions would be appreciated.*

### TYPE declarations

- `TYPE`: this is the `rdf:type` for the row. Because ROBOT is focused on ontology development, the default value is `owl:Class` and this column is optional. When creating an OWLIndividual, specify the class to which it belongs in this column.
    - `class` or `owl:Class`
    - `object property` or `owl:ObjectProperty`
    - `data property` or `owl:DataProperty`
    - `annotation property` or `owl:AnnotationProperty`
    - `datatype` or `owl:Datatype`


robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
```diff
@@ -27,6 +27,7 @@
 import org.obolibrary.obo2owl.OWLAPIOwl2Obo;
 import org.obolibrary.oboformat.model.OBODoc;
 import org.obolibrary.oboformat.writer.OBOFormatWriter;
+import org.obolibrary.robot.template.TemplateHelper;
```
jamesaoverton: Looks like this isn't used.

robot-core/src/main/java/org/obolibrary/robot/QuotedEntityChecker.java
```diff
@@ -332,28 +311,31 @@ public String getLabel(IRI iri) {
...
   }
 
   /**
-   * Find an annotation property with the given name. Quotation marks will be removed if necessary.
+   * Find an annotation property with the given name, or create one. Quotation marks will be removed
+   * if necessary.
    *
    * @param name the name of the entity to find
-   * @param create when true and an IOHelper is defined, create the property
+   * @param create if false, do not create a new annotation property
```
jamesaoverton: I don't understand this change. Is it just the wording, or has the code changed?

robot-core/src/main/java/org/obolibrary/robot/TemplateOperation.java
```diff
@@ -144,9 +146,9 @@ public static OWLDatatype getDatatype(QuotedEntityChecker checker, String name)
    * @return a new annotation with property and string literal value
    * @throws Exception if the annotation property cannot be found
    */
-  public static OWLAnnotation getStringAnnotation(
-      QuotedEntityChecker checker, String template, String value) throws Exception {
-    return TemplateHelper.getStringAnnotation(checker, template, value);
+  public static void getStringAnnotation(QuotedEntityChecker checker, String template, String value)
```
jamesaoverton: It doesn't make sense to me that these signatures are changed to `void`, even if they're deprecated.

robot-core/src/main/java/org/obolibrary/robot/TemplateOperation.java
```diff
@@ -365,11 +341,11 @@ public static OWLOntology template(
         if (template.isEmpty()) {
           continue;
         }
-        if (!validateTemplateString(template)) {
-          throw new Exception(
-              String.format(
-                  unknownTemplateError, tableName, column + 1, headers.get(column), template));
-        }
+        // if (!validateTemplateString(template)) {
```
jamesaoverton: If this code isn't used, don't just comment it out, delete it. The same goes for the commented-out code below.

robot-core/src/main/java/org/obolibrary/robot/template/Template.java
```diff
@@ -0,0 +1,1852 @@
...
+
+  /** Namespace for error messages. */
+  private static final String NS = "template#";
+
+  /**
+   * Error message when the number of header columns does not match the number of template columns.
+   * Expects: table name, header count, template count.
+   */
+  private static final String columnMismatchError =
+      NS
```
jamesaoverton: Formatting problem?

robot-core/src/main/java/org/obolibrary/robot/template/Template.java
```diff
@@ -0,0 +1,1852 @@
...
+   * Given a template name, a list of rows, and an IOHelper, create a template object with a new
+   * QuotedEntityChecker. The rows are added to the object, new labels from the rows are added to
+   * the checker, and a Manchester Syntax parser is created.
+   *
+   * @param name template name
+   * @param rows list of rows (lists)
+   * @param ioHelper IOHelper to resolve prefixes
+   * @throws Exception on issue adding table to template object
+   */
+  public Template(String name, List<List<String>> rows, IOHelper ioHelper) throws Exception {
```
jamesaoverton: This constructor is almost identical to the previous one. I wish the code could be more DRY, but maybe that's just Java.

robot-core/src/main/java/org/obolibrary/robot/template/Template.java
```diff
@@ -0,0 +1,1852 @@
...
+            String.format(unknownTemplateError, name, column + 1, headers.get(column), template));
+      }
+
+      // Get the location of important columns
+      switch (template) {
+        case "ID":
+          idColumn = column;
+          break;
+        case "LABEL":
+        case "A rdfs:label":
```
jamesaoverton: This is cheating. What if the user chooses a different prefix or label for `rdfs:label`? It should still work.

robot-core/src/main/java/org/obolibrary/robot/template/Template.java
```diff
@@ -0,0 +1,1852 @@
...
+
+  /**
+   * Process each of the table rows. First, get an entity based on ID or label. If the template
+   * contains an ID column, but it is empty, skip that row. If it does not contain an ID column,
+   * skip if the label is empty. Add axioms based on the entity type (class, object property, data
+   * property, annotation property, datatype, or individual).
+   *
+   * @throws Exception on issue creating axioms from template
+   */
+  private void processRows() throws Exception {
```
jamesaoverton: As a matter of style, I would prefer putting this logic in a singular `processRow(...)` method.

robot-core/src/main/java/org/obolibrary/robot/template/Template.java
```diff
@@ -0,0 +1,1852 @@
...
+    String nextTemplate;
+    try {
+      nextTemplate = templates.get(column + 1);
+    } catch (IndexOutOfBoundsException e) {
+      nextTemplate = null;
+    }
+
+    // Support for annotations on: superclasses, domains, and ranges
+    Set<OWLAnnotation> axiomAnnotations = new HashSet<>();
+    if (nextTemplate != null && !nextTemplate.trim().isEmpty() && (nextTemplate.startsWith(">C"))) {
```
jamesaoverton: ">C" seems wrong to me. Shouldn't it be ">A"? I see the same thing below. We definitely need a test that covers these annotations.

robot-core/src/main/java/org/obolibrary/robot/template/Template.java
```diff
@@ -0,0 +1,1852 @@
...
+            "functional",
+            "inversefunctional",
+            "inverse functional",
+            "irreflexive",
+            "reflexive",
+            "symmetric",
+            "transitive"));
+    if (propertyTypes.size() > 1) {
+      // TODO - can only have one of: subproperty, equivalent, disjoint, or inverse
+      throw new Exception();
```
jamesaoverton: Please throw a nicer exception, with a description and link to the docs.

docs/template.md
```diff
@@ -20,24 +20,96 @@ The `template` command accepts an optional input ontology, either using the `--i
...
 - **axiom annotations**: ROBOT can also annotate logical and annotation axioms.
     - `>A` annotation on annotation: Annotates the annotation axiom created from the cell to the left with the cell value. The column to the left must be an `A*` template string.
     - `>C` annotation on class expression: Annotates the class expression axiom created from the cell to the left with the cell value. The column to the left must be a `C` template string.
```
jamesaoverton: This seems wrong to me. Why do we care what the type of the expression to the left is? We care about the type of the current column, which should be one of the various 'A*' types.

robot-command/src/main/java/org/obolibrary/robot/TemplateCommand.java
```diff
@@ -137,7 +135,16 @@ public CommandState execute(CommandState state, String[] args) throws Exception
       tables.put(templatePath, TemplateHelper.readTable(templatePath));
     }
 
-    OWLOntology outputOntology = TemplateOperation.template(tables, inputOntology, null, ioHelper);
```
jamesaoverton: Switching to a `Template` class makes sense and is fine, but I want to maintain our overall design pattern: `TemplateCommand` should call a static method on `TemplateOperation`.

So I would prefer to define a `TemplateOperation.template2(...)` method, even if all it does is run this little bit of code that instantiates a new `Template`.

robot-core/src/main/java/org/obolibrary/robot/template/TemplateHelper.java
```diff
@@ -153,147 +170,168 @@ public static OWLAnnotation getAnnotation(
...
+        expressions.addAll(parser.parseDataPropertyList());
+      }
+    }
+
+    return expressions;
   }
 
   /**
-   * Find an annotation property with the given name or create one.
+   * TODO
```
jamesaoverton: TODO what?

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
rctauber: We have mirrored read methods in `IOHelper` and `TemplateHelper`. (`readCSV`, `readTSV`, `readTable`...). The `IOHelper` methods are never used but removing them would cause a backwards compatibility issue. The `IOHelper` methods just use:
```
return TemplateHelper.readTable(path);
```

robot-core/src/main/java/org/obolibrary/robot/QuotedEntityChecker.java
rctauber: Just the wording has changed. I can change it back if the first way makes more sense.

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
jamesaoverton: Ok

robot-core/src/main/java/org/obolibrary/robot/TemplateOperation.java
rctauber: I think that was an oversight on my part - changing these back.

robot-core/src/main/java/org/obolibrary/robot/template/Template.java
rctauber: I think the next string was just too long and the auto-formatter did this weird thing. I'll divide it up better.

rctauber:
`>C` and `>A` is how we've been doing it, but I can see how that doesn't make sense. Should I switch everything to `>A`?

robot-core/src/main/java/org/obolibrary/robot/template/Template.java
rctauber: I left a comment in the PR in regards to this.

robot-core/src/main/java/org/obolibrary/robot/template/TemplateHelper.java
rctauber: I was trying to add support for data ranges and it was pretty messy. The current template implementation doesn't support data ranges, but it could be added in the future. I'm going to remove this method for now since it doesn't actually do anything.

jamesaoverton: Ok

jamesaoverton:
I'd prefer to switch everything to `>A*`, since it's more flexible and makes more sense. I didn't catch this in previous versions -- my fault.

For backwards compatibility, I guess we could allow `>C` as an undocumented option, treating it as `>A`. Maybe we could do a string substitution early in the processing and log a warning.

robot-core/src/main/java/org/obolibrary/robot/template/Template.java
rctauber: The `LABEL` column is used to be able to specify terms by label. If a user is using a different AP, it will still be added to the class. But how would we know if a given AP is what the user considers a "label"?

jamesaoverton: `LABEL` is a special keyword. That’s fine. I’m saying that if we accept `A label` then I don’t think that it shouldn’t be handled as a special case: we should check each `A foo` to see if the name resolves to `rdfs:label`, based on previously declared prefixes and labels.

rctauber: Changed to only `case "LABEL"`

robot-core/src/main/java/org/obolibrary/robot/template/Template.java
rctauber: Changed to all `>A*`. `>C*` can still be used but it will log a warning to use `>A*` instead.

rctauber:
As rows are parsed, exception messages will be logged as `ERROR`s, e.g.:
```
2019-03-15 09:59:00,005 ERROR org.obolibrary.robot.Template - MANCHESTER PARSE ERROR the expression 'is_specified_output_of some 'protein-protein interaction detection'' at row 171, column 5 cannot be parsed: encountered unknown 'protein-protein interaction detection'
2019-03-15 09:59:00,006 ERROR org.obolibrary.robot.Template - MANCHESTER PARSE ERROR the expression 'is_specified_output_of some 'extracellular electrophysiology recording'' at row 173, column 5 cannot be parsed: encountered unknown 'extracellular electrophysiology recording'
```
If there were any exception messages after parsing all rows, the operation will fail.

You can override this with `--force true` (default is `false`, so it will fail if there are any errors in the template). It will still attempt to create an ontology, even if there were parse errors. Those axioms will probably just not be included in the output.

The operation will fail immediately if there are any errors with the headers.

jamesaoverton:
For many users, I think that it's easier to fail on the first error, then fix and try again. That was the old behaviour. It's better to start fixing errors at the top than the bottom, but by printing all errors to the terminal you'll see the last error first.

I can see how advanced users might want all errors at once, but that requires a more complex mental model of the operation.

rctauber:
I can see how that can be confusing. I find it frustrating when I have a template and I need to keep running it after fixing one thing at a time. 

Maybe the default behavior, instead of showing all errors, is to fail on the first error. Then we can add another option that lets you print all errors? I'm not sure what to call it though, some ideas:
* `--log-all-errors true` (default false)
* `--fail-first-error false`(default true)
* `--fail-fast false` (default true)

Or the behavior could just be if `--force false` then fail on the first error. If `--force true`, log all errors and attempt to create an ontology.

jamesaoverton:
I like your final suggestion.

rctauber:
@ontodev/robot-admin

While working on GAZ, I found that this new template rework handles very large template files MUCH better than the old template. It's faster, and doesn't fail if the file is too large.

I ran template on a 20.5MB template CSV with this new code, and it processed in 28 seconds. If I run the same command using `master`, it failed after 52 minutes with `java.lang.OutOfMemoryError: Java heap space`.

I think it would be good to get this merged into `master` soon, assuming the code has no breaking changes. The command line interface should run exactly the same, and as far as I can tell, it is entirely backwards compatible with existing templates.

I had to make a change to the labels in order to get it backwards-compatible with the OBI templates...

It makes sense that you should be able to define a new class and use it by-label in the template. But, we changed the special label column to only work with `LABEL`, since people may use custom label properties. Therefore, if you're not using `LABEL`, you can't reference new classes. I think a good workaround is to assume that `rdfs:label` is the label property, but allow users to specify a different label property in the command:
```
robot template --label-property ns:my_label --template template.csv --output template.owl
```

If you're using `A rdfs:label` as a template string, you don't need to do anything and you can reference new labels in the template. This `--label-properety` option will also set a different property in the `QuotedEntityChecker` so that all labels are picked up using the custom property.

docs/examples/template.csv
```diff
@@ -1,5 +1,5 @@
-ID,Label,Definition,Definition Source,See Also,Editor,RDF Type,Class Type,Parent IRI
-ID,A rdfs:label,A IAO:0000115,A IAO:0000119,AI rdfs:seeAlso,A IAO:0000117,TYPE,CLASS_TYPE,CI
-ex:F344N,F 344/N,An inbred strain of rat used in many scientific investigations.,James A. Overton,http://www.informatics.jax.org/external/festing/rat/docs/F344.shtml,James A. Overton,owl:Class,subclass,NCBITaxon:10116
-ex:B6C3F1,B6C3F1,An inbred strain of mouse used in many scientific investigations.,James A. Overton,http://jaxmice.jax.org/strain/100010.html,James A. Overton,owl:Class,subclass,NCBITaxon:10090
-ex:rat-1234,rat 1234,,,F 344/N,,ex:F344N,,
```
jamesaoverton: This should not need to be changed. If it does, then something has broken backwards compatibility.

docs/template.md
```diff
@@ -16,28 +16,124 @@ Each template file must be set up in the following format:
...
-- `LABEL`: If a term exists in an ontology, or its ID has been defined elsewhere (perhaps in a previous template), then the `LABEL` column can specify an `rdfs:label` that uniquely identifies the target term. This can be easier the numeric IDs for human readers.
+### Generic Template Strings
+
+- `ID`: Every term must have an IRI to identify it. This can be specified with an `ID` column. Usually this will be a prefixed ID like `GO:12345`. See the `--prefix` options for details.
+    - If an entity already exists in the `--input` ontology, you can refer to it by `LABEL` instead (see below).
+    - Rows with no `ID` or `LABEL` will be skipped.
+- `LABEL`: If a term exists in an ontology, or its ID has been defined elsewhere (perhaps in a previous template), then the `LABEL` column can specify an `rdfs:label` that uniquely identifies the target term. This can be easier than the numeric IDs for human readers. 
+    - The `LABEL` column DOES NOT create an `rdfs:label` annotation for an entity.
```
jamesaoverton: Does this differ from the old behaviour?

docs/template.md
```diff
@@ -16,28 +16,124 @@ Each template file must be set up in the following format:
...
+    - `equivalent`: the created class will be asserted to be equivalent to the intersection of all the templated class expressions
+- `C` **class expression**: If the template string starts with a `C` and a space then it will be interpreted as a class expression. The value of the current cell will be substituted into the template, replacing all occurrences of the percent `%` character. Then the result will be parsed into an OWL class expression. 
+    - ROBOT uses the same syntax for class expressions as Protégé: [Manchester Syntax](http://www.w3.org/2007/OWL/wiki/ManchesterSyntax). This means that an entity can be referred to by its `rdfs:label` (enclosing in single quotes if it has a space in it). 
+    - If it does not recognize a label, ROBOT will assume that you're trying to refer to a class by its IRI (or compact IRI). This can lead to unexpected behaviour, but it allows you to refer to classes (by IRI) without loading them into the input ontology. This is particularly useful when the input ontology would be too large, such as the NCBI Taxonomy.
+    - Properties in class expression **must** be referred to by label in order to be parsed.
```
jamesaoverton: "class expressions"

docs/template.md
```diff
@@ -16,28 +16,124 @@ Each template file must be set up in the following format:
...
+
+#### Example of Class Template Strings
+
+| Label | Entity Type | Class Type | Related classes | Other axioms |
+| --- | --- | --- | --- | --- |
```
jamesaoverton: It's a bit of a pain, but please use spaces to align these columns so the table reads better in raw Markdown.

docs/template.md
```diff
@@ -16,28 +16,124 @@ Each template file must be set up in the following format:
...
+| --- | --- |
+| LABEL | C % |
+| Class 4 | |
+| Class 5 | (part_of some 'Class 4') |
+
+In this template, Class 5 would be a subclass of `part_of some 'Class 4'`.
+
+### Property Template Strings
+
+- `PROPERTY_TYPE`: ROBOT creates a property for each row of data that has a `TYPE` of either an object or data property. The property type can be (any type followed by a \* can ONLY be used for object properties):
```
jamesaoverton: Aren't annotation properties included? They're mentioned in the next line.

docs/template.md
```diff
@@ -16,28 +16,124 @@ Each template file must be set up in the following format:
...
+    - **logical types**: these types link the created property to other properties (annotation properties can *only* use `subproperty`). You may only include ONE of these in each `PROPERTY_TYPE`.
```
jamesaoverton: Does this column type allow splits?

docs/template.md
```diff
@@ -16,28 +16,124 @@ Each template file must be set up in the following format:
...
+      - `subproperty`: the created property will be a subproperty of each templated property expression (default)
+      - `equivalent`: the created property will be equivalent to all of the templated property expressions
+      - `disjoint`: the created property will be disjoint from each templated property expression and the values cannot be the same
+      - `inverse`\*: the created object property will be the inverse of each templated property expression
+    - **property types**: these types define the type of the created property, and will not work with annotation properties. You may include any number of these.
```
jamesaoverton: These property types have very specific meanings for logicians and mathematicians. I'm worried that these simplifications are not precise enough.

docs/template.md
```diff
@@ -16,28 +16,124 @@ Each template file must be set up in the following format:
...
+| OP:1 | owl:ObjectProperty | subproperty | Property 1 | Class 1 | Class 2 |
+| DP:1 | owl:DataProperty | functional | Property 2 | Class 2 | xsd:string |
+
+The `functional` data property will still default to a `subproperty` logical axiom for the `P %` template string, unless a different logical property type (`equivalent`, `disjoint`) is provided. Property type can be split, e.g. `PROPERTY_TYPE SPLIT=|`.
+
+### Individual Template Strings
+
+If the `TYPE` is a defined class, `owl:Individual`, or `owl:NamedIndividual`, an instance will be created. If the `TYPE` does not include a defined class, that instance will have no class assertions. You may include a `SPLIT=` in `TYPE` if you wish to provide more than one class assertion for an individual.
+
+- `INDIVIDUAL_TYPE`: ROBOT creates an individual for each or of data that has a `TYPE` of another class. The individual type can be:
```
jamesaoverton: I don't understand "each or of data"?

jamesaoverton:
I started reviewing the documentation page and didn't finish, but I'll submit the review anyway and try again tomorrow.

docs/template.md
rctauber: No - it should only be one of the logical types (e.g., not both equivalent and subproperty). I added a note to make this clear.

docs/examples/template.csv
rctauber: Fixed to old version

docs/template.md
rctauber: I thought the old behavior was the same as this, but I was wrong. I've updated the `LABEL` column to create an `rdfs:label` annotation.

docs/examples/template.csv
jamesaoverton: This is not the original version. The [original version](https://github.com/ontodev/robot/blob/master/docs/examples/template.csv) has a `CI` pattern that the new code does not accept:

```
UNKNOWN TEMPLATE ERROR could not interpret template string "CI" for column 9 ("Parent IRI") in table "template.csv".
```

So the new code is breaking backwards compatibility.

docs/examples/uberon_template.csv
```diff
@@ -1,4 +1,4 @@
-ID,Label,Definition,Definition Source,See Also,Editor,RDF Type,Class Type,Parent IRI
-ID,A rdfs:label,A IAO:0000115,A IAO:0000119,AI rdfs:seeAlso,A IAO:0000117,TYPE,CLASS_TYPE,CI
-UBERON:0000995,brain,The brain is the center of the nervous system in all vertebrate...,UBERON,http://en.wikipedia.org/wiki/Brain,Becky Tauber,owl:Class,subclass,UBERON:0000062
-UBERON:0000211,ligament,Dense regular connective tissue...,UBERON,http://purl.obolibrary.org/obo/go/references/0000034,Becky Tauber,owl:Class,subclass,UBERON:0000062
\ No newline at end of file
+ID,Label,Definition,Definition Source,See Also,Editor,RDF Type,Parent IRI
+ID,A rdfs:label,A IAO:0000115,A IAO:0000119,AI rdfs:seeAlso,A IAO:0000117,TYPE,SC %
```
jamesaoverton: Like the other integration test, if these have to be changed for the new code to run then we have broken backward compatibility. This file has to stay exactly the same as on `master`: https://github.com/ontodev/robot/blob/master/docs/examples/uberon_template.csv

robot-core/src/main/java/org/obolibrary/robot/TemplateHelper.java
```diff
@@ -622,6 +944,178 @@ public static OWLEntity getEntity(
...
+    } else if (template.matches("^(>?C|>{0,2}A[LTI]?|>?P|>?D|>?I|>?R|>?DT) .*")) {
+      return true;
+    } else if (template.startsWith("PROPERTY_TYPE")) {
+      return true;
+    } else if (template.startsWith("CHARACTERISTIC")) {
+      return true;
+    } else if (template.startsWith("INDIVIDUAL_TYPE")) {
+      return true;
+    } else {
+      return template.matches("^(SC|EC|DC|SP|EP|DP|IP|NI|SI|DI|CI|PI|RI|II).*");
```
jamesaoverton: Listing the options explicitly is more clear. But some of these must be followed by a space, and a few are not.

Either way, this overlaps with the previous regex. We should just have one complete regex.

robot-core/src/test/resources/template.csv
```diff
@@ -1,5 +1,5 @@
-Label,Label Comment,ID,Synonyms,Type,Parent,Parts,Parts Annotation
-A rdfs:label,>AL rdfs:comment@en,ID,A IAO:0000118 SPLIT=|,CLASS_TYPE,C %,C part_of some %,>C rdfs:comment
```
jamesaoverton: I would prefer for this unit test to stay the same as it was.

robot-core/src/main/java/org/obolibrary/robot/Template.java
```diff
@@ -0,0 +1,2047 @@
...
+      // Add a type if the type is not owl:Individual or owl:NamedIndividual
+      if (!type.equalsIgnoreCase("individual")
+          && !type.equalsIgnoreCase("named individual")
+          && !type.equalsIgnoreCase("owl:NamedIndividual")
+          && !type.equalsIgnoreCase("owl:Individual")) {
+        OWLClass typeCls = checker.getOWLClass(type);
+        if (typeCls != null) {
+          axioms.add(dataFactory.getOWLClassAssertionAxiom(typeCls, individual));
+        } else {
+          /
```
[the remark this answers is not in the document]
rctauber: I don't think I understand. Where do you want to not allow wrapping in single quotes? The parser uses the `QuotedEntityChecker` as well.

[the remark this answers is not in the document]
jamesaoverton: Hmm. On the "Coversation" tab this is not showing the right line of code, but on the "Files changed" tab I see it in the right place.

I'm worried about `String propStr = template.substring(3).replace("'", "");` which won't work if the label contains an apostrophe.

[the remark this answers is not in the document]
jamesaoverton: This is for 1706 `String propStr = template.substring(2).replace("'", "");`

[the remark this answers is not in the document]
rctauber: Thanks! That makes more sense.

rctauber:
> The NI option is not documented, and I'm not sure that I understand it.

This `NI` option isn't really necessary anymore. It's exactly the same as `I` so I'm going to remove it. Let me know if you think otherwise.

jamesaoverton:
It's fine to remove `NI`.\

robot-core/src/main/java/org/obolibrary/robot/TemplateHelper.java
rctauber: Changed to regex for each type of template string (annotations, classes, etc...)

robot-core/src/test/resources/template.csv
rctauber: Reverted

rctauber:
OK - I made the requested changes and added a new integration test with the new template strings.

robot-core/src/main/java/org/obolibrary/robot/Template.java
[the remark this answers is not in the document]
rctauber: I've changed the documentation to explicitly state not to wrap these properties in single quotes.

docs/examples/template.csv
```diff
@@ -1,5 +1,6 @@
...
-ex:F344N,F 344/N,An inbred strain of rat used in many scientific investigations.,James A. Overton,http://www.informatics.jax.org/external/festing/rat/docs/F344.shtml,James A. Overton,owl:Class,subclass,NCBITaxon:10116
-ex:B6C3F1,B6C3F1,An inbred strain of mouse used in many scientific investigations.,James A. Overton,http://jaxmice.jax.org/strain/100010.html,James A. Overton,owl:Class,subclass,NCBITaxon:10090
-ex:rat-1234,rat 1234,,,F 344/N,,ex:F344N,,
\ No newline at end of file
+ID,Label,Definition,Definition Source,See Also,Editor,Type,Parent Class,Disjoint Class,Domain,Range,Individual Weight
+ID,A rdfs:label,A IAO:0000115,>A IAO:0000119,AI rdfs:seeAlso,A IAO:0000117,TYPE,SC %,DC %,DOMAIN,RANGE,I weight in kilograms
+ex:F344N,F 344/N,An inbred strain of rat used in many scientific investigations.,James A. Overton,http://www.informatics.jax.org/external/festing/rat/docs/F344.shtml,James A. Overton,class,NCBITaxon:10116,,,,
+ex:B6C3F1,B6C3F1,An inbred strain of mouse used in many scientific investigations.,James A. Overton,http://jaxmice.jax.org/strain/100010.html,James A. Overton,class,NCBITaxon:10090,F 344/N,F 344/N' or B6C3F1,,
```
jamesaoverton: The third row is a class with domain "'F 344/N' or B6C3F1". That can't be right.

docs/examples/template-old.csv
```diff
@@ -0,0 +1,5 @@
+ID,Label,Definition,Definition Source,See Also,Editor,RDF Type,Class Type,Parent IRI
```
jamesaoverton: Renaming this file is fine, but we need the test to run somewhere. Now I think it is not.

jamesaoverton:
🎉 🎂 🎈 !!!

PR #406, opened 2018-12-05
Allow multiple labels with different lang tags

rctauber:
As discussed on the OFOC call yesterday, we should allow labels in different languages. This changes the report "multiple label" query to allow more than one label as long as they have different language tags. If there is no language tag, it assumes the language is English. If one label has no tag and the other has an `@en` tag, it will be reported as a violation.

See also https://github.com/information-artifact-ontology/ontology-metadata/issues/27

### DIFF
--- a/robot-core/src/main/resources/report_queries/multiple_labels.rq
+++ b/robot-core/src/main/resources/report_queries/multiple_labels.rq
@@ -1,6 +1,6 @@
 # # Multiple Labels
 #
-# Problem: An entity has more than one label. This may cause confusion or misuse, and will prevent translation to OBO format. Excludes deprecated entities.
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Thanks for giving us an implementation to discuss.

Internationalization is really important and really hard. I don't like this appraoch at all.

1. It's simply wrong to assume that all `xsd:string` literals (default for RDF 1.1) are `@en` literals. Xrefs aren't English, for example.
2. Even if we make this English assumption, downstream tools won't. If some OBO ontologies correctly tag their literals and some rely on this assumption, then SPARQL queries will get mixed results. Non-OBO datasets that we want to integrate with won't make this assumption.
3. If we only make the English assumption for `rdfs:label` (then 'definition', then 'example of usage', then 'editor note', and so on) then we're just breeding more inconsistencies in our queries: inconsistent between different predicates, this time.

The right solution is for ontology authors to tag their literals properly and consistently at the source. We should build tools to help them do that.

cmungall:
Agreed, and this should be part of a larger obo-wide plan that provides recommendations for tools like OLS

Issue #293, opened 2018-05-08
Add rename command


cmungall:
Not 100% sure this belongs in ROBOT but this functionality would be useful. It would essentially be a wrapper to OWLEntityRenamer

First use case is rename given a set of mappings IRI1->IRI2, specified as a TSV/CSV

The second use case is to rename based on prefixes, given partial mappings IRIBase1->IRIBase2

jamesaoverton:
Yeah, I think this would be good for ROBOT.

rctauber:
Is this similar to #61?

jamesaoverton:
Looks like they're kind of dual operations.

This is the kind of thing that I would do with string-substitution, and end up shooting myself in the foot somehow.

@rctauber Please draft an implementation that covers #293 and #61. No rush.

rctauber:
Would it make more sense to have two different input options for the two different use cases (full vs. partial)? Like this:
```
-m, --mappings <arg>        table of entities to rename
-r, --partial <arg>         table of entities to rename prefixes
```

Or just one input and then a boolean for partial vs. full, like this:
```
-m, --mappings <arg>        table of entities to rename
-r, --partial <arg>         if true, only rename prefixes (default false)
```

jamesaoverton:
Can you give an example of `--partial` please? I'm not sure that I understand.

rctauber:
I was thinking `--partial` would rename the base IRIs, as @cmungall requested in the original issue.

e.g.:
```
Old Base,New Base
http://purl.obolibrary.org/obo/,http://foo.bar/
```

So all entities with the prefix `http://purl.obolibrary.org/obo/<ID>` would be changed to `http://foo.bar/<ID>`.

Maybe "partial" isn't the right word... any other ideas?

jamesaoverton:
Ok. Maybe `--prefix-mappings`? Other suggestions welcome.

jamesaoverton:
Added in #419

PR #419, opened 2018-12-28
Add rename command

rctauber:
See #293

```
usage: robot rename --input <file> --mapping <file> --output <file>
 -A,--add-prefix <arg>      add a new prefix to ontology file header
    --catalog <arg>         use catalog from provided file
 -f,--full <arg>            table of mappings for renaming
 -h,--help                  print usage information
 -i,--input <arg>           load ontology from a file
 -I,--input-iri <arg>       load ontology from an IRI
 -noprefixes                do not use default prefixes
 -o,--output <arg>          save ontology to a file
 -p,--prefix <arg>          add a prefix 'foo: http://bar'
 -P,--prefixes <arg>        use prefixes from JSON-LD file
 -r,--partial <arg>         table of partial mappings for renaming
 -V,--version               print version information
 -v,--verbose               increased logging
 -vv,--very-verbose         high logging
 -vvv,--very-very-verbose   maximum logging, including stack traces
 -x,--xml-entities          use entity substitution with ontology XML
                            output
```

The `--add-prefix` option makes sure that the prefix appears in the final output file, which does not happen if you just use `--prefix`.

Right now, the `--full` and `--partial` files expect a header - should this be assumed? Or should we do no headers?

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/RenameCommand.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/RenameCommand.java
@@ -59,7 +59,7 @@ public class RenameCommand implements Command {
     o.addOption("m", "mappings", true, "table of mappings for renaming");
     o.addOption("r", "prefix-mappings", true, "table of prefix mappings for renaming");
     o.addOption("A", "add-prefix", true, "add a new prefix to ontology file header");
-    o.addOption("f", "allow-duplicates", true, "allow two or more terms to be renamed to the same full IRI");
+    o.addOption("d", "allow-duplicates", true, "allow two or more terms to be renamed to the same full IRI");
     options = o;
   }
 
--- a/robot-command/src/main/java/org/obolibrary/robot/RenameCommand.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/RenameCommand.java
@@ -1,0 +1,245 @@
+package org.obolibrary.robot;
+
+import com.opencsv.CSVParserBuilder;
+import com.opencsv.CSVReader;
+import com.opencsv.CSVReaderBuilder;
+import java.io.*;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Scanner;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.Options;
+import org.semanticweb.owlapi.model.OWLOntology;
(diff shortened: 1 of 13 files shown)


jamesaoverton:
Hey everyone. We think this one is ready to go. Please try it out and give us some feedback.

robot-command/src/main/java/org/obolibrary/robot/RenameCommand.java
```diff
@@ -0,0 +1,243 @@
...
+        if (mappings.containsValue(nextLine[1]) && warnDuplicateValues) {
+          Scanner s = new Scanner(System.in);
+          System.out.println(
+              String.format(
+                  "---------- WARNING ----------\n"
+                      + "Line %d in file '%s' contains a duplicate value ('%s').\n"
+                      + "This will rename two separate entities to have the same IRI, resulting in a merge."
+                      + "\nDo you wish to continue? [y/N]",
+                  lineNum, mappingsFile.getPath(), nextLine[1]));
+          String cont = s.nextLine();
```
jamesaoverton: I given this some thought, and I've decided that promtping for user input is a bad idea. ROBOT is meant to be used for automation. None of our other commands are interactive. `make` will get stuck forever if it hits this warning, which will surprise and annoy our users.

I think the better alternatives are:

1. just log a warning
2. add a `--fail-on-duplicates true/false` option

robot-command/src/main/java/org/obolibrary/robot/RenameCommand.java
```diff
@@ -0,0 +1,245 @@
...
+  /** Initialze the command. */
+  public RenameCommand() {
+    Options o = CommandLineHelper.getCommonOptions();
+    o.addOption("i", "input", true, "load ontology from a file");
+    o.addOption("I", "input-iri", true, "load ontology from an IRI");
+    o.addOption("o", "output", true, "save ontology to a file");
+    o.addOption("m", "mappings", true, "table of mappings for renaming");
+    o.addOption("r", "prefix-mappings", true, "table of prefix mappings for renaming");
+    o.addOption("A", "add-prefix", true, "add a new prefix to ontology file header");
+    o.addOption("f", "allow-duplicates", true, "allow two or more terms to be renamed to the same full IRI");
```
jamesaoverton: `--allow-duplicates` is a better name, but now “f” is not a good short name. Maybe “a” or “d”, or no short name.

PR #436, opened 2019-02-07
Add an --intermediates option to extract

rctauber:
## Intermediates

When extracting (especially with MIREOT), sometimes the hierarchy can have too many intermediate classes, making it difficult to identify relevant relationships. For example, you may end up with this after extracting `adrenal gland`:
```
 - material anatomical entity
   - anatomical structure
     - multicellular anatomical structure
       - organ
         - abdomen element
           - adrenal/interrenal gland
             - adrenal gland (*)
     - lateral structure
       - adrenal gland (*)
```

By specifying how to handle these intermediates, you can reduce unnecessary intermediate classes:

* `--intermediates all`: default behavior, do not prune the ontology
* `--intermediates minimal`: only include intermediate intermediates with more than one *sibling* (i.e. the parent class has another child)
* `--intermediates none`: do not include any intermediates
  * For MIREOT, this will only include top and bottom level classes.
  * For any SLME method, this will only include the classes directly used in the logic of the input terms

The above example, with `--intermediates minimal`, would become:
```
- material anatomical entity
  - anatomical structure
    - adrenal gland (*)
    - organ
      - adrenal gland (*)
```

You can chain this output into `reduce` to further clean up the structure, as some redundant axioms may appear.

With `--intermediates none`:
```
- material anatomical entity
  - adrenal gland
```

Any term specified as an input term will not be pruned.


rctauber:
If you have a class that is asserted to be a subclass of itself, the `remove` (and `filter`) operation goes into an infinite loop and fails during spanning gaps (minimal intermediates).

*Ideally* an ontology shouldn't have circular subclasses, but in case it does happen, do we want this operation to fail with an error? Or should we just log a `WARN` and remove that superclass from the next iteration of `spanGapsHelper`? @jamesaoverton

cmungall:
If redundant axioms are removed with reduce then there will be no cycles in the subclass graph

rctauber:
True. I can just make it throw a message that says something like:
```
CIRCULAR SUBCLASS ERROR your ontology contains a circular subclass definition, 
  see x for details
```

balhoff:
I don't think this is an error. It might be unusual, but I think the algorithm should just handle it. Maybe warn about it or provide an option to turn it into an error, but it is a valid situation.

rctauber:
Closed in favor of #441

PR #439, opened 2019-02-21
Add axiom generators to reason

rctauber:
See https://github.com/ontodev/robot/issues/273#issuecomment-458566408

Specifically:
* Adds `--axiom-generators "<gens>"` (see comment for usage)
* Adds `--include-indirect <bool>` option for `SubClass`, `ClassAssertion`, and `SubObjectProperty`

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/ReasonerHelper.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/ReasonerHelper.java
@@ -1,11 +1,7 @@
 package org.obolibrary.robot;
 
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
 import java.util.stream.Collectors;
 import org.obolibrary.robot.exceptions.IncoherentRBoxException;
 import org.obolibrary.robot.exceptions.IncoherentTBoxException;
@@ -13,6 +9,7 @@ import org.obolibrary.robot.exceptions.InconsistentOntologyException;
 import org.semanticweb.owlapi.model.*;
 import org.semanticweb.owlapi.model.parameters.Imports;
 import org.semanticweb.owlapi.reasoner.OWLReasoner;
+import org.semanticweb.owlapi.util.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.ac.manchester.cs.owlapi.modularity.ModuleType;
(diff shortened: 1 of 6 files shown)


robot-core/src/main/java/org/obolibrary/robot/ReasonerHelper.java
```diff
@@ -225,6 +229,62 @@ private static OWLOntology createIncoherentModule(
     return module;
   }
 
+  /**
```
jamesaoverton: Please add a JavaDoc string

robot-core/src/main/java/org/obolibrary/robot/ReasonerHelper.java
```diff
@@ -225,6 +229,62 @@ private static OWLOntology createIncoherentModule(
...
+      gens.add(new InferredSubClassAxiomGenerator());
+      return gens;
+    }
+    for (String ax : axGenerators) {
+      gens.add(getInferredAxiomGenerator(ax));
+    }
+    return gens;
+  }
+
+  /**
```
jamesaoverton: Please add a JavaDoc string

rctauber:
@jamesaoverton I just added the unit tests.

robot-core/src/main/java/org/obolibrary/robot/ReasonerHelper.java
```diff
@@ -225,6 +229,66 @@ private static OWLOntology createIncoherentModule(
...
+      case "subclass":
+      case "":
+        return new InferredSubClassAxiomGenerator();
+      case "disjointclasses":
+        return new InferredDisjointClassesAxiomGenerator();
+      case "equivalentclass":
+        return new InferredEquivalentClassAxiomGenerator();
+      case "datapropertycharacteristic":
+        return new InferredDataPropertyCharacteristicAxiomGenerator();
+      case "equivalentdataproperty":
```
jamesaoverton: For consistency, I think this should be "properties" plural to match the class name.

rctauber: I was debating that one, since (oddly) the corresponding object property class is `InferredEquivalentObjectPropertyAxiomGenerator` while the data property one uses "properties". I guess as long as the correct one shows up in the docs, it will be clear.

jamesaoverton: Ok. I'd rather stick to the OWLAPI names. Either they have good reasons, or we can blame inconsistencies on them. :)

rctauber: Fixed!

Issue #446, opened 2019-03-04
Filter or remove entities by pattern-matching IRIs


beckyjackson:
During the ODK call, there was a discussion of removing entities that are not "part of a namespace of interest". I think this would be a beneficial addition to the `remove` and `filter` commands.

You would input an IRI pattern, e.g. `http://purl.obolibrary.org/obo/OBI_*` and filter (or remove) all terms that use this namespace.

This could be implemented as a `--select` argument. It would be based on simple pattern matching. You could include more than one namespace of interest.

For `--select`:
```
robot filter --input input.owl \
      --select "IRI=http://purl.obolibrary.org/obo/OBI_*" \
      --output remove.owl
```

We could do something like:
* `IRI=<IRI-pattern>`
* `IRI=~<IRI-pattern>`
* `IRI~<IRI-pattern>`
* or just: `--select <IRI-pattern>`

Currently this can be implemented by running a query to find all terms that use that namespace, then using that set of terms as the `--term-file`. This would just take out one step and make it easier for users who are not familiar with SPARQL.

CC @matentzn

matentzn:
Actually we have an action item for a test in QC that flags up (warns) about axioms that do not seem to belong to the ontology ('accidentally slipped in during protege editing').  Maybe we can use this to design such a test as well.

jamesaoverton:
I agree that it's a good feature. I guess the angle-bracket syntax is the most straightforward.

(Looking at the `remove` docs just now, the `CURIE=CURIE` stuff is not very clear without immediate example.)

@matentzn How would you detect an axiom that does not seem to belong?

matentzn:
We currently play with a feature in the ODk that allows us to define a set of iri patterns, usually one, that constitutes iris that are somewhat ‘owned’ by the ontology. An axiom should be flagged as potentially foreign if it does both contain at least one entity with an owned iri in its signature!

rctauber:
@jamesaoverton the angle-bracket syntax as in `--select <http://purl.obolibrary.org/obo/OBI_*>`?

jamesaoverton:
Yeah

beckyjackson:
This feature was added by #448 and is active in 1.4.1.

PR #448, opened 2019-03-07
Add IRI pattern matching to remove/filter selectors

rctauber:
See #446, also adds examples to the CURIE matching.

@matentzn I would appreciate if you could test this out for your use case, if you get a chance! Thanks!

----
### Patterns

...

It is also possible to select terms based on parts of their IRI. You may include an IRI pattern using one or more wildcard characters (`*` and/or `?`) enclosed in angle brackets.

- `<IRI-pattern>`
    - e.g. `<http://purl.obolibrary.org/obo/EX_*>`
    
If you wish to match a more complicated pattern, you may also use a regex pattern here by preceding the pattern with a tilde (`~`):
- `<~IRI-regex>`
    - e.g. `<~^.+EX_[0-9]{7}$>`

...

## Examples

...

Remove all axioms containing an entity in the BFO namespace from the UBERON module:

    robot remove --input uberon_module.owl \
      --select "<http://purl.obolibrary.org/obo/BFO_*>" \
      --output results/remove_bfo.owl

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -42,6 +42,10 @@ Many commands involve creating IRIs from provided string representations of the 
 
 When using CURIEs, make sure the prefix is defined, or add it with `--prefix`.
 
+### Invalid IRI Pattern Error
+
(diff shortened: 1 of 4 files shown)


matentzn:
@rctauber if this technically works, can we defer testing for until it is in a release? Pretty full plate at the moment, and I am VERY keen on this feature :P

Issue #451, opened 2019-03-13
Feature request: make --copy-ontology-annotations global for all of ROBOT


matentzn:
I currently have the problem that I wish to filter axioms from an ontology when generating a release artefact, for example filtering all axioms that contain an entity in a signature term file or filtering all logical axioms. Unfortunately, the filter command strips out ontology annotations as well, which often contain credits and metadata that I wish to preserve in the new release artefact. @rctauber pointed out to be that extract has the --copy-ontology-annotations option, which is exactly what I need. Would it be possible to make this global to all ROBOT commands (in particular filter)? Or should I go about this in another way altogether?

jamesaoverton:
The use case is clear, and I agree that ROBOT should address it.

I don't thing a global option is the right thing. We have a lot of commands, and `--copy-ontology-annotations` would be irrelevant or confusing for many of them, e.g. `diff`, `convert`, `merge`. For `filter` and `remove` it could work the same as `extract`, and I see the benefit. For `template` or some variations on `query` it could sometimes make sense, but other times be confusing.

A more general solution would be adding a `--copy foo.owl` option to the `annotate` command that would copy all ontology annotations from the target ontology to the input ontology.

Or we could have some command that just extracts the ontology metadata to a file, and then you could `merge` that file with whatever you want. A related idea is for `remove` and `filter` to be able to target the ontology metadata. I think that this is not currently supported.

dosumis:
> A related idea is for `remove` and `filter` to be able to target the ontology metadata. I think that this is not currently supported.  

This would be useful.

rctauber:
> A related idea is for remove and filter to be able to target the ontology metadata. I think that this is not currently supported.

I second this method. I think it fits better with the current remove and filter methods.

@matentzn would `--select ontology-annotations` cover your use case?

matentzn:
Thanks both!

@rctauber : If 
```
robot filter --term-file $(ONTOLOGYTERMS) --trim false --select ontology-annotations \
```
Gives me all axioms with signature in ONTOLOGYTERMS plus the ontology annotations, then yes!

jamesaoverton:
I'd prefer `--select ontology` (maybe with `--axioms annotation`) if we can make that work. Either way, it shouldn't interfer with whatever the main job of the filter is.

rctauber:
[...]

PR #452, opened 2019-03-13
Add 'ontology' selection for remove and filter

rctauber:
To remove or filter for the ontology annotations:
```
robot filter --input x.owl --select ontology --output y.owl
```
Returns a new ontology with *only* the ontology annotations from `x.owl`

```
robot remove --input x.owl --select ontology --output y.owl
```
Returns a copy of `x.owl` *without* any ontology annotations.

See #451

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/FilterCommand.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/FilterCommand.java
@@ -106,14 +106,14 @@ public class FilterCommand implements Command {
 
     IOHelper ioHelper = CommandLineHelper.getIOHelper(line);
     state = CommandLineHelper.updateInputOntology(ioHelper, state, line);
-    OWLOntology ontology = state.getOntology();
+    OWLOntology inputOntology = state.getOntology();
     OWLOntologyManager manager = OWLManager.createOWLOntologyManager();
 
     // Get a set of entities to start with
     Set<OWLObject> objects = new HashSet<>();
     if (line.hasOption("term") || line.hasOption("term-file")) {
       Set<IRI> entityIRIs = CommandLineHelper.getTerms(ioHelper, line, "term", "term-file");
-      objects.addAll(OntologyHelper.getEntities(ontology, entityIRIs));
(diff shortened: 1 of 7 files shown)


matentzn:
Just to be annoying (sorry):

```
robot filter --input x.owl --select ontology --term-file terms.txt --trim false --output y.owl
```

Returns a new ontology y with the ontology annotations from x.owl and all axioms containing terms in terms.txt?

robot-command/src/main/java/org/obolibrary/robot/FilterCommand.java
```diff
@@ -128,49 +128,81 @@ public CommandState execute(CommandState state, String[] args) throws Exception
...
         selectGroup.remove("annotations");
       }
-      selectGroups.add(selectGroup);
-    }
-
-    // If there are no objects, add all the objects from the ontology
-    if (objects.isEmpty()) {
-      objects.addAll(OntologyHelper.getObjects(ontology));
+      // The ontology keyword retrieves the ontology annotations
+      if (selectGroup.contains("ontology")) {
```
jamesaoverton: What about "imports"?

rctauber: The filter command doesn't work on imports (for remove, it just removes the import statements). I guess we could add it in filter to just filter for the import statements...

jamesaoverton:
I think this is right. It has always bugged me that "imports" is handled in a special way, and now "ontology" is handled in a special way, but I don't see an alternative.

Please include a test that addresses @matentzn's specific use case.

rctauber:
Side note, @matentzn - I would also add the `annotations` selector to that command to make sure any classes that get pulled in as a result of the filter have labels, etc (otherwise they will only have the declarations).

robot-command/src/main/java/org/obolibrary/robot/FilterCommand.java
rctauber: I added `--select imports` for filter - it will copy the import declarations from the input to the output ontology.

matentzn:
Thanks @rctauber!

PR #458, opened 2019-04-04
Remove major performance problem in materialize command.

balhoff:
This makes `materialize` at least 50 times faster (not sure, I didn't wait long enough for the slow version to complete).

@cmungall how long has it been taking you to build Uberon?


cmungall:
a long time!

Issue #432, opened 2019-02-02
Enhance existing diff command, separate logic from display


cmungall:
Core requirements:

 - __separate diff logic from diff rendering__
 - define a JSON representation of a diff that can be used in a variety of renderers, stored in a database, etc
 - include a default markdown renderer in ROBOT, with options to show at different levels of granularity
    - complete diff on a per axiom basis, grouped by objects (see diffs on http://uberon.org)
    - a high level summary, e.g. new classes and obsoleted classes (see examples on https://github.com/EnvironmentOntology/envo/releases)

Additional requirements that could be added later:

 - reasoner diff
 - ability to collapse low-level axiom-oriented diffs into user-centered model (e.g. move class, obsolete class)
 - reverse operation, aka patch - e.g. take a diff file and apply it
 - services (@balhoff has this: https://stars-app.renci.org/owldiff/)
     - would be good if this could be invoked a la ontobot via a message in the ticket
 - github agents for showing markdown of PRs and for making PRs for patches

These more advanced requirements are described in a separate doc [INCA Change Language](https://docs.google.com/document/d/1__7p64FOI5ZhiZ6F2TXtUc8JN1XXGwglOiVRrlg9G_c/edit)

The core idea is to separate logic from display. Implement logic once, make sure it is scalable.

All kinds of nice displays are possible once we do this, implementable in the most appropriate language; e.g. the old go-diffs that used coloration to nice effect to see textual diffs on a per string literal level.

#33 #360

cmungall:
Looping @lpalbou into conversation

lpalbou:
@balhoff 

Go-stats original ticket: https://github.com/geneontology/go-site/issues/842#issuecomment-514530467

The ontology diff (json): https://geneontology.s3.amazonaws.com/go-ontology-changes.json
Tsv version: https://geneontology.s3.amazonaws.com/go-ontology-changes.tsv

These files should soon be on the current.geneontology.org too.

cmungall:
I believe this ticket can be closed? Some of the stretch reqts could go to a new ticket.

I would still like a summary report such as the one Laurent showed, but maybe this could be done trivially by operating on a json output?

jamesaoverton:
Ok. The primary issue has been addressed, so I'll close this.

PR #461, opened 2019-04-10
Use owl-diff library for diff operation. Add markdown and html diff output formats.

balhoff:
This addresses some of the issues in #432, especially "separate diff logic from diff rendering". It also includes a Markdown diff rendering, although currently different levels of granularity are not configurable.

Diff in this PR is implemented by calling out to functionality in my owl-diff library, which is also used in the owl-diff server. This diff is faster than the old version (I think the main slowdown in the old one was rendering of all axioms to string before performing comparisons). Also, label rendering in the simple output is done with a custom ShortFormProvider rather than regex operations.

This version passes the existing tests, although I made one small change to one test data file to correct what I believe to be an inconsistency in CURIE rendering in the previous code.


cmungall:
Perfect timing as we work on making the "end parts" of the release pipeline (diff for release notes and publishing large files to github) easier here: https://github.com/INCATools/ontology-development-kit/issues/205

balhoff:
@cmungall please take a look at the Markdown output and let me know if I can make any improvements.

docs/diff.md
```diff
@@ -22,7 +22,8 @@ If `--output` is provided then a report will be written with any differences bet
 
 See <a href="/examples/release-diff.txt" target="_blank">`release-diff.txt`</a> for an example. In the output, 'Ontology 1' corresponds to your `--left` input and 'Ontology 2' corresponds to your `--right` input.
 
-The output is in OWL Manchester syntax, but you can include entity labels with `--labels true`.
```
jamesaoverton: @balhoff Please confirm:

The old documentation was wrong, because the output was always Functional, not Manchester. So the behaviour is not changing, and this PR is backwards compatible.

robot-core/pom.xml
```diff
@@ -148,5 +148,10 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.geneontology</groupId>
+      <artifactId>owl-diff_2.12</artifactId>
```
jamesaoverton: This is a weird `artifactId` but I guess it's correct.

docs/diff.md
balhoff: Yes, that's right, the output was always Functional.

robot-core/pom.xml
balhoff: Scala artifacts get the Scala major version appended, so that you can publish separate releases for different Scala versions, which aren't generally binary-compatible.

jamesaoverton:
Thanks @balhoff.

Are we using other Scala stuff in ROBOT already? Does the dependency on `scala-lang` make a big difference to the size of `robot.jar`?

balhoff:
@jamesaoverton I think this is the first Scala dependency. The change to `robot.jar` from this PR is about 5 MB, and I would guess most of that is the Scala lib, since owl-diff is tiny and doesn't add any other new dependencies.

balhoff:
@jamesaoverton I think this is the first Scala dependency. The change to `robot.jar` from this PR is about 5 MB, and I would guess most of that is the Scala lib, since owl-diff is tiny and doesn't add any other new dependencies.

robot-command/src/main/java/org/obolibrary/robot/DiffCommand.java
```diff
@@ -47,7 +47,12 @@ public DiffCommand() {
     o.addOption("r", "right", true, "load right ontology from file");
     o.addOption("R", "right-iri", true, "load right ontology from IRI");
     o.addOption("o", "output", true, "save results to file");
-    o.addOption(null, "labels", true, "if true, use labels in place of entity IRIs");
+    o.addOption(
+        null,
+        "labels",
+        true,
+        "if true, append labels after entity IRIs in the simple format output");
+    o.addOption("f", "format", true, "format for diff output: simple (default) | html | markdown");
```
rctauber: The other formats are actual formats (HTML or MD) whereas simple is just plain-text. Maybe instead of `simple`, it could be `txt` or `text` for plain-text?

robot-core/src/main/java/org/obolibrary/robot/DiffOperation.java
```diff
@@ -151,6 +174,7 @@ public static boolean compare(
    * @param axiom a string representation of an OWLAxiom
    * @return a string with labels inserted next to IRIs
    */
+  @Deprecated
```
rctauber: The `@Deprecated` methods should also have a `@deprecated` Javadoc tag

robot-core/src/main/java/org/obolibrary/robot/DiffOperation.java
```diff
@@ -112,35 +112,58 @@ public static boolean compare(
...
+                  ontologyProvider,
+                  pm,
+                  pm,
+                  Collections.singletonList(OWLManager.getOWLDataFactory().getRDFSLabel()),
+                  Collections.emptyMap());
+          OBOShortenerShortFormProvider iriProvider = new OBOShortenerShortFormProvider(pm);
+          DoubleShortFormProvider doubleProvider =
+              new DoubleShortFormProvider(iriProvider, labelProvider);
+          writer.write(BasicDiffRenderer.render(diff, doubleProvider));
+        } else {
```
rctauber: With simple format, no labels, I get outputs like:
`- SubClassOf(<http://purl.obolibrary.org/obo/ECO_0000000> owl:Thing)`

When using labels, I see:
`- SubClassOf(<ECO:0000000>[evidence] <owl:Thing>)`

Could you use `OBOShortenerShortFormProvider iriProvider` for the else case of `useLabels` as well?

robot-core/src/test/resources/simple.diff
```diff
@@ -1,4 +1,4 @@
 0 axioms in Ontology 1 but not in Ontology 2:
 
 1 axioms in Ontology 2 but not in Ontology 1:
-+ SubClassOf(<https://github.com/ontodev/robot/robot-core/src/test/resources/simple.owl#test1>[Test 1] owl:Thing)
++ SubClassOf(<https://github.com/ontodev/robot/robot-core/src/test/resources/simple.owl#test1>[Test 1] <owl:Thing>)
```
rctauber: It's a very minor thing, but when using labels, `owl:Thing` is enclosed in the angle brackets. When not using labels, it isn't enclosed. I wish there were a way to make the simple output a little easier to read but... this might be as good as we're going to get.

cmungall:
As well as getting complete diffs, it's useful to get simple lists e.g of new terms and of obsoleted terms. These are particularly useful for including in release notes, e.g https://github.com/EnvironmentOntology/envo/releases

Should there be separate report types?

Having a JSON format could also be useful, particularly for summary stats and lists

Would the logic for this reside in owl-diff, robot, or external? The way James Malone had bubastis work had a nice separation of logic, with the html generated from xslt. Here it could be something like mustache.

But I think this can be done incrementally, I am good with merging this

cc @lpalbou

jamesaoverton:
@cmungall I'd prefer to merge this PR before adding more features, unless you want to head in a very different direction.

@rctauber added some comments and I would like @balhoff to address them.

robot-command/src/main/java/org/obolibrary/robot/DiffCommand.java
balhoff: I agree, I will update.

robot-core/src/main/java/org/obolibrary/robot/DiffOperation.java
balhoff: Okay.

robot-core/src/main/java/org/obolibrary/robot/DiffOperation.java
balhoff: I think so—you're wanting it to look like this, correct?

`- SubClassOf(ECO:0000000 owl:Thing)`

robot-core/src/test/resources/simple.diff
balhoff: I think it is more consistent this way. For the label output, we have an identifier enclosed in angle brackets, possibly followed by a label in square brackets. To me it is confusing if sometimes the identifier is not in angle brackets.

jamesaoverton:
@balhoff I would like to merge this, but there are a few small things that still need to be resolved.

balhoff:
Thanks for the reminder, @jamesaoverton, I have been meaning to finish this up. Will do so soon.

robot-command/src/main/java/org/obolibrary/robot/DiffCommand.java
balhoff: Addressed in 3048bb8935c9a65d56480abccf7ab3d7d648100c.

robot-core/src/main/java/org/obolibrary/robot/DiffOperation.java
balhoff: Addressed in 26ede14fbd1e9f655672325f61541a668a6049d4.

robot-core/src/main/java/org/obolibrary/robot/DiffOperation.java
balhoff: Addressed by c0597da90e1cb5360eeb439324f0e6dcc2bb768f.

balhoff:
@jamesaoverton @rctauber this should be ready now.

jamesaoverton:
The `docs/examples/release-diff.txt` file has been changed to use CURIEs instead of full IRIs. This breaks the old behaviour, which we try hard not to do. The CURIEs are nicer but somebody downstream might be depending on the full IRIs.

I have two suggestions and a problem. To keep the old behaviour we could either/both:
1. introduce a `--curies` option that defaults to false
2. add a `plain` default/legacy format, and a `pretty` text format with labels and CURIEs

When I tried to do implement (1) by using just labels and not CURIEs, it broke the `testCompareModifiedWithLabels` unit test. I think it's because the `labelProvider` returns an empty string when there's no label. I didn't see a way to just use the IRI in this case.

cmungall:
I like the idea of having an open-ended list of formats. This could be orthogonal to whether CURIEs, URIs are used; and orthogonal to whether the markdown uses `[..](URL)` or plain embedding.

I don't think we want to get boxed in by early formatting choices. I think we will learn more as people use this feature and we will want to incorporate new formats or options.

Also if we have a standard JSON people can easily customize the display layer as they like.

One thing we really need is a "summary" diff that makes something suitable for release notes. E.g. list of new terms and list of obsoleted terms. The number of hacks I see for doing this around different OBO GH repos makes me sad...

balhoff:
@jamesaoverton I was trying to address this comment from @rctauber:

> With simple format, no labels, I get outputs like:
> `- SubClassOf(<http://purl.obolibrary.org/obo/ECO_0000000> owl:Thing)`
>
> When using labels, I see:
> `- SubClassOf(<ECO:0000000>[evidence] <owl:Thing>)`
>
> Could you use OBOShortenerShortFormProvider iriProvider for the else case of useLabels as well?

We could go back to the previous output (which I think did automatically shorten some IRIs, such as owl:Thing).

For your label issue, we ought to be able to create a label provider with an alternative backup using [this constructor](http://owlcs.github.io/owlapi/apidocs_4/org/semanticweb/owlapi/util/AnnotationValueShortFormProvider.html#AnnotationValueShortFormProvider-java.util.List-java.util.Map-org.semanticweb.owlapi.model.OWLOntologySetProvider-org.semanticweb.owlapi.util.ShortFormProvider-).

jamesaoverton:
Ok, I added a `plain` format using the `renderPlain` function, and a `pretty` format that uses labels and CURIEs. It passes all the old tests. Is this ok with you @balhoff ?

balhoff:
This sounds pretty good. I think now `--labels` is kind of a holdover, since it's just a way to trigger `pretty` when used in combination with `plain`. But that is kind of unavoidable when introducing a large set of formats.

jamesaoverton:
Yes, `--labels` becomes a legacy way to upgrade `plain` to `pretty`. It's lame but it should prevent anybody's code from breaking.

jamesaoverton:
Thanks for the PR. It's nice to have Markdown and HTML formats supported, at long last! I'm happy to include more formats.

Issue #466, opened 2019-04-29
Documentation of -- trim is back to front


dosumis:
"By default, filter will include all axioms from the input ontology that contain one or more entities from the specified set. To be more strict and only include axioms in which all entities in the axiom are in the specified set, use --trim false."

--> use --trim true

rctauber:
Thanks for catching this. It looks like the CLI option description could be a little better as well. I'll make these fixes.

jamesaoverton:
#467 should address this.

beckyjackson:
Resolved by #467

PR #467, opened 2019-04-29
Add better --trim docs

rctauber:
See #466


jamesaoverton:
I'm regretting the use of the word "trim" now, but hopefully these changes to the docs make the behaviour easier to understand.

PR #476, opened 2019-05-16
Is lower-case defintion-test correct?

drseb:
Hi,

for hpo I get the report:

`INFO	lowercase_definition	HP:0030271	IAO:0000115	2,3-diphosphoglycerate (2,3-DPG) controls the movement of ...
`

Is this expected behaviour? This PR could fix this maybe, if this not intended behaviour.

### DIFF
--- a/robot-core/src/main/resources/report_queries/lowercase_definition.rq
+++ b/robot-core/src/main/resources/report_queries/lowercase_definition.rq
@@ -10,6 +10,6 @@ SELECT DISTINCT ?entity ?property ?value WHERE {
   VALUES ?property { obo:IAO_0000115
                      obo:IAO_0000600 }
   ?entity ?property ?value .
-  FILTER (!regex(?value, "^[A-Z]"))
+  FILTER (!regex(?value, "^[A-Z0-9]"))
 }
 ORDER BY ?entity


drseb:
In case this seem like good move, you would also have to change this file: https://github.com/ontodev/robot/blob/ac231ab68dbb6b01ec9560f10567dc18eac87ffc/docs/report_queries/lowercase_definition.md

jamesaoverton:
I would be OK with changing this query, and that's the right place to make the change. The Markdown file is generated from the SPARQL file.

But I'm not sure about making this change, because the example from HP:0030271 is not actually a definition. The check is supposed to help (in a small way) to ensure good definitions. I'm having trouble thinking of an example of a proper definition that would start with a number. Do you have a better example?

matentzn:
I think it happens often that this pattern is used for writing:

LABELOFCLASS is a DEFINITION.

So **2,3-diphosphoglycerate (2,3-DPG) is a XYZ** is a distinct possibility? :P

matentzn:
OR
```
2,3-diphosphoglycerate (2,3-DPG) levels are considered abnormal when...
```

drseb:
@jamesaoverton : Why is it not a definition. I agree it is not an optimal definition, but I think all the information is in there. Just in case you haven't seen it, the whole definition is: 

> 2,3-diphosphoglycerate (2,3-DPG) controls the movement of oxygen from red blood cells to tissues. Anemia is usually accompanied by an increased level of 2,3-DPG in order to promote tissue oxygenation. This term refers to an inappropriate low 2,3-DPG concentration in erythrocytes.

drseb:
But again, if you think this test should also check that the definition starts with a letter, than I am happy to consider changing the definition. Thanks

matentzn:
Just imagine defining the chemical compound! Its highly possible that the definitions starts with the label, which means it can start with the number.. The question is not whether definitions are starting with numbers or not; they can. The question is whether QC would miss typical cases of errors if we merged this PR.. Is that the case @jamesaoverton ?

jamesaoverton:
OBO Principle 6 on textual definitions recommends Aristotelian form: <http://www.obofoundry.org/principles/fp-006-textual-definitions.html>, which would be "A LABELOFCLASS is a LABELOFPARENT that CONDITIONS". So this form starts with "A" or "An" and not a label.

The example for HP:0030271 includes enough information for a definition, but that information is not provided in the form of a definition.

The point of this INFO-level check (not ERROR or WARN) is to encourage the recommended definition format. After thinking about this some more: If we allow numbers, why not allow lowercase letters, since our labels are not usually capitalized? Then this check is pointless.

We ran these checks by the OBO Operations Committee multiple times, and nobody objected to this one. If you want to make a case that we should not have this check at all, please open a new tracker item and we can discuss it with the OFOC.

Your questions are good ones, and much appreciated. We'll revise the documentation for this check to point to Principle 6.

matentzn:
Ah ok, now that convinced me.. I was not aware that there were such detailed recommendations made by us! Fair enough! :P

PR #482, opened 2019-05-21
Add --print option to report

rctauber:
See discussion in #429. From docs:

If you do not provide an `--output`, all violations will be printed to the terminal in either the specified format or TSV (if no format is specified). You can also choose to print the details of the first `n` number of violations using `--print <n>`:

```
robot report --input edit.owl \
 --print 5 \
 --output report.tsv
```

This will print the first five violations and also write the report to `report.tsv`. If you do not provide an `--output` while using `--print <n>`, *only* the first `n` violations will be printed to the terminal.

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/ReportCommand.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/ReportCommand.java
@@ -1,5 +1,6 @@
 package org.obolibrary.robot;
 
+import java.util.Map;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Options;
 import org.semanticweb.owlapi.model.OWLOntology;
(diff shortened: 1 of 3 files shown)


robot-command/src/main/java/org/obolibrary/robot/ReportCommand.java
```diff
@@ -99,26 +101,21 @@ public CommandState execute(CommandState state, String[] args) throws Exception
...
     // If any violations are found of the fail-on level, this will be false
-    // If failOn is "none" or if no violations are found, this will be true
-    boolean success =
-        ReportOperation.report(
-            ontology, ioHelper, profilePath, outputPath, format, failOn, useLabels);
+    // If fail-on is "none" or if no violations are found, this will be true
+    boolean success = ReportOperation.report(ontology, ioHelper, outputPath, reportOptions);
     if (!success) {
       logger.error("Report failed!");
       System.exit(1);
```
jamesaoverton: Is there a reason that we exit like this rather than using the new error handling system?

rctauber: I did it this way because I didn't want the exception formatting from the error handling - I think the printed output would start to look cluttered. I can change it if you think it would be better that way.

Issue #483, opened 2019-05-21
Proposal to supplement `--trim` option


jamesaoverton:
`remove` and `filter` include a `--trim <true/false>` option to determine which axioms are removed/filtered. The idea is to trim (or not) any "dangling" references.

In the implementation we build a set of target OWLObjects, then for each axiom test whether the target set intersects ([`getPartialAxioms`](https://github.com/ontodev/robot/blob/master/robot-core/src/main/java/org/obolibrary/robot/RelatedObjectsHelper.java#L147)) or contains ([`getCompleteAxioms`](https://github.com/ontodev/robot/blob/master/robot-core/src/main/java/org/obolibrary/robot/RelatedObjectsHelper.java#L92)) the set of OWLObjects for the axiom. OWLAPI has an [`OWLAxiom.getSignature`](http://owlcs.github.io/owlapi/apidocs_5/org/semanticweb/owlapi/model/HasSignature.html#getSignature--) method that returns the set of (named) OWLEntities for the axiom. But we also wanted to be able to target unnamed OWLObjects, so we wrote a [`getObjects`](https://github.com/ontodev/robot/blob/master/robot-core/src/main/java/org/obolibrary/robot/OntologyHelper.java#L338) method that starts with the `getSignature` set and adds OWLObjects based on the axiom type.

#473 shows that there are cases where we don't care about the unnamed OWLObjects, and `getSignature` would be better than `getObject`.

While maintaining backwards compatibility, we could add an option that lets you use `getSignature` instead of `getObject` and thus only consider the named OWLEntities when checking each axiom. I'm not sure what to call it: `--compare named`, `--compare entities`, `--signature named`.

A distinct but related question is whether we care about axiom annotations when checking each axiom. The current implementation of `getCompleteAxioms` will add the annotated axiom if the annotations are contained in the target set, or the unannotated axiom of only the axiom itself is contained in the target set. This suggests that we could have a more general option that would select from several comparison strategies.

dosumis:
It seems overly complex to add yet another option  &  I suspect `--signature named` will be obscure to anyone not familiar with the details of the OWL-API.  Would it not be reasonable to have 

`--strict true --select anonymous` do this?  It's what I expected that option to do in the first place.  It doesn't seem to do anything much on it's own right now.

jamesaoverton:
[...]

PR #484, opened 2019-05-22
Add --signature option to remove & filter

rctauber:
See proposal at #483 

Currently implemented as `--signature <true/false>` (`-S <true/false>`) but this can easily be changed. This works with `--trim`, for example:

```
filter --trim false --signature true
```
Returns all axioms that have **at least one** target entity in their signature. Alternatively:
```
filter --trim true --signature true
```
Returns all axioms that have signatures containing **only** the entities in the target set.
```
remove --trim false --signature true
```
Removes all axioms that have signatures containing **only** the entities in the target set. Alternatively:
```
remove --trim true --signature true
```
Removes all axioms that have **at least one** target entity in their signature. 

The default behavior is `--signature false`.


jamesaoverton:
Updated based on discussions from yesterday. As always, it was not as straightforward as I had hoped. I still need to update documentation.

jamesaoverton:
I think that the new `--signature true` option is behaving as designed. I've updated almost all the integration tests / examples to use `--signature true` and they pass. I've updated the docs to reflect the discussion in #483.

I'm going to merge this and make a release. Then we'll see what the next set of problems is.

PR #487, opened 2019-05-23
Update relax.md

cmungall:
Added more examples of combining reason-relax-reduce


cmungall:
good catch, fixed the numbering

PR #490, opened 2019-05-24
Documenting errors with equiv classes

cmungall:
Documenting the `-e` option and reasons why a use may see an error 

https://github.com/INCATools/ontology-development-kit/issues/218

### DIFF
--- a/docs/reason.md
+++ b/docs/reason.md
@@ -53,6 +53,18 @@ robot reason --reasoner ELK \
 
 If the input module contains at least one import, axioms in the debug module will be tagged with the source ontology, to assist in debugging.
 
+## Equivalent Class Axioms
+
+By default, ROBOT ignores one-to-one equivalent classes (e.g. `A EquivalentTo B`). But, in many cases, inferring equivalence between two classes is usually a sign that something has gone wrong. 
+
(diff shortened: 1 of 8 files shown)


jamesaoverton:
This is good documentation, but in reviewing it I noticed that this error is not following our current conventions. TLDR: I'll ask @rctauber to fix it, which will mean that `ReasonOperation.java` throws rather than exiting.

When we updated the documentation system, @rctauber went through and ensured that library functions all throw errors rather than just exiting. Individual commands catch the library errors and re-throw them with helpful error messages. The overall `CommandManager` will ultimately print an error message and exit. The name of the error message is used to link to the specific error on our documentation page. For example, the `ReasonerHelper.java` throws an error named `axiomGeneratorError` here:

https://github.com/ontodev/robot/blob/113bad231d98d7a78310901a03a098684a7553f2/robot-core/src/main/java/org/obolibrary/robot/ReasonerHelper.java#L28

and that links to this specific error information on the documentation page with the URL fragment matching the error name and therefore the anchor in the HTML:

http://robot.obolibrary.org/reason#axiom-generator-error

Two things:

1. This documentation system isn't documented anywhere!

2. The equivalent classes bit of the reasoning library uses `System.exit(1)`, which we're now trying to avoid:

https://github.com/ontodev/robot/blob/3ea52dfcf521ef663f078e15802b31cfeddec194/robot-core/src/main/java/org/obolibrary/robot/ReasonOperation.java#L547

Then there is no corresponding error message in `ReasonCommand.java` to catch and link to.

Issue #489, opened 2019-05-24
Update AUTHORS file


jamesaoverton:
The AUTHORS.md file is out of date. I would like to include everyone who has contributed code, i.e. everyone in the Git log and shown under the GitHub Contributors page (should be identical?). However I want to double check that every code contributor is willing to be included as an author.

I'll follow up with everyone offline, but feel free to let us know your choice in this tracker item.

rctauber:
Fine by me!

balhoff:
Sounds great.

jamesaoverton:
Closed by #495

PR #495, opened 2019-06-05
Removed out of date list of authors, link to GH instead

cmungall:
Or might we just want to delete this page, since acknowledgments is covered under README?


jamesaoverton:
Yeah, this is the simplest thing. Closes #489.

Issue #492, opened 2019-06-03
Improvements to repair


cmungall:
* Documentation can be improved (carried over from #312)
     * clarify behavior of curie literals vs IRIs in repair
     * clarify that this used the IAO replace-by property
     * document `--merge-axiom-annotations` option
 * Add ability to save report arising from repair operation
 * Add ability to annotate axioms if any IRI has been repaired
 * ?? Add an interactive mode to allow user to select whether to replace, especially when oio:consider is used (this is probably not appropriate for robot...)
 * Add ability to drive replacements from a separate TSV rather than annotation assertions (use case: obsoletions have consider axioms, user needs to decide whether to use oio:consider on a case-by-case basis, easiest done if the user can make a table)
 * Add cascading delete option. E.g. if no replaced-by, then any axiom that references the obsolete class could be marked deprecated or removed

cmungall:
Case 1: obsolete class has an xref axiom

```
ontology: xref-need-of-repair

[Term]
id: GO:1
name: leaf node
is_a: GO:2

[Term]
id: GO:2
name: original parent
is_obsolete: true
replaced_by: GO:3
xref: FOO:1

[Term]
id: GO:3
name: replacement parent
```

output of repair:

```
format-version: 1.2
ontology: xref-need-of-repair

[Term]
id: GO:1
name: leaf node
is_a: GO:3 ! replacement parent

[Term]
id: GO:2
name: original parent
xref: FOO:1
is_obsolete: true
replaced_by: GO:3

[Term]
id: GO:3
name: replacement parent
```

 * [ ] document this is the intended behavior @cmungall 
 * [ ] add an option to migrate selected (or ALL) annotation assertions, make a test using this example @balhoff 

Case: xref to an obsolete term

```
ontology: xref-need-of-repair

[Term]
id: GO:1
name: leaf node
is_a: GO:2

[Term]
id: GO:2
name: original parent
is_obsolete: true
replaced_by: GO:3

[Term]
id: GO:3
name: replacement parent

[Term]
id: FOO:1
name: external term
xref: GO:2
```

PR #510, opened 2019-06-14
Update `repair` command to allow migration of annotation values

balhoff:
See #492.

Syntax: `robot repair --input xref-need-of-repair.obo --annotation-property oboInOwl:hasDbXref --output results/xref-repaired.obo`

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/RepairOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/RepairOperation.java
@@ -109,7 +109,9 @@ public class RepairOperation {
     Set<InvalidReferenceViolation> violations =
         InvalidReferenceChecker.getInvalidReferenceViolations(ontology, true);
     repairInvalidReferences(ioHelper, ontology, violations, migrateAnnotations);
-    mergeAxiomAnnotations(ontology);
+    if (mergeAxiomAnnotations) {
+      mergeAxiomAnnotations(ontology);
+    }
   }
 
   /**
--- a/robot-core/src/main/java/org/obolibrary/robot/RepairOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/RepairOperation.java
@@ -88,7 +88,7 @@ public class RepairOperation {
       IOHelper ioHelper,
       Map<String, String> options,
       boolean mergeAxiomAnnotations) {
-    repair(ontology, ioHelper, options, false, Collections.emptySet());
+    repair(ontology, ioHelper, options, mergeAxiomAnnotations, Collections.emptySet());
   }
 
   /**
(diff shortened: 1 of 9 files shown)


balhoff:
This PR addresses @cmungall's request:

> add an option to migrate selected (or ALL) annotation assertions

However I didn't implement the ALL option. I don't think we would want to accidentally migrate `owl:deprecated` or `rdfs:label`. I suppose ALL could exclude just those two, but I think it is less magical to list the desired properties. But I am open to changing course.

jamesaoverton:
@balhoff Thanks for including automated tests. ROBOT will run the integration tests during `mvn verify` with the example code from `repair.md` and check the example files, so it should not be necessary to have the same files tested as a unit test in `RepairOperationTest.java`.

When just comparing inputs to outputs, I prefer just the integration tests. When comparing different method calls, just the unit tests are better. If that's not clear, we can add better documentation to `CONTRIBUTING.md`.

jamesaoverton:
@rctauber suggests that using an option map and resolving the annotation properties inside the operation will make life easier going forward, as more options are added to the command.

robot-core/src/main/java/org/obolibrary/robot/RepairOperation.java
```diff
@@ -52,6 +48,22 @@ public static void repair(
     repair(ontology, ioHelper, getDefaultOptions(), mergeAxiomAnnotations);
   }
 
+  /**
+   * Repairs ontology
+   *
+   * @param ontology the OWLOntology to reapir
```
cmungall: Typo!

balhoff:
@jamesaoverton @rctauber can you say more about what you mean by "option map"? Is there another command that works this way which I can look at?

Oh—I think you mean passing a Map structure to the Operation.

jamesaoverton:
Yes, I think Chris pioneered the use of `Map<String,String>` structures to pass options from a Command to an Operation. One benefit is avoiding the need to update the method signatures every time. Here's an example: https://github.com/ontodev/robot/blob/master/robot-core/src/main/java/org/obolibrary/robot/ReasonOperation.java#L43

balhoff:
@jamesaoverton the method signatures in `RepairOperation` do already have a parameter for an option map. I looked at accepting the list of annotation properties this way, but it bothers me to mix command-line parsing into the operation (this is done in my PR in the command class using `CommandLineHelper`). I like the distinction in the ROBOT readme: "They should not contain IO or CLI code." So if you think method signatures have proliferated too much (I left the previous versions to not break binary compatibility) I can adapt, but I'm inclined to leave it this way.

jamesaoverton:
@balhoff Ok, I'm fine with that. My only remaining comment is that the `properties` set is constructed but no used in `RepairCommand.java`.

balhoff:
> @balhoff Ok, I'm fine with that. My only remaining comment is that the `properties` set is constructed but no used in `RepairCommand.java`.

Thanks for catching this! 😬 I fixed that, and also another issue I discovered with `mergeAxiomAnnotations` not being read correctly. Sorry for the long delay; this should be ready to review/merge.

jamesaoverton:
Thanks @balhoff!

Issue #522, opened 2019-07-02
Extract --imports exclude does not exclude imports


matentzn:
The documentation says:

```
By default, extract will include imported ontologies. To exclude imported ontologies, just add --imports exclude for any non-MIREOT extraction method.
```

I deduced from this sentence that adding the option will extract a module only from the axioms in the root ontology. 

I have added a minimal test here:

O1:

```
B sub A
```

O2:

```
imports: O1
C sub B
```

Running:
```
robot extract -i O2.owl --term "http://www.purl.obolibrary.org/test/C" --imports exclude --method BOT --output out.owl
```
Returns an ontology with:

```
B sub A
C sub B
```

I expected to see only:

```
C sub B
```

Am I misreading the docs perhaps?

You can run the test like this:

```
git clone https://github.com/monarch-ebi-dev/robot_test.git
cd robot_test
make test_extract_import_exclude
```

matentzn:
(robot 1.4.1)

rctauber:
Thanks for catching this! This is a bug for sure.

We excluded the *entities* from the imported ontologies, but did not exclude the *axioms*, so these still got extracted. Please take a look at the fix in PR #523 if you get a chance to test it. Using your test repo, the `fail_extract_import_exclude.owl` only file includes:

    
    <owl:Class rdf:about="http://www.purl.obolibrary.org/test/B"/>

    
    <owl:Class rdf:about="http://www.purl.obolibrary.org/test/C">
        <rdfs:subClassOf rdf:resource="http://www.purl.obolibrary.org/test/B"/>
    </owl:Class>

Whereas `fail_extract_import.owl` includes:

    
    <owl:Class rdf:about="http://www.purl.obolibrary.org/test/A"/>
    
    
    <owl:Class rdf:about="http://www.purl.obolibrary.org/test/B">
        <rdfs:subClassOf rdf:resource="http://www.purl.obolibrary.org/test/A"/>
    </owl:Class>
    
    
    <owl:Class rdf:about="http://www.purl.obolibrary.org/test/C">
        <rdfs:subClassOf rdf:resource="http://www.purl.obolibrary.org/test/B"/>
    </owl:Class>

rctauber:
Resolved by #523

PR #523, opened 2019-07-02
Exclude imported axioms from module when `--imports exclude`

rctauber:
Fix for bug reported in #522


matentzn:
Cool, looks ok I think! Just FYI: I think o.getImportsClosure() includes the axioms of the ontology itself (check spec). Not that this matters, since you are dropping them in a set, but if you dropped them in a list, it would cause unnecessary memory strain.

jamesaoverton:
The code looks good. Please add a test.

PR #530, opened 2019-07-08
Add --branch-from-term to docs

rctauber:
Add a note about `--branch-from-term` option for MIREOT.


jamesaoverton:
I don't think it's correct that `--branch-from-iri` gets "all ancestors". It looks like it just calls [`getDescendants`](https://github.com/ontodev/robot/blob/master/robot-command/src/main/java/org/obolibrary/robot/ExtractCommand.java#L272).

Issue #518, opened 2019-06-24
robot reason concept of new axioms needs to be changed


cmungall:
Currently in reason we have `-a` for annotating inferred axioms, and `-n` for placing these in a new ontology.

I am responsible for these, and I under-specified these. I don't think the current behavior is useful. The intent was for both of these options to produce information on *new* axioms. With the current behavior `-a` causes *every* subClassOf to be stamped, and `-n` causes *every* axiom to be placed in a new ontology. Trivially, every axiom entails itself but this is not useful!

For example, given this an input:

```
[Term]
id: X:1
is_a: X:2
```

running reason with `-a true` gives

```
[Term]
id: X:1
is_a: X:2 {is_inferred="true"}
```

(using obo format here as it's the most compact and readable way to write simple annotated axioms but the format is irrelevant).

I think even though the docs are underspecified, this can be treated as a bug, and we can go ahead and fix and not worry about compatibility. There is no way anyone could have been *meaningfully* relying on current behavior.

I could have sworn the behavior *used* to be different, perhaps some change in the reason code had the effect on changing this, but it's on me for not producing sufficient tests.

Note: I believe @dosumis may have been confused about the behavior here: #291

jamesaoverton:
I agree that this sounds like a bug. There are some unit tests for this functionality, so it would be good to know why the bad behaviour is allowed.

balhoff:
Possibly related: when ontology A imports ontology B, I find that ROBOT inserts inferred subclass axioms into ontology A, even if they already are asserted in ontology B. I wouldn't think this is intended.

jamesaoverton:
That also sounds like a bug.

It would have been nice to get all this right the first time around, years ago. If it’s too badly broken now we can always create a new `reason2` command or something.

balhoff:
@jamesaoverton this was my mistake: I somehow missed the`--exclude-duplicate-axioms` option. I think the default should be `true` instead of `false`, but at least it's there!

balhoff:
[...]

PR #533, opened 2019-07-09
Some refactoring of ReasonOperation. 

balhoff:
Create specialized axiom generators to handle direct vs. indirect. This will eliminate some cases of materializing axioms twice.


matentzn:
Hey Jim, does this handle the fact somehow that the reason command as it currently exists actually removes asserted subsumptions when there are equivalences involved? So I currently get if:

```
A sub B
B = C
```

I might get after running the reason command:
```
A sub C
B=C
```

which is very unfortunate if you really need the A sub B to be there.. :P

jamesaoverton:
@balhoff I'm back from vacation and trying to catch up. Is this work in progress, or ready for me to review?

balhoff:
@jamesaoverton ready to review. I will add `--exclude-tautologies` in a separate PR. If you have time, also please take a look at the conversation in #518.

Issue #536, opened 2019-07-16
Error: Not in GZIP format


matentzn:
```
robot -vvv merge -I http://purl.obolibrary.org/obo/chebi.owl.gz -o chebitest.owl
```

results in 

```
2019-07-16 13:53:16,659 DEBUG org.obolibrary.robot.ExceptionHelper - Missing exception ID: Not in GZIP format
Not in GZIP format
java.util.zip.ZipException: Not in GZIP format
	at java.util.zip.GZIPInputStream.readHeader(GZIPInputStream.java:165)
	at java.util.zip.GZIPInputStream.<init>(GZIPInputStream.java:79)
	at java.util.zip.GZIPInputStream.<init>(GZIPInputStream.java:91)
	at org.obolibrary.robot.IOHelper.loadCompressedOntology(IOHelper.java:1193)
	at org.obolibrary.robot.IOHelper.loadOntology(IOHelper.java:437)
```

Is this a bug in the GZIP input reader?

I recently saw that chebi offered a gzipped version as well, so I would be really keen to be able to read it with ROBOT (1.4.1)!

rctauber:
It looks like it's a problem loading GZIP with the `--input-iri` option, since this works fine:

```
curl -Lk -o chebi.owl.gz http://purl.obolibrary.org/obo/chebi.owl.gz \
&& robot merge -i chebi.owl.gz -o chebitest.owl
```

We need to follow the redirects in a URL manually for the input stream. See PR #537

rctauber:
Resolved by #537 - please reopen if this needs more discussion.

PR #537, opened 2019-07-16
Follow redirects for gzip input IRIs

rctauber:
Right now, we just pass the `--input-iri` URL to an input stream. If the URL redirects (e.g., for a purl), this does not follow the redirect and attempts to open a GZIP input stream from the non-redirected page.

This PR adds a method to follow redirects before attempting to open the input stream for an `--input-iri` ending in `.gz`. See #536.


jamesaoverton:
@rctauber We usually use OWLAPI to handle `--input-iri`, which follows redirects. For the special case of `.gz` IRI we have custom code that was not following redirects, and this PR fixes that. Correct?

rctauber:
That's exactly correct.

PR #548, opened 2019-08-06
Add python command to access ROBOT via Py4J

beckyjackson:
# Python

The `python` command enables access of ROBOT operations and objects in Python using [Py4J](https://www.py4j.org/). This command does not accept any input ontologies and does not produce any outputs. You *cannot* chain this command with other ROBOT commands.

By default, the `python` command starts a Py4J gateway server on port `25333` (the Py4J default). To change this, use the `--port` option:
```
robot python --port 8000
```

To run this command in the background (Unix, Linux, MacOS), use:
```
robot python &
```

Once the server has started, you can dynamically access ROBOT through the gateway in a Python script. You can access any ROBOT objects through the gateway, for example, the `IOHelper` to load an ontology:
```
from py4j.java_gateway import JavaGateway
gateway = JavaGateway()
io_helper = gateway.jvm.org.obolibrary.robot.IOHelper()
ont = io_helper.loadOntology('my-ontology.owl')
```

Objects must start with `gateway.jvm` followed by the package name (e.g., `org.obolibrary.robot`).

Once finished, you can stop the gateway server from the command line with Ctrl+C. If you are running it in the background, first run the following command to bring it back to the foreground (Unix, Linux, MacOS):
```
fg
```


jamesaoverton:
I stumbled upon Py4J last week and thought we should try it out. Does anybody else have experience with Py4J to share? Do you know if it's stable and reliable?

beckyjackson:
This is now up-to-date with master. There are a few small changes to `Report.java` prompted by the use of Report objects with the OBO dashboard project (which uses Py4J):

* Added a `getViolationCount(String)` method to use for OBO dashboard. This lets you specify a rule name and get the number of violations for that specific rule. It is not used elsewhere in ROBOT, but it's a handy feature for anyone using the `robot-core` library.
* Deprecated two `Report(...)` init methods. Any time an `OWLOntology` object is passed without the `useLabels` option, the ontology is never used by the Report object. These two deprecated methods have replacements in the doc comments.
* Fixed `QuotedEntityChecker` initializing. Unless an ontology is given and `useLabels` is true, the checker will be null. If the checker is null, `maybeGetLabel(String)` returns null.

jamesaoverton:
@ontodev/robot-team We'd love to get some feedback on this. You can get a JAR for this PR and read the new docs here:

- https://build.obolibrary.io/job/ontodev/job/robot/job/py4j/ 
- https://github.com/ontodev/robot/blob/py4j/docs/python.md

jamesaoverton:
We haven't heard anything about this feature, one way or the other. I like it, so I'm going to merge it.

PR #549, opened 2019-08-07
Added Unsatisfiable Classes example to explain.md

matentzn:


jamesaoverton:
It's good to add more examples to the docs as we go. This should be quick and easy going forward. Unfortunately there are some bumps to smooth out this time.

Our documentation doubles as a set of integration tests: every indented code block that starts with `robot` is executed by the test harness. So `mvn verify` should actually run this new example command and generate a `docs/examples/results/uvular_muscle_unsat.ofn` file that must match a known-good `docs/examples/uvular_muscle_unsat.ofn` file in the repo.

This might seem weird, but I think that "executable documentation" is a good idea, and it helps ensure that our docs are correct.

However there seems to be a bug in the test harness: this example is not being processed because there code block is not followed by a line of text to terminate it. When I add some text after this example, the integration tests fail because `docs/examples/uvular_muscle_unsat.ofn` is missing.

So:

1. We can *either*:
    A. add a known-good `docs/examples/uvular_muscle_unsat.ofn` file to the repo, OR
    B. don't bother using this example as an integration test: instead of an indented code block, use a fenced code block (three backticks), and the testing harness will ignore it
2. We should fix the test harness, regardless
3. I can't find proper documentation for this test system, so we should add some

matentzn:
> but I think that "executable documentation" is a good idea

I think it is a SUPER COOL idea! 100 %. So basically the only thing I need to do is:

1. Make sure I have a new line in the end of my example (at least with a whitespace)
2. Create input files like docs/examples/uvular_muscle_unsat.ofn that I need for my added example?

jamesaoverton:
1. No, it looks like a whitespace line is not enough to make the current test harness happy. A line of unindented text would work. But we'll fix this in #551.
2. Yes, if you want this example to be executed as a test, then please add a known-good `docs/examples/uvular_muscle_unsat.ofn` file to this PR.

matentzn:
Well, lets see whether it works! :P

jamesaoverton:
Yup. I double checked it locally, too. Thanks!

PR #558, opened 2019-08-19
Optimize report for large ontologies

beckyjackson:
## Executing on Disk

`report` may fail on some very large ontologies, due to ROBOT loading the entire ontology into memory.

`report` allows you to load the ontology to a mapping file on disk with the `--tdb true` option. This is supported by [Jena TDB Datasets](http://jena.apache.org/documentation/tdb/datasets.html):

```
robot report --input edit.owl \
  --tdb true \
  --output my-report.tsv
```

Please note that this will only work with ontologies in RDF/XML (`.owl` or `.rdf`) or TTL syntax (`.ttl`). Attempting to load an ontology in a different syntax will result in a [Syntax Error](#syntax-error). ROBOT will create a directory to store the ontology as a dataset, which defaults to `.tdb`. You can change the location of the TDB directory by using `--tdb-directory <directory>`.

Once the report is complete, ROBOT will remove the TDB directory. You can include `--keep-tdb-mappings true` to prevent ROBOT from removing the TDB directory (which may be beneficial if you want to reuse it with [query](/query#executing-on-disk)). This will greatly reduce the execution time of subsequent TDB-based operations on the ontology.

The `--labels` option cannot be used with TDB, as the ontology is never loaded as an OWL API object.

## Limiting Results

Large numbers of results from the report queries may cause an `OutOfMemoryError`. To prevent this, you can limit the number of results with `--limit <INTEGER>`:

```
robot report --input edit.owl \
  --limit 10000 \
  --output my-report.tsv
```

This example will only include the first 10,000 results for each report query, meaning that some violation counts may be incomplete. Typically, you should not need to include a limit, but when working with large ontologies (using TDB), there may be hundreds of thousands of results.

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -117,6 +117,10 @@ If a prefix is incorrectly formatted, or if the prefix target does not point to 
 robot -p "robot: http://purl.obolibrary.org/robot/"
 ```
 
+### TDB Format Error
+
+`--tdb true` can only be used with RDF/XML (`.owl` or `.rdf`) or TTL syntax (`.ttl`).
+
 ### Unknown Arg Error
 
 This error message may appear for one of two common reasons:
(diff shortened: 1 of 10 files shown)


jamesaoverton:
Lack of `--labels` with TDB is a little weird. Could we support that somehow?

beckyjackson:
We could run a query to create the label map, instead of using OWLAPI. I'll try that out.

PR #564, opened 2019-09-03
Fix nested axiom annotations

beckyjackson:
Regular axiom annotations (`>A*`) work fine, but nest axiom annotations (`>>A*`) are not being processed by the new template rework. Logger was also incorrectly warning that `>C*` templates shouldn't be used, but this method is also used for axiom annotations on logical axioms.


jamesaoverton:
I think this is good. It needs a test.

PR #569, opened 2019-09-13
Fix method for axiom annotations from templates

beckyjackson:
If you have two axiom annotations next to each other:

| A prop:0 | >A prop:1 | >A prop:2 |
| --- | --- | --- |
| val 0 | val 1 | val 2 |

"val 2" will become a nested axiom annotation on the "val 1" axiom annotation, which is an annotation on "val 0". Instead, these should both be axiom annotations on the "val 0" annotation.

This is only an issue when you use two different `>` columns on one axiom.

The original template test still passes fine, but I can update the template to test different `>` columns if you'd like.

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/Template.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/Template.java
@@ -1,6 +1,7 @@
 package org.obolibrary.robot;
 
 import java.util.*;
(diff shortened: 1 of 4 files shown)


jamesaoverton:
Yes, it would be good to test this.

beckyjackson:
OK - test added

Issue #571, opened 2019-09-25
'remove' based on language attributes


allysonlister:
I would like to remove all rdfs:labels that have anything other than "anystring"@en, also preserving those rdfs:labels that have no language attribute set.

To try to get this started, I attempted to just remove those labels that have "anything"@zh as follows (a good test ontology for this is CLO at http://www.clo-ontology.org/ as it has multiple languages):

```
java -jar build/robot.jar  remove --input input.owl --term rdfs:label --select "rdfs:label=~'.*'@zh"  --output build/test.owl
```

However the above statement removes all labels whether or not they have a language set. I can't help feeling that I'm circling around the appropriate command, but I just can't see it. Please could I have some pointers?

Thanks :) Allyson

jamesaoverton:
Hmm. It looks like we don't support selection by language tags in `remove`/`filter`. We do support selection with datatypes (`^^`), so that's an oversight that we should fix. @beckyjackson Can you please add this?

The `=~` matching rule you're trying to use will only match the content of an annotation. The language information is stored separately, so that won't do what you want.

My best suggestion with current ROBOT features is to use SPARQL UPDATE:

```
java -jar build/robot.jar query --input input.owl --update delete-en.ru  --output build/test.owl
```

where `delete-en.ru` is a file with this content:

```
DELETE { ?s ?p ?o }
WHERE { ?s ?p ?o; FILTER(lang(?o) = "en") }
```

That should work, but I haven't tested it. More documentation here: http://robot.obolibrary.org/query#sparql-update

allysonlister:
Thanks very much. Will keep an eye out on the remove/filter command. In the meantime, your sparql update suggestion worked perfectly. For reference, here is the .ru file I used in the end:

```sparql
# Delete any rdfs:label where the language is set to 'zh'
prefix owl: <http://www.w3.org/2002/07/owl#>
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>

DELETE {
  ?class rdfs:label ?display_name
}
WHERE {
  ?class rdf:type owl:Class.
  ?class rdfs:label ?display_name
  FILTER(lang(?display_name) = "zh")
}
```

allysonlister:
Thanks very much! And yes, the SPARQL UPDATE worked great for me - I included the code above for reference. I'm happy for you to close the ticket - just didn't want to do it myself in case you were waiting for something internally first.

[...]

PR #574, opened 2019-10-02
Add language tags to annotation selectors

beckyjackson:
See #571 - allows selecting annotations with language tags:
`CURIE='literal'@language`, e.g. `rdfs:label='example label'@en`

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/RelatedObjectsHelper.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/RelatedObjectsHelper.java
@@ -834,6 +834,7 @@ public class RelatedObjectsHelper {
   protected static Set<OWLAnnotation> getAnnotations(
       OWLOntology ontology, IOHelper ioHelper, String annotation) throws Exception {
     OWLDataFactory dataFactory = OWLManager.getOWLDataFactory();
+
     // Create an IRI from the CURIE
     String propertyID = annotation.split("=")[0];
     IRI propertyIRI = ioHelper.createIRI(propertyID);
@@ -841,6 +842,7 @@ public class RelatedObjectsHelper {
       throw new IllegalArgumentException(
           String.format(invalidIRIError, "annotation property", propertyID));
     }
+
     // Get the annotation property and string representation of value
     OWLAnnotationProperty annotationProperty = dataFactory.getOWLAnnotationProperty(propertyIRI);
     String value = annotation.split("=")[1];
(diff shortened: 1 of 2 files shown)


robot-core/src/main/java/org/obolibrary/robot/RelatedObjectsHelper.java
```diff
@@ -947,12 +949,13 @@ private static OWLAnnotation getLiteralAnnotation(
       OWLAnnotationProperty annotationProperty,
       String value)
       throws Exception {
-    // ioHelper.addPrefix("xsd", "http://www.w3.org/2001/XMLSchema#");
     OWLAnnotationValue annotationValue;
-    if (value.contains("^^")) {
+    Matcher datatypeMatcher = Pattern.compile("'(.*)'\\^\\^([a-z]*)").matcher(value);
+    Matcher langMatcher = Pattern.compile("'(.*)'@([a-z]*)").matcher(value);
```
jamesaoverton: I don't think `[a-z]*` is general enough to handle `'foo'^^<http://foo>`, `'foo'@en-US`. It should probably be `.*`.

As usual, I'd feel better if there were some tests.

beckyjackson:
Ah, I see, I was looking at your comment. We don't have a datatype/language selector but we could do something like: `--select CURIE=@en` and `--select CURIE=^^xsd:string`?

Then you could accomplish this with:
```
robot remove --input in.owl \
  --select "rdfs:label=@en" \
  --select complement \
  --select classes \
  --include-term rdfs:label \
  --trim false \
  --axioms annotation \
  --output out.owl
```

Note that `--include-term` is coming from #570. Ideally, we would merge this first to avoid any merge conflicts...

This selects *all entities* labeled with an `@en` language tag, then gets the complement of that (so everything *without* an `@en` label). Then, it selects only the classes from that set - so you're left with all classes that do not have an `@en` label. You include `rdfs:label` in the final set of objects.

`--trim false` means that only axioms containing objects in the selected set (all those classes plus `rdfs:label`) are returned. Then, the `--axioms annotation` option makes sure *only* annotation axioms are selected, and not any logical connections between these classes.

jamesaoverton:
`--select CURIE=@en` and `--select CURIE=^^xsd:string` sound good.

I think the goal in #571 is just to remove the `rdfs:label=@en` annotations, not to remove any of the classes.

beckyjackson:
Yeah - the above command does that with the `--include-term`. I'll make the changes to add the language and datatype selectors here, but I'll wait on `--include-term` for #570.

jamesaoverton:
Ok, I reread #571 and now I think I understand. The general goal is to remove all the `rdfs:label`s that are not `@en` or plain literals. That's complicated. But there's a sub-goal that this PR will support with just `remove --select 'rdfs:label=@zh'`.

PR #577, opened 2019-10-11
Add row length error message

beckyjackson:
Add an error message when a row has more columns than the header in a template.

Also remove unused `private` method. Axiom annotations were reworked in #569 but I missed removing this old method.


jamesaoverton:
Actually I'd like to allow rows that are longer than the headers. The fix is that ROBOT shouldn't throw an error when that happens. Here's my thinking:

- if a column has a template, then it must have a header
- if a column does not have a template, then we ignore all values in that column
    - so if a column doesn't have a template, it doesn't need a header, and we ignore all values in the column

PR #578, opened 2019-10-11
Add a collapse command

beckyjackson:
# Collapse

Sometimes, a class hierarchy can contain more intermediate classes than necessary, especially when extracting modules. ROBOT includes `collapse` to remove intermediate classes based on a minimal number of subclasses, using the `--threshold` option.

```
robot collapse \
 --input module.owl \
 --threshold 3 \
 --output minimized_module.owl
```

Any intermediate class (has one or more subclasses) that has *less* than the threshold number of subclasses will be removed. Top-level classes (do not have a named superclass) and bottom-level classes (do not have any subclasses) will not be removed. 

For example, given `--threshold 2`:

```
- class:A
    - class:B
    - class:C
        - class:D
           - class:E
    - class:F
        - class:G
        - class:H
```

Becomes:

```
- class:A
    - class:B
    - class:E
    - class:F
            - class:G
            - class:H
```

`class:C` and `class:D` are removed because they each only have one subclass. `class:F` is kept because it has two subclasses, which is the threshold.

If there are any classes that you don't want removed, you can keep them regardless of the number of subclasses using `--precious <IRI/CURIE>` (for a set of terms in a file, use `--precious-terms <term-file>`). 

    robot collapse \
     --input uberon_module.owl \
     --threshold 3 \
     --precious UBERON:0000483 \
     --output results/uberon_minimized.owl

For example, given `--threshold 2` and `--precious class:D`, that same example from above would become:

```
- class:A
    - class:B
    - class:D
        - class:E
    - class:F
            - class:G
            - class:H
```


jamesaoverton:
@ontodev/robot-team Anyone have any thoughts on this implementation? Do you have code that does something similar?

jamesaoverton:
Before merging this, I'd like @beckyjackson to do a comparison with OntoFox computed intermediates for some reasonably fancy ontology.

beckyjackson:
@jamesaoverton I think the `extract --intermediates minimal` is more like OnfoFox's computed intermediates. 

Lower terms:
```
http://purl.obolibrary.org/obo/OBI_0000210
http://purl.obolibrary.org/obo/OBI_0002618
http://purl.obolibrary.org/obo/OBI_0002050
http://purl.obolibrary.org/obo/OBI_0000654
http://purl.obolibrary.org/obo/OBI_0001670
```

ROBOT method:
```
robot extract --input test.owl \
  --method mireot \
  --lower-terms terms.txt \
  --intermediates minimal \
  --output robot.owl
```

OntoFox file:
```
[URI of the OWL(RDF/XML) output file]


[Source ontology]
OBI

[Low level source term URIs]
http://purl.obolibrary.org/obo/OBI_0000210
http://purl.obolibrary.org/obo/OBI_0002618
http://purl.obolibrary.org/obo/OBI_0002050
http://purl.obolibrary.org/obo/OBI_0000654
http://purl.obolibrary.org/obo/OBI_0001670

[Top level source term URIs and target direct superclass URIs]
http://purl.obolibrary.org/obo/BFO_0000001

[Source term retrieval setting]
includeComputedIntermediates

[Source annotation URIs]
includeAllAnnotationProperties

[Source annotation URIs to be excluded]
```

These two methods produce the exact same hierarchy:
image

Using `minimize`, on the other hand, usually produces something different since you can specify the threshold. In this example, if you use `--threshold 2`, you actually get the same result since it keeps anything that has two or more subclasses:
```
robot extract --input test.owl \
  --method mireot \
  --lower-terms terms.txt \
minimize --threshold 2 \
  --output minimize.owl
```

Once you set the threshold up to `3`, it removes pretty much all intermediates because nothing has 3 superclasses.

jamesaoverton:
@ontodev/robot-team We're planning to merge this soon. Any feedback would be appreciated!

cmungall:
I don't have strong opinions. I don't really understand the motivation. I am always suspicious about counting levels as they are not often meaningful. Can you give an example of where this is to be used?

I worry a bit that it uses a quite general term `minimize` for what seems like a very niche(?) use case. What about `prune` or `truncate` or `trim` or `chop` or ...

beckyjackson:
I'm OK with changing the name, but I just want to make sure it correctly conveys that we are removing intermediate nodes, not any bottom or top-level things.

Our original use case for this was to minimize the ChEBI tree for IEDB. We have a set of "precious" terms that are the terms that we care about for the IEDB, but there are too many intermediate nodes to get there.

It does indiscriminately remove terms, although if there are important intermediate terms (e.g., carbohydrate in ChEBI) you can add them to the "precious" terms and they won't be removed.

cmungall:
Ah I get it now, for some reason I was reading `threshold` as being depth, but it's actually the direct subclass count (as you stated clearly I just didn't read closely).

OK, so I can see how this is useful for certain kinds of ontology structures where there are lots of single children. I expect ncbitaxon too. But IMO this is really just poor ontology design to have this orphan pattern (ncbitaxon ok as it's really a species taxonomy, and doesn't generally include extinct taxa). So I don't see it being used outside a handful of odd ontologies? Not to say that doesn't mean we shouldn't have it, just observing.

I don't really see a use case for values other than `2`?

jamesaoverton:
We'll look for a more specific name. When building simplified trees to show to users, we are using thresholds higher than 2.

@beckyjackson will check whether the code for the existing `extract --intermediates minimal` can be replaced as a special case of this more general code.

beckyjackson:
What about `collapse`?

wdduncan:
+1 collapse

matentzn:
+1 to collapse!

beckyjackson:
I changed the method to `collapse` and also rewrote the `extract --intermediates minimal` feature to reuse this code.

jamesaoverton:
We renamed this command and plan to merge it in the next few days. Last chance to comment!

jamesaoverton:
What about a default `--threshold` value of 2?

beckyjackson:
That makes sense to me. I just added it.

Issue #562, opened 2019-08-23
Feature request: remove --axioms "tautologies"


matentzn:
Once @balhoff [approach](https://github.com/ontodev/robot/pull/560) is merged, it would be AMAZING (<3) if we had tautology removal as a feature to prune the asserted ontology as well. I would suggest to simply extend remove --axioms to allow the "tautologies" parameter, and then refactor Jims tautology code here.

```
robot remove --axioms "tautologies"
```

Thaaaanks! :)

jamesaoverton:
This makes good sense to me.

Should we also support `robot filter --axioms tautologies`, to help find and fix tautologies?

balhoff:
@matentzn how do you want to handle `all` vs `structural` options?

matentzn:
Oh yeah. That is maybe more a question for Becky. perhaps instead of just `tautologies`, have `tautologies_structural` and `tautologies` as two alternative options..

@jamesaoverton I am down with that idea!

matentzn:
I think this works now; but the structural option is missing something I think: remove declaration of owl:Nothing. One of the primary use cases here was to get rid of owl:Nothing from our release ontologies; now `owl:Nothing sub: owl:Nothing` is removed, but `Declaration(owl:Nothing)` remains, so it still looks like this after running:

![image](https://user-images.githubusercontent.com/7070631/65970997-6fc94c00-e45f-11e9-8dfe-415cbf04d194.png)

beckyjackson:
I must have missed this issue originally, sorry for taking so long to respond. I just made a PR with an implementation in #579 

Hopefully Jenkins will be back up soon, and we would have a JAR file archived for easy testing. If you'd like me to email you a JAR for now, let me know 🙂

matentzn:
Hey @beckyjackson I will be more back to normal around 10 November forward; some pretty bad deadlines to wrestle with. Do you now remove all declarations or just the one?

beckyjackson:
Any built-in declarations, which includes OWL Nothing.

matentzn:
Built in declarations is perfect; I am a bit back and forth between "any declarations" (since they are logically empty they are kind of tautological) but just to document it here, I think built-in declarations is the better choice. I believe now that Declarations would classify better as annotation axioms rather than logical, so tautologies checkers should ignore them. 

If this could be consistent with the reasoning tautology exclusion method, that would be ideal :) Thanks!

beckyjackson:
You can see the `isTautological` method here:
[...]

PR #579, opened 2019-10-14
Add tautologies & structural-tautologies axiom selectors

beckyjackson:
See #562 
New axiom selectors:
 - `tautologies`: all axioms that are *always* true; these would be entailed in an empty ontology. WARNING: this may remove more axioms than desired.
- `structural-tautologies`: all axioms that match a set of tautological patterns (e.g., `X SubClassOf owl:Thing`, `owl:Nothing SubClassOf X`, `X SubClassOf X`)

The implementation uses features from #560 

Also fixes an issue with the extract integration tests. The mitochondrion-full file was incorrect, but we weren't catching it because the example didn't point to the results file. I did a spot check on it and this version looks correct. I can move this to a different PR if that's easier to review.

**This relies on #570 to prevent merge conflicts, but if we decide to take a different approach there I can update it.**


robot-core/src/main/java/org/obolibrary/robot/ReasonOperation.java
```diff
@@ -112,6 +112,83 @@ public static void reason(
...
+    } else if (tautologyChecker != null) {
+      if (axiom instanceof OWLAnnotationAssertionAxiom) {
+        // ignore annotation assertions
+        return false;
+      } else if (axiom instanceof OWLDeclarationAxiom) {
+        // ignore declaration UNLESS it is built in
+        OWLDeclarationAxiom declaration = (OWLDeclarationAxiom) axiom;
+        return declaration.getEntity().isBuiltIn();
+      }
+      // otherwise check if axiom is entailed by empty ontology
```
matentzn: Not sure it is necessary, but I would suggest to add a check here:

```
else if (axiom instanceof OWLLogicalAxiom) { ...}
```

I am not sure otherwise what would happen if for example irrelevant axioms (rules etc) are send to isEntailed(). Should not be a problem, but no harm here if you add it.

PR #580, opened 2019-10-22
Add integration test on OBI

beckyjackson:
Right now, this only tests for the `obi-test` branch. We need to change this to `BRANCH_NAME = 'master'` (line 14) before merging it in.

This clones OBI and uses the build ROBOT JAR to `make test`.

We can add other ontology repos to test in the future, as well, following this pattern (assuming the repo uses a `build/robot.jar` and has `make test`)
```
try {
    sh 'git clone [REPO URL]'
    sh 'mkdir -p [REPO]/build'
    sh 'cp bin/robot.jar [REPO]/build/robot.jar'
    sh 'cd [REPO] && make test'
} finally {
    sh 'rm -rf [REPO]'
}
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,14 +7,14 @@ pipeline {
     stages {
         stage('Build') {
             steps {
-                sh 'mvn clean package'
+                sh 'rm -rf bin/original-robot.jar'
             }
         }
 
         stage('Test') {
             steps {
                 script {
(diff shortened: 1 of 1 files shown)


jamesaoverton:
The result of this code is (currently) a Zipfile containing about a hundred HTML files. That's OK for debugging, but eventually we want to share this content. We could:

- commit it back to a subdirectory of this repo
- push it to some static web server
- push it to S3
- something else?

Issue #592, opened 2019-11-15
Ensure owl:deprecated has a boolean value


cmungall:
Noticed by @drseb 

Nothing in the usual owl stack will complain if you have a string value for owl:deprecated

It's kind of nuts we have to write rearguard retrospective sparqls for each of these possibilities

We need a more declarative schema here, with an owl shape and a more restricted obo shape, cc @balhoff @matentzn @dougli1sqrd @wdduncan

jamesaoverton:
I agree that we should make this change. @cmungall Do you want to assign this to someone on your team? It would be a good "getting started with ROBOT development" issue.

cmungall:
Can you add @wdduncan here https://github.com/orgs/ontodev/teams/robot-team/members?add=true -- I don't think I have permission to do so.

Bill, I think you are already familiar with the robot report framework, can you go ahead and add and configure a sparql for this.

We may want to be more strict and allow only xsd:true for a setting. The canonical way to say false is to omit the annotation. Better to have one way of doing something.

jamesaoverton:
@cmungall Ok, I added Bill to @ontodev/robot-team. You're a Maintainer of that group, so I think you have permissions to modify the team.

cmungall:
Thanks! Assigned. I think I still lack the power to get people into the org which is a prereq for team, but all good as it doesn't happen so often

wdduncan:
Looking at the directory `robot/robot-core/src/main/resources/report_queries` I don't see a query directly aimed at handling this. When I grep, I see a lot of statements like this:
```
FILTER NOT EXISTS { ?entity owl:deprecated true }
```
It seems more efficient to simply add a new query to verify owl:deprecated has a boolean values rather than adding such statements to all queries.  

@jamesaoverton If I simply add a new `.rq` file, will robot automatically use it?

beckyjackson:
I think we need to keep these statements:
```
FILTER NOT EXISTS { ?entity owl:deprecated true }
```

There are some checks that ignore deprecated entities and if we remove those statements, those checks will look at everything. For example, [missing definitions](https://github.com/ontodev/robot/blob/master/robot-core/src/main/resources/report_queries/missing_definition.rq), etc. This only works if the values are already booleans, though, so I agree that this is an important check!

[...]

PR #595, opened 2019-11-19
Ensure owl:deprecated has a boolean value #592

wdduncan:
@beckyjackson  @jamesaoverton @cmungall 

I've added a new report named `deprecated_boolean_datatype.rq` and updated the `report_profile.txt` to have level ERROR.

The query finds all entities that have a non-boolean value for an owl:deprecated annotation propery.

This addresses issue #592


jamesaoverton:
Thanks @wdduncan !

Issue #602, opened 2019-12-02
Simplify loading code for tests


jamesaoverton:
The ROBOT `master` builds on build.obolibrary.io are failing about half the time, although Travis is consistently green. For example #18 passed and #19 failed:

https://build.obolibrary.io/job/ontodev/job/robot/job/master/18/
https://build.obolibrary.io/job/ontodev/job/robot/job/master/19/

The failure logs say that various test files can't be found. The failing paths include a "%40" that I find suspicious, like this:

	/var/lib/jenkins/workspace/ontodev_robot_master%402/robot-core/target/test-classes/import_test.owl

When I compare the logs that pass to the ones that fail, the passing logs all say

	Running on Jenkins in /var/lib/jenkins/workspace/ontodev_robot_master

while the failing logs all say

	Running on Jenkins in /var/lib/jenkins/workspace/ontodev_robot_master@2

So I suspect the "@" in the path is the problem. I bet we can duplicate the problem locally by adding an `@` to the file path for a local copy of ROBOT, but I haven't tried.

Not every call to `loadOntology` is failing. Looking more closely, there are some weird differences when loading test files. We're trying to figure out the path to a class then find the `resources` directory. I forget why we do this (probably to handle catalog files?), but it's not consistent:

https://github.com/ontodev/robot/blob/master/robot-core/src/test/java/org/obolibrary/robot/CoreTest.java#L47

The fix might be as simple as escaping or unescaping in one of these calls. But it would be best for the code to be more direct, if possible.

beckyjackson:
All `CoreTest.loadOntology(...)` calls look at the resource, but it looks like some load it as a stream and some load it as a file. We load it as a file when we want to guess a catalog path. 

Instead of finding the file path, if we get the resource path as a `URI`, it handles weird characters.

beckyjackson:
We think this works, but we haven't had any master builds on the second builder since this issue first came up.

PR #603, opened 2019-12-03
Use URIs to fix resource path issues

beckyjackson:
See #602


beckyjackson:
Also, the `Jenkinsfile` had this weird command:
```
java -jar bin/robot.jar notacommand
```

This may have been a test that I forgot to remove, but it should have been the `help` message.

PR #605, opened 2019-12-23
WIP: Import configuration files & 'simple' extract

beckyjackson:
## MIREOT-RDFXML

Loading very large ontologies into ROBOT can require a lot of time and memory. Because of this, it can be easier to parse large RDF/XML files using a streaming XML processor instead of fully loading them into memory.

**Please note** that this method can only be used on files in RDF/XML format.

    robot extract --method mireot-rdfxml \
        --input uberon_fragment.owl \
        --term UBERON:0000465 \
        --term UBERON:0001017 \
        --term UBERON:0002369 \
        --output uberon_simple.owl
        
For this method, one or more `--annotation-property` options can be provided in order to add annotations to the entities. If not included, all annotations will be included in the output. To provide a text file of annotation properties, use `--annotation-properties`.

The `mireot-rdfxml` method is similar to MIREOT in that no anonymous class expressions or equivalent classes are included.

## Import Configuration File

Instead of specifying each option on the command line, ROBOT offers a `--config <file>` option in which you can specify a text configuration file. The basic format is:
```
--option-1
arg

--option-2
arg 1
arg 2
...
```

The following command-line options are supported:
* `--input` - expects one line (path to the input ontology)
* `--input-iri` - expects one line (IRI to the input ontology)
* `--output-iri` - expects one line (IRI to save the output ontology with)
* `--method` - expects one line (a valid [extraction method](#overview))
* `--intermediates` - expects one line (a valid [intermediates option](#intermediates))
* `--annotate-with-source` - expects one line (`true` or `false`, defaults to `true`)
* `--imports` - expects one line (how to [handle input imports](#handling-imports), defaults to `include`)
* `--lower-terms` - expects one or more lines (lower terms for [MIREOT](#mireot))
* `--upper-terms` - expects one or more lines (upper terms for [MIREOT](#mireot))
* `--terms` - expects one or more lines (terms for [SLME](#syntactic-locality-module-extractor-slme))

Either an `--input` or `--input-iri` is required. The `--method` is required. For MIREOT extractions, `--lower-terms` is a required option. For SLME and `mireot-rdfxml` extractions, `--terms` is a required option.

When using the `--config` option, an `--output` must be specified:

    robot extract --config uberon_config.txt \
     --output results/uberon_mireot.owl

### Using Labels

All labels are loaded from the `--input` or `--input-iri` ontology and can be used in the terms instead of CURIEs or IRIs. If you prefer, you can always reference a term by CURIE or IRI instead.

The config file expects that no term label will have a tab character in it. This is recommended practice, but if for some reason you have a label with a tab character, be sure to surround it with single quotes.

We have included a `--target` option to specify the target ontology. No terms are extracted from the target ontology, but it is loaded so that you can use labels when specifying replacement parents and annotation properties.

```
--target
path/to/my-ontology.owl
```

### Annotations

The config file also supports an `--annotations` option to specify how to handle annotations.

For MIREOT and `mireot-rdfxml` extractions, all desired annotation properties must be listed. For SLME methods, all annotation properties are always included, but the `--annotations` option can still be used for `mapTo` and `copyTo` annotations (see below).

The format for each argument is (with `\t` representing a tab):
```
<original label or ID> \t <keyword> \t <copy/replace label or ID>
```

The first `<label or ID>` is required. This is the annotation property to include when extracting (MIREOT and `mireot-rdfxml`). The `<keyword>` and second `<label or ID>` are optional. These are special operations to determine what to do with the annotation after extraction.

The keywords are:
* `copyTo` - the original annotation is kept in this case and the annotation value is copied to the given annotation property on the same subject.
* `mapTo` - the original annotation property is replaced with the new annotation property.

### Annotating With Source Ontology

Note that this option is currently only implemented with the MIREOT and `mireot-rdfxml` methods for config files.

If `--annotate-with-source` is not included, this defaults to `true`.

If `annotate-with-source` is `true`, each extracted term will be annotated with `rdfs:isDefinedBy <source-ontology-iri>`. If the source ontology does not have an IRI, this annotation is not added.

### Terms and Parents

The default parents can be replaced by specifying new parents for any of the arguments in `terms`, `lower-terms`, or `upper-terms`. The format is:

```
<label or ID> \t <new parent label or ID>
```

For all term options, the first `<label or ID>` is required. This specifies which term to extract from the input ontology. The `<new parent label or ID>` is optional. You can specify as many new parents as you'd like, separating each with a tab on the same line.

### Example

Here is an example config file to extract a set of terms from UO for OBI. These terms are specified to be children of the term 'measurement unit label', which is loaded from the `--target` ontology.

The annotations included from the input ontology are 'label' and 'definition'. The labels are copied to the 'editor preferred term'.

```
--input-iri
http://purl.obolibrary.org/obo/uo.owl

--target
src/ontology/obi.owl

--output-iri
http://purl.obolibrary.org/obo/obi/uo_import.owl

--method
MIREOT

--annotations
label       copyTo	editor preferred term
definition

--intermediates
none

--lower-terms
concentration unit	measurement unit label
frequency unit		measurement unit label
length unit		measurement unit label
```


jamesaoverton:
What do you thing about replacing `! ` with `--`, so these look more like the command-line options? Maybe that's misleading?

beckyjackson:
I just picked `!` because it wasn't used anywhere else at the start of a line, and it probably won't be used in labels. I don't see `--` being used at the start of a line anywhere, either, so I think that would be fine!

jamesaoverton:
1. I think `--` is better for this use case.
2. The OBI+UO example is a actually a bit weird. UO defined its various units as classes, but OBI decided long ago to consider specific units as instances. This code is not doing that, right?
3. I think we need comments: lines starting with `#`
4. I'm still not sure about this behaviour "Unlike MIREOT, 'simple' will not include axiom annotations on annotations." We could probably support this with a little more work. Does anyone have an opinion on this?

matentzn:
Simple already has a [meaning in the OBO world](https://github.com/INCATools/ontology-development-kit/blob/master/docs/ReleaseArtefacts.md). Would it be possible to change the name of the method to something like `skeleton` or `mireot-`, or something similar?

jamesaoverton:
Sure, I'm fine with changing the name. Maybe `mireot-rdfxml`. I'm more worried about whether skipping deep annotations will turn out to be a problem.

matentzn:
I would agree a bit with you.. I would personally like to maintain them. But on the other hand, there will be people that want this to be as simple as possible.

I vote for maintaining deep annotations, just because I perceive them conceptually as a proper part of the axioms.

jamesaoverton:
I think we should try to slurp up the `<owl:Axiom>` elements that immediately follow a Class or whatever we're actually extracting. But I'm worried that will collect too much. Oh well, let's try.

beckyjackson:
Some major changes:

1. I've replaced `! ` with `--` in the config files and updated the documentation
2. Renamed `simple` to `mireot-rdfxml` (also updated in docs)
3. Added support for `owl:Axiom` elements that have an `owl:annotatedSource` in the target IRIs for the `mireot-rdfxml` extraction method

jamesaoverton:
Thanks! We need a test for (3) that includes a nested OWL Annotation Axiom.

jamesaoverton:
This is really two features, both of which are in-progress: mireot-rdfxml and import. So I'm going to close this PR now, and @beckyjackson will open two new PRs.

PR #606, opened 2019-12-30
Switch whitespace queries to regex

beckyjackson:
The report queries to catch annotation and label whitespace only look for a space character at the start or end of the string. This misses leading/trailing tabs and newlines, which can cause issues (https://github.com/DiseaseOntology/HumanDiseaseOntology/issues/782).

This switches to a regex pattern to look for spaces, tabs, and newlines at the start or end of labels and other annotations.

### DIFF
--- a/robot-core/src/main/resources/report_queries/annotation_whitespace.rq
+++ b/robot-core/src/main/resources/report_queries/annotation_whitespace.rq
@@ -10,10 +10,10 @@ PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
 SELECT DISTINCT ?entity ?property ?value WHERE {
   {?property a owl:AnnotationProperty .
    ?entity ?property ?value .
-   FILTER STRENDS(str(?value), " ")}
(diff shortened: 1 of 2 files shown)


jamesaoverton:
I think `\s+` is better than `[ \t\r\n]+`, as long as the regex is capturing line endings (sometimes called multiline mode).

beckyjackson:
I tried this out with:
```
FILTER REGEX(str(?value), "\s+$")
```

And it seems like it returns some false positives. For example, this annotation contains new lines, but they are not at the start or end of the string:
```
Xref MGI.
OMIM mapping confirmed by DO. [SN].
Updating outdated UMLS CUI. 
Removing Caroli disease as a synonym as this is a distinct disease.
```

I would guess for labels, this is what we want to prevent, so that would be fine... but is it OK to have newlines in other annotations?

jamesaoverton:
I didn't re-check this before I merged. Is this double-escaped `\\s` correct?

beckyjackson:
Running it without the double escape results in an error, I'm not totally sure why, because the other ones (e.g., `\t`) don't require double escapes.

PR #608, opened 2020-01-20
Add maven plugin to update OBO context

beckyjackson:
[See the README](https://github.com/ontodev/robot/blob/86882e6d5dac6f4ac817c5710232921b6340a1d9/robot-maven-plugin/README.md)

We can add this to one of the default maven lifecycle phases, but I didn't want this to run every time we `mvn clean package` (in case somebody isn't connected to the internet).

Maybe `mvn site`? Or we can just run this every so often with `mvn robot:UpdateContext -N`...


jamesaoverton:
This is cool!

I'm worried that some prefixes are being dropped from the context file. Can you please do some investigating?

beckyjackson:
I'm not sure where this file originally came from, but I took a look at the original `obo_context.jsonld` file from the OBO Foundry repo and most of these are how they were in that file. Most of them are just changes to uppercase -

"ExO" -> changed to EXO
"FBsp" -> changed to FBSP
"LiPrO" -> changed to LIPRO
"miRNAO" -> changed to MIRNAO
"NIF_Cell" -> changed to NIF_CELL
"NIF_Dysfunction" -> changed to NIF_DYSFUNCTION
"VariO" -> changed to VARIO
"vHOG" -> changed to VHOG
"WBls" -> changed to WBLS

NCIT is a kind of major change, but I don't think I've seen the prefix `ncithesaurus` used:
"ncithesaurus" -> changed to NCIT

The only truly missing one is `uberon-basic`, but I can't figure out where that originally came from.

jamesaoverton:
Ok, so nothing important was dropped. But I'm still a bit worried about breaking pipelines when these prefixes change.

@cmungall What do you think of these prefix changes?

cmungall:
The only problematic one is WBls, which didn't have the preferred mixed-case prefix registered. I fixed this in the registry. You can hand-edit this one and merge this PR, or wait til the new prefixes file percolates through

beckyjackson:
I'll just change this line in the file:
```
"WBls": "http://purl.obolibrary.org/obo/WBls_"
```

PR #619, opened 2020-02-07
Add more efficient mode for `reduce` when only considering named classes.

balhoff:
This is prompted by a failure GO had when trying to update to a newer ROBOT version. The `reduce` command was failing to complete on the large go-lego ontology. This additional implementation walks the reasoner taxonomy, and so is more efficient than the previous approach, but it is also limited to only considering subclass axioms between named classes (there are some corner cases where the approach of generating a name for the anonymous expressions doesn't work with the new algorithm). There is a new command-line option `--named-classes-only` to engage the faster version.


jamesaoverton:
Looks good to me. Thanks!

PR #628, opened 2020-02-22
WIP: Adding guide on how to handle import closure with imports

cmungall:
I believe the documented behavior is true, but should be checked

### DIFF
--- a/docs/report.md
+++ b/docs/report.md
@@ -124,6 +124,16 @@ robot report --input edit.owl \
 
 This example will only include the first 10,000 results for each report query, meaning that some violation counts may be incomplete. Typically, you should not need to include a limit, but when working with large ontologies (using TDB), there may be hundreds of thousands of results.
 
+## Working with import closures
+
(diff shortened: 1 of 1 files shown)


cmungall:
Hmm, I'm not so sure how useful my advice is. I am trying to get RO to pass and it's hard.

If I don't merge, then entities are missing labels.

If I do merge, I get dupe labels (since labels for IAO_0000115 are MIREOTed in to the imported ontologies)

matentzn:
Please ignore this comment, since its pure hack alert: I think you can get around some of these issues by deleting Declarations of foreign terms. I dont remember why this works (or whether it even should work), but I think it does. Obviously, this wont work with ROBOT (since its using the OWL API for serialisation, so you cant remove Declarations), but you can hack your way with sed.

jamesaoverton:
Your advice for `report` is good, and worth adding to the docs.

The dashboard should be run on a base artifact -- that seems like a separate issue?

jamesaoverton:
@beckyjackson Can you please review this, and merge if it looks good to you?

beckyjackson (approved):

PR #645, opened 2020-03-04
Add JSON format to export

### LINKED ISSUE #631: Export basic JSON
This is one part of issue #629, adding to PR #481. We should be able to export a JSON-equivalent of TSV. This is basic JSON, not JSON-LD, so no `@id` or `@value` stuff -- that could be added later.

We should use a proper JSON library to handle escaping and stuff.

- implement as `--format json`, with `displayRenderer = RendererType.OBJECT_RENDERER`
- use an array as the top-level structure, respecting the `--sort` order
- each row is an object, with column names as keys, cells as values
- I'd prefer if keys had the same order as `--headers` too, but that might be too much to ask
- IRI/ID/CURIE must be single values
- otherwise, I think we need arrays of values; this is always a pain but I don't know what else to do

Example based on [`nucleus-sorted.csv`](https://github.com/ontodev/robot/blob/c4ad3e6eccde684eede1c3d55e9baa23dc0b3fff/docs/examples/nucleus-sorted.csv):

```
[{"CURIE": "GO:0005622",
  "LABEL": ["intracellular"],
  "SubClass Of": ["cell part"]},
 {"CURIE": "GO:0044424",
  "LABEL": ["intracellular part"],
  "SubClass Of": ["cell part", "'part of' some intracellular"]},
  ...
]
```

Feedback welcome!

beckyjackson:
Resolves #631

Add `export --format json`. This PR builds on #481.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add [`export`] command [#481] and JSON export format [#645]
+
 ### Fixed
 - Fix filtering axioms with multiple axiom selectors [#644]
 
--- a/docs/examples/nucleus.json
+++ b/docs/examples/nucleus.json
@@ -1,0 +1,169 @@
+[
+  {
+    "ID": "GO:0005575",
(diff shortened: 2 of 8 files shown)


jamesaoverton:
Thanks! We'll just wait on #481.

PR #650, opened 2020-03-08
WIP: Add RDFXML extract method

beckyjackson:
Create an `RDFXML` extraction method for big ontologies. This uses XML parsing instead of loading the ontology as an `OWLOntology` object.

Previously this was `mireot-rdfxml` but I changed it to `RDFXML` (or `rdfxml`). If people prefer the former, I'll revert it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add [`export`] command [#481]
 - Add JSON format to [`export`] in [#645]
 - Add Excel format to [`export`] in [#646]
+- Add `RDFXML` method to [`extract`] in [#650]
 
 ### Fixed
 - Fix filtering axioms with multiple axiom selectors [#644]
@@ -177,6 +178,7 @@ First official release of ROBOT!
 [`report`]: http://robot.obolibrary.org/report
 [`template`]: http://robot.obolibrary.org/template
 
+[#650]: https://github.com/ontodev/robot/issues/650
 [#646]: https://github.com/ontodev/robot/issues/646
 [#645]: https://github.com/ontodev/robot/issues/645
 [#644]: https://github.com/ontodev/robot/issues/644
--- a/docs/examples/uberon_simple.owl
+++ b/docs/examples/uberon_simple.owl
@@ -1,0 +1,1059 @@
+<?xml version="1.0"?>
+<rdf:RDF xmlns="http://purl.obolibrary.org/obo/uberon.owl#"
+     xml:base="http://purl.obolibrary.org/obo/uberon.owl"
+     xmlns:obo="http://purl.obolibrary.org/obo/"
+     xmlns:owl="http://www.w3.org/2002/07/owl#"
+     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+     xmlns:xml="http://www.w3.org/XML/1998/namespace"
+     xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
+     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
+     xmlns:oboInOwl="http://www.geneontology.org/formats/oboInOwl#">
+    <owl:Ontology rdf:about="http://purl.obolibrary.org/obo/uberon.owl"/>
+    
+
+
+    <!-- 
+    ///////////////////////////////////////////////////////////////////////////////////////
+    //
+    // Annotation properties
+    //
+    ///////////////////////////////////////////////////////////////////////////////////////
+     -->
+
+    
+
+
+    <!-- http://purl.obolibrary.org/obo/IAO_0000115 -->
+
+    <owl:AnnotationProperty rdf:about="http://purl.obolibrary.org/obo/IAO_0000115"/>
+    
+
+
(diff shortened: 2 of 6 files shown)


jamesaoverton:
I prefer `mireot-rdfxml`. Any other opinions?

jamesaoverton:
This code uses a streaming XML reader to rip apart an RDFXML and then create new axioms in an OWLOntology. I was thinking that it would be better to output to an streaming XML writer, just copying selected elements, which should be much more efficient. However if this is used in a chain then ROBOT would have to read the output RDFXML into an OWLOntology anyway. So we'd have to check whether it was being used in a chain, which is kind of ugly.

Thoughts?

cmungall:
Curious about this. No rush to respond to these Qs.

 - how much does this rely on the particular serialization used by the OWLAPI? I have seen some pretty bizarre valid serializations as rdf/xml (e.g older versions of jena)
 - it looks like this is intended for only traversal between subClassOf between named classes? I'm not sure it makes sense to use Uberon as an example in the docs as we encourage traversal over at least the partonomy. NCBITaxon would be a great example.
 - it also looks like there is an assumption the ontology is `relax`ed? E.g. it won't traverse over `AND`s?

jamesaoverton:
@beckyjackson Should reply in detail.

I'm thinking of this as an optimization for MIREOT, and thus OK with it being fragile. So I'm OK if this only works for RDFXML generated by OWLAPI, and expects relaxed input.

I take your point that Uberon is not a great example for this case. We use Uberon for a lot of ROBOT examples (because it's great!) and I bet that Becky was just trying to reuse existing examples.

beckyjackson:
> how much does this rely on the particular serialization used by the OWLAPI? I have seen some pretty bizarre valid serializations as rdf/xml (e.g older versions of jena)

I would say it relies pretty heavily on the OWLAPI serialization, though I have not tested it out with others. If you have a file you can point me to, I can test it out.

> it looks like this is intended for only traversal between subClassOf between named classes? I'm not sure it makes sense to use Uberon as an example in the docs as we encourage traversal over at least the partonomy. NCBITaxon would be a great example.

As James said, I was just trying to reuse existing examples. I have nothing against changing the examples, but it would just require adding more files to the repo.

> it also looks like there is an assumption the ontology is relaxed? E.g. it won't traverse over ANDs?

Correct.

cmungall:
Thanks.

Do you think it is worth coming up with an XML schema for owlapi-rdfxml, if we are to rely on it?

Here is a thought (inspired by my initially horrified reaction at Bill picking up my old hacky [obo scripts](https://github.com/cmungall/obo-scripts) ):

Why not use obo as a target rather than a specific variant of rdf/xml?

 - there is no guarantee that the rdf/xml version of an ontology will be serialized by the owlapi. We don't mandate this in OBO.
 - even if it is, we don't have an actual spec for this specific serialization. There is no guarantee the owlapi will continue to use the same idioms, and there could potentially be breaking changes. The obo spec may be less rigorous than a w3 spec, but at least there is a spec!
 - obo format is optimized for time and memory-efficient streaming, and the file size is considerably smaller
 - extending the behavior to relational graphs should be easy, whereas this gets potentially awkward for rdf/xml
 - most ontologies in OBO have an obo format version as part of their standard release, and if they don't it should be easy to add
 - on the downside, we have previously talked about more formally obsoleting obo format. But note that many in the RDF community talk about obsoleting rdf/xml. And while obo format does have it's flaws, it was developed with certain requirements and optimizations in mind, and doesn't do a bad job with those.

Perhaps my worries about using owlapi-rdf/xml are unfounded in pragmatic terms. But I still have mental scars from the early days of OWL, battling with attempting to use XML-level tools such as XSLT in totally inappropriate ways..

To give an example, this is a perfectly valid of rendering some class axioms. This isn't particularly perverse, a totally valid sane RDF-level renderer could render like this if it doesn't attempt to follow nesting. And this doesn't even get into perversities like `rdf:ID` or explicit blank nodes for existentials... (shudder). In fact predicting correct behavior is even harder given that the OWLAPI makes inferences in mapping to OWL. You could leave out the rdf:type declaration altogether and it will infer we have a class. Is the file still valid? It depends....

```
    <rdf:Description rdf:about="http://purl.obolibrary.org/obo/ZFA_0005587">
        <obo:IAO_0000115 rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Cluster of cells comprising a portion of tissue which gives rise to new mesonephric nephrons.  The cluster contains self-renewing stem cells.</obo:IAO_0000115>
        <rdfs:label rdf:datatype="http://www.w3.org/2001/XMLSchema#string">nephron progenitor</rdfs:label>
    </rdf:Description>

    <!--  ... -->
    <!-- other random triples here ... -->
    <!--  ... -->

    <rdf:Description rdf:about="http://purl.obolibrary.org/obo/ZFA_0005587">
        <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/ZFA_0001477"/>
    </rdf:Description>

    <rdf:Description rdf:about="http://purl.obolibrary.org/obo/ZFA_0005587">
        <rdf:type rdf:resource="http://www.w3.org/2002/07/owl#Class"/>
    </rdf:Description>
```

jamesaoverton:
@cmungall You make good points. I'll give this some thought.

jamesaoverton:
I'm no longer convinced that this is a good idea, and I think we'll go in another direction. So I'm closing this PR.

PR #652, opened 2020-03-10
WIP: Add config option to extract [WAIT FOR #650]

### LINKED ISSUE #212: Better import configuration

beckyjackson:
Builds on #650, resolves #212

## Import Configuration File

Instead of specifying each option on the command line, ROBOT offers a `--config <file>` option in which you can specify a text configuration file. The basic format is:
```
--option-1
arg

--option-2
arg 1
arg 2
...
```

The following command-line options are supported:
* `--input` - expects one line (path to the input ontology)
* `--input-iri` - expects one line (IRI to the input ontology)
* `--output-iri` - expects one line (IRI to save the output ontology with)
* `--method` - expects one line (a valid [extraction method](#overview))
* `--intermediates` - expects one line (a valid [intermediates option](#intermediates))
* `--annotate-with-source` - expects one line (`true` or `false`, defaults to `true`)
* `--imports` - expects one line (how to [handle input imports](#handling-imports), defaults to `include`)
* `--lower-terms` - expects one or more lines (lower terms for [MIREOT](#mireot))
* `--upper-terms` - expects one or more lines (upper terms for [MIREOT](#mireot))
* `--terms` - expects one or more lines (terms for [SLME](#syntactic-locality-module-extractor-slme))

Either an `--input` or `--input-iri` is required. The `--method` is required. For MIREOT extractions, `--lower-terms` is a required option. For SLME and RDFXML extractions, `--terms` is a required option.

When using the `--config` option, an `--output` must be specified:

    robot extract --config uberon_config.txt \
     --output results/uberon_mireot.owl

### Using Labels

All labels are loaded from the `--input` or `--input-iri` ontology and can be used in the terms instead of CURIEs or IRIs. If you prefer, you can always reference a term by CURIE or IRI instead.

The config file expects that no term label will have a tab character in it. This is recommended practice, but if for some reason you have a label with a tab character, be sure to surround it with single quotes.

We have included a `--target` option to specify the target ontology. No terms are extracted from the target ontology, but it is loaded so that you can use labels when specifying replacement parents and annotation properties.

```
--target
path/to/my-ontology.owl
```

### Annotations

The config file also supports an `--annotations` option to specify how to handle annotations.

For MIREOT and RDFXML extractions, all desired annotation properties must be listed. For SLME methods, all annotation properties are always included, but the `--annotations` option can still be used for `mapTo` and `copyTo` annotations (see below).

The format for each argument is (with `\t` representing a tab):
```
<original label or ID> \t <keyword> \t <copy/replace label or ID>
```

The first `<label or ID>` is required. This is the annotation property to include when extracting (MIREOT and RDFXML). The `<keyword>` and second `<label or ID>` are optional. These are special operations to determine what to do with the annotation after extraction.

The keywords are:
* `copyTo` - the original annotation is kept in this case and the annotation value is copied to the given annotation property on the same subject.
* `mapTo` - the original annotation property is replaced with the new annotation property.

### Annotating With Source Ontology

Note that this option is currently only implemented with the MIREOT and RDFXML methods for config files.

If `--annotate-with-source` is not included, this defaults to `true`.

If `annotate-with-source` is `true`, each extracted term will be annotated with `rdfs:isDefinedBy <source-ontology-iri>`. If the source ontology does not have an IRI, this annotation is not added.

### Terms and Parents

The default parents can be replaced by specifying new parents for any of the arguments in `terms`, `lower-terms`, or `upper-terms`. The format is:

```
<label or ID> \t <new parent label or ID>
```

For all term options, the first `<label or ID>` is required. This specifies which term to extract from the input ontology. The `<new parent label or ID>` is optional. You can specify as many new parents as you'd like, separating each with a tab on the same line.

### Example

Here is an example config file to extract a set of terms from UO for OBI. These terms are specified to be children of the term 'measurement unit label', which is loaded from the `--target` ontology.

The annotations included from the input ontology are 'label' and 'definition'. The labels are copied to the 'editor preferred term'.

```
--input-iri
http://purl.obolibrary.org/obo/uo.owl

--target
src/ontology/obi.owl

--output-iri
http://purl.obolibrary.org/obo/obi/uo_import.owl

--method
MIREOT

--annotations
label       copyTo	editor preferred term
definition

--intermediates
none

--lower-terms
concentration unit	measurement unit label
frequency unit		measurement unit label
length unit		measurement unit label
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add [`export`] command [#481]
 - Add JSON format to [`export`] in [#645]
 - Add Excel format to [`export`] in [#646]
+- Add `RDFXML` method to [`extract`] in [#650]
 
 ### Fixed
 - Fix filtering axioms with multiple axiom selectors [#644]
@@ -177,6 +178,7 @@ First official release of ROBOT!
 [`report`]: http://robot.obolibrary.org/report
 [`template`]: http://robot.obolibrary.org/template
 
+[#650]: https://github.com/ontodev/robot/issues/650
 [#646]: https://github.com/ontodev/robot/issues/646
 [#645]: https://github.com/ontodev/robot/issues/645
 [#644]: https://github.com/ontodev/robot/issues/644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add JSON format to [`export`] in [#645]
 - Add Excel format to [`export`] in [#646]
 - Add `RDFXML` method to [`extract`] in [#650]
+- Add `--config` option to [`extract`] in [#652]
 
 ### Fixed
 - Fix filtering axioms with multiple axiom selectors [#644]
@@ -178,6 +179,7 @@ First official release of ROBOT!
 [`report`]: http://robot.obolibrary.org/report
 [`template`]: http://robot.obolibrary.org/template
 
+[#652]: https://github.com/ontodev/robot/pull/652
 [#650]: https://github.com/ontodev/robot/issues/650
 [#646]: https://github.com/ontodev/robot/issues/646
 [#645]: https://github.com/ontodev/robot/issues/645
--- a/docs/examples/uberon_config.txt
+++ b/docs/examples/uberon_config.txt
@@ -1,0 +1,15 @@
+--input
+uberon_fragment.owl
+
+--method
+MIREOT
+
+--annotate-with-source
+false
+
+--lower-terms
+UBERON:0001017
+UBERON:0002369
+
+--upper-terms
+UBERON:0000465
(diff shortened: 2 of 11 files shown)


jamesaoverton:
Since I closed #650, I'll close this too. We might want to come back to this idea, but then I would want to generalize it to work for all ROBOT commands.

PR #656, opened 2020-03-23
Change OBO provider to CURIE provider

beckyjackson:

Fix provider used for short form IDs in `export` to match the IRI to longest namespace possible. If no match is found in prefixes, return the full IRI and log an error.


jamesaoverton:
I'd like a few more unit tests for the CURIE provider, please. We should have tests for "overlapping" prefixes to be absolutely sure that the longer base is used.

beckyjackson:
@jamesaoverton - I've added a test for a few IRIs. Please let me know if you want any more tested.

jamesaoverton:
@beckyjackson That covers the important cases. Thanks!

@ontodev/robot-team This change addresses a problem we had in `export` where the current ShortFormProviders will just make up a prefix if no matching prefix is defined. This code will log an error and return the IRI string if it can't find a prefix.

Should it just be a warning? Should we fail harder?

jamesaoverton:
No comments in more than a week, so I'm going to merge.

PR #675, opened 2020-04-22
Changed invalid reference violations to warnings.

balhoff:
This is a tiny change to make invalid reference violation output logged as WARN instead of ERROR.  This is a frequent source of confusion (see #674). I think WARN is more accurate, because these messages are printed every time I use `robot reason`, and I still continue with the results.


jamesaoverton:
I spoke to @cmungall about this. He said he's fine with downgrading these warning, so I'll merge.

He would like a new option to fail if these violations are found -- I added that as issue #680.

PR #683, opened 2020-04-27
Render annotation value IRIs correctly

beckyjackson:

Minor problem found while using `export` on OBI. I would expect annotation value IRIs to render based on the tag in the header (`ID`, `LABEL`, etc.), but they always appear as IRIs because there is no check for `isIRI()` in parsing annotation values.

This fixes it so that entities can be rendered correctly when they appear as annotation values.

I don't think this requires updates to the changelog.


beckyjackson:
`mvn verify` passed locally for me, but for some reason I'm not seeing the Travis results on this PR...

PR #687, opened 2020-05-07
Add entity-select option with fine-grained control to export

### LINKED ISSUE #682: Allow per-column control for named/anonymous in export
The new `export` command has a `--exclude-anonymous` option that applies to the whole sheet. We need to allow per-column overrides the global option.

This only applies to logical columns. Two options that occur to me are:

1. add a "Named" variant to each local column type: "Named SubClassOf"
2. add a "named" option to the `[options]` for the column: "SubClassOf [ID named]"

Other suggestions?

beckyjackson:
Resolves #682

---

The export will include both named entities and anonymous expressions (subclasses, equivalent classes, property expressions). You can change this with the `--entity-select` option:
* `--entity-select ANY`: include both named and anonymous expressions in all columns (default)
* `--entity-select NAMED`: include only named entities in all columns
* `--entity-select ANON` or `--entity-select ANONYMOUS`: include only anonymous expressions in all columns

For example:

    robot export --input nucleus_part_of.owl \
      --header "LABEL|SubClass Of|part of" \
      --entity-select NAMED \
      --export results/nucleus.csv

Note that in the example above, the first two headers are special keywords and the third is the label of a property used in the ontology.

These export-wide defaults can be overridden in specific columns by including the keyword in a square-bracket-enclosed tag following the column name:
* `col name [ANY]`: include both named entities and anonymous expressions in this column
* `col name [NAMED]`: include only named entities in this column
* `col name [ANON]`/`col name [ANONYMOUS]`: include only anonymous expressions in this column

For example:

    robot export --input nucleus_part_of.owl \
      --header "LABEL|SubClass Of [ANON]" \
      --export results/nucleus_anon.csv
      
Without the `ANON` tag in the example above, `SubClass Of` would include both anonymous expressions and named classes. Now, it only includes the anonymous expressions.

These tags can be combined with the [rendering tags](#rendering-cell-values), for example:
* `col name [NAME ANON]`: renders anonymous expressions using the name (label or CURIE) of the entities used in the expression
* `col name [ID ANON]`: renders anonymous expressions using the CURIE of entities used in the expression
* `col name [ID NAMED]`: renders named entities only using their CURIE
* `col name [NAME ANY]`: renders with both named entities and anonymous expressions

It is not recommended to use the `ANON` tag in combination with `LABEL`, as if an entity used in an anonymous expression does not have a label, it will be rendered as an empty string. Additionally, the entity selection tags will have no effect on annotation property values.


robot-core/src/main/java/org/obolibrary/robot/ExportOperation.java
```diff
@@ -126,13 +143,27 @@ public static Table createExportTable(
...
     for (String c : columnNames) {
-      String tag = entityFormat;
+      String currentEntityFormat = entityFormat;
+      String currentEntitySelect = entitySelect;
       String colName = c;
       // Determine if this has a tag for rendering
-      Matcher m = Pattern.compile("(.+) \\[(ID|id|IRI|iri|LABEL|label|NAME|name)]").matcher(c);
+      Matcher m =
+          Pattern.compile(
+                  "(.+) \\[(ID|id|IRI|iri|LABEL|label|NAME|name|NAMED|named|ANON|anon|ANONYMOUS|anonymous).*]")
```
jamesaoverton, line 152: Please just use a case insensitive regex.

jamesaoverton:
There's a missing file: Integration test file '../docs/examples/results/nucleus_anon.csv' does not have a corresponding example file '../docs/examples/nucleus_anon.csv'

Looks like I have to update Travis.

beckyjackson:
Fixed!

Issue #387, opened 2018-10-26
Support DL query


jamesaoverton:
We should be able to run DL queries using the OWLAPI.

This might work as an extension of `query`.

@hoganwr has an implementation here that we can look at: 
https://github.com/ufbmi/dron-query

balhoff:
What do you think about integrating [Owlet](https://github.com/phenoscape/owlet) into this? It would give you a way to return other stuff along with just the term IRIs (e.g. labels, definitions). With Owlet you could just smuggle the DL queries into the existing SPARQL `query` command. If you are interested I could make some cleanups to Owlet, since it hasn't been touched for a while.

jamesaoverton:
@balhoff I’d be happy to see Owlet in ROBOT. I think it should be a separate issue.

balhoff:
Sounds good.

cmungall:
+1

hoganwr:
Note that I just updated dron-query to be able to accept multiple DL queries from the command line, and direct the output of each query to multiple outputs.  In effect, this just means that the --query and --output parameters are now multiple cardinality.  

However, the benefit is vast: for DrOn at least, a full reasoning with Hermit or Fact++ takes on the order of hours.  Having to do that every time for each query is inefficient.  Being able to send multiple queries to run after just one full reasoning is much more efficient.

jamesaoverton:
Very nice.

FYI, we're working on a `robot validate` command using DL query. Work in progress is here:

- draft documentation https://github.com/lmcmicu/robot/blob/add_validate_operation/docs/validate.md
- development branch https://github.com/lmcmicu/robot/tree/add_validate_operation

cmungall:
What's relationship of this to shape languages?

jamesaoverton:
We're using DL query and OWLAPI, so I'd say "no relationship".

cmungall:
I meant more on the use case - is this for validating ABoxes against a kind
of schema?

jamesaoverton:
The main thing I want to do is validate ROBOT templates before building them, so mostly TBoxes. They can contain class expressions in Manchester syntax. The simple case is to make sure all the values in a column are subclass of X, e.g. the OBI assay template has a "Device" column, and if a cell in that column has a value then it the validator enforces that it must be a subclass of the OBI 'device' class. We also have some fancier inter-column validation, e.g. when a value in column A is subclass of X then the corresponding value in column B is part of some Y.

[...]

PR #691, opened 2020-05-12
Add validate operation

### LINKED ISSUE #636: Revise `validate` to use `export` output code
@lmcmicu implemented a new `robot validate` command, based on #387. I've been using this for real work already, and I love it.

One thing this does is output to HTML and Excel tables, with highlighted cells. I want to reuse the new `export` output code instead. There's no need for `validate` to export to CSV or TSV, but maybe JSON and other formats would work if they can support comments.

beckyjackson:
Resolves #636


beckyjackson:
I've made some updates to how `validate` works. I'm happy to rework some of these if they don't sound good, but I think these changes work a bit better with our standard `Makefile` workflows...

* By default, `validate` will fail if there are any validation failures in any table (exit code `1`). This is so `validate` can be used in `Makefile` workflows and stop the workflow on error. I put in a new `--no-fail` option so this can be overridden.
* I added tooltips to the HTML tables, which requires the Bootstrap JS. Bootstrap JS is only included in the HTML file if there is at least one comment. To this extent, I changed `--standalone` to default to `false` so that users will always get the CSS and JS unless they say otherwise.
* I added in "not" rules (`not-subclass-of`, `not-superclass-of`, `not-instance-of`). I originally tried `subclass-of not X` but this did not properly validate.
* `validate` will now print an error message at the end on any validation failures, and print the names of the table(s) that failed with the path to the output (unless no output format):
```
VALIDATION FAILED - the following table(s) had one or more rule violation:
- immune_exposures.csv (results/immune_exposures.txt)
```
* You can choose to print all validation failure messages with `--silent false` (messages will always be printed if there is no output format)

beckyjackson:
Also, this has a fix to `template` - the rule rows were being parsed even though they don't have an ID/LABEL.

cmungall:
Perhaps this should be called something like `validate-table` or `validate-template-data`? `validate` sounds like it's validation of the ontology itself, which is covered by `report`.

Or maybe this should be a flag in the `template` command, if I understand the intent correctly.

I had a look at the comments on the DL-query ticket
https://github.com/ontodev/robot/issues/387#issuecomment-538524519

and at `validate.md` in this PR. I thought I grokked the overall idea, but maybe I'm being dense, I think I need a gentler introduction as to what the overall use case is here, sorry

matentzn:
Wait, I thought that the tables contain validation rules that are checked against the `-i` input ontology? So my worry was more whether this `validate` here clearly delineates from `report`. (is it that `report` is for making generic structural checks, and `validate` being a kind of unit testing framework?) The way I always understood this was more like a kind of shacl/shex thing for TBoxes: ensure that a `phenotype` is not subclass of a `disease`, but a subclass of say `dependent continuant`. And so on. Can we get a few examples/use cases explained in a non-technical way to understand how we will use it?

jamesaoverton:
Yes, we can update the documentation.

This `validate` command accepts a table, then returns the table with problem cells highlighted. So it's is quite different from `report` in behaviour and implementation. It can run on ROBOT templates (parts of the ontology) or just tables of data, so it's distinct from `template` as well.

cmungall:
This sounds pretty useful. Does it belong as part of robot? I'm not saying
it shouldn't be, more than happy for it to be here (maybe with the name
validate-table). But there may be advantages to you having it be
independent with its own release cycle, perhaps imported where a subset of
its functionality is exposed in template validation. Somewhat analogous to
owl diff. I can see a general owl enabled table validator evolving into an
incredibly powerful tool in its own right.

beckyjackson:
As a side note, I updated the docs to try and clarify the input/output files. I also put details on the HTML `--standalone` option, and added a new `--write-all` option. The `--write-all true` writes all tables to the output directory. When `--write-all false` (default), only tables with failed validations are written to the output directory. I think this will be easier for using in `Makefile` workflows where people only want to see what has a problem.

docs/validate.md
```diff
@@ -1,31 +1,304 @@
...
+
+`validate` will only print a summary message at the end if there were any failures (unless no `--format` is specified, in which case it will always print to STDOUT). If you would like to print all invalid data messages, include `--silent false`.
+
+#### Output Files (`--write-all`)
+
+`validate` will only write tables with failed validations to the output directory. If you wish to write _all_ tables, including those that did not have any failed validation, specify `--write-all true`.
+
+#### HTML Tables (`--standalone`)
+
+If the output format is HTML, all output tables will be written as "standalone" files. This means that they have a header containing the Bootstrap stylesheet and scripts ([for tooltips](https://getbootstrap.com/docs/4.5/components/tooltips)). If you want to plug the table data into an existing HTML file, you can use `--standalone true` to generate _just_ the table element.
```
jamesaoverton, line 68: `--standalone true`  should mean a complete HTML file. `--standalone false` should mean just the HTML table. (This is the usage I've seen in lots of other tools, e.g. pandoc.) So this documentation is the opposite of what I would l like -- I'm not sure about the implementation.

jamesaoverton:
There's a new file `robot-core/test.txt` that should not be there.

jamesaoverton:
Are the two `robot-core/src/test/resources/nucleus.*` files being used? They're new in this PR, but I didn't seem them used in this PR.

jamesaoverton:
In the `immune_exposures.csv` example files, I don't understand why the IDs use underscores, e.g. `NCBITaxon_11103`. They should be proper CURIEs. This might be a holdover from a long time ago, but it should be fixed before merging. And there should not be any special code anywhere for pretending that these are actually CURIEs -- they are wrong and they should fail.

In the `immune_exposures.html` output file, if the input value is an ID/CURIE, then the HTML cell should show that same value. The HTML cells values should always match the input values, except that we add links when we can. 

I would like `immune_exposures.owl` cleaned up to remove everything in the "mike" namespace, which was just for development. I think that the only object property we actually need is IDO:0000664 'has material basis in'. Most of the other classes are in ONTIE. This is much closer to what we want: https://github.com/jamesaoverton/immune-exposure-validation

beckyjackson:
I removed those couple of files that are not used anywhere. I also cleaned up the immune exposures test files.

> And there should not be any special code anywhere for pretending that these are actually CURIEs -- they are wrong and they should fail.

It looks like the parser from OWLAPI is actually OK with these. I changed them in the file, but do we want to add a piece of code that confirms that CURIEs are valid? Otherwise it won't fail on CURIEs that just have an underscore.

> The HTML cells values should always match the input values, except that we add links when we can.

Agreed and updated, **except** for anonymous expressions. The HTML renderer handles these and I think if we want to keep using the exact values, we would have to write a new renderer, or at least rework something... what do you think?

docs/validate.md
beckyjackson: You're right. This is how it's implemented, the `--standalone true` in the docs was just a typo. I changed this to say `--standalone false`. The default behaviour is `true`.

jamesaoverton:
This has performance problems, and I want to add an alternative format for specifying the rules. But it's blocking other PRs, so I'm going to merge it, and we'll continue to improve it.

As to whether it belongs in ROBOT: This works hand-in-hand with `template`, which is 50%+ of what I use ROBOT for. It also reuses outputs code for `export` and other commands. So I am confident that it belongs.

As to whether the name `validate` is too general: The focus is currently on validating a target table given DL query rules and an input ontology then output a table of results, but I think there's lots of room to extend this starting point to validate a target graph or OWL file, against SHACL rules or something.

PR #692, opened 2020-05-12
Export fix 2

### LINKED ISSUE #682: Allow per-column control for named/anonymous in export

beckyjackson:
Resolves #682

- [x] tests added

Fix regex pattern for tags to capture multiple tags

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/ExportOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/ExportOperation.java
@@ -157,9 +157,9 @@ public class ExportOperation {
         String tag = m.group(2);
         // Process one or more tags
         for (String subTag : tag.split(" ")) {
-          if (entityFormatTags.contains(subTag.trim().toLowerCase())) {
+          if (entityFormatTags.contains(subTag.toLowerCase())) {
             currentEntityFormat = subTag;
-          } else if (entitySelectTags.contains(subTag.trim().toLowerCase())) {
+          } else if (entitySelectTags.contains(subTag.toLowerCase())) {
             currentEntitySelect = subTag;
           } else {
             throw new Exception(String.format(unknownTagError, c, subTag));
--- a/robot-core/src/main/java/org/obolibrary/robot/ExportOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/ExportOperation.java
@@ -149,7 +149,7 @@ public class ExportOperation {
       // Determine if this has a tag for rendering
       Matcher m =
           Pattern.compile(
-                  "(.+) \\[(id|iri|label|named|name|anon|anonymous|any) ?.*]",
+                  "(.+) \\[((id ?|iri ?|label ?|named ?|name ?|anon ?|anonymous ?|any ?)+)]",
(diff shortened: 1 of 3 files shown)


robot-core/src/main/java/org/obolibrary/robot/ExportOperation.java
```diff
@@ -149,7 +149,7 @@ public static Table createExportTable(
       // Determine if this has a tag for rendering
       Matcher m =
           Pattern.compile(
-                  "(.+) \\[(id|iri|label|named|name|anon|anonymous|any) ?.*]",
+                  "(.+) \\[((id ?|iri ?|label ?|named ?|name ?|anon ?|anonymous ?|any ?)+)]",
```
jamesaoverton, line 152: I think the regex should be simplified to just capture anything inside `[]`. Then the code below should:

- lowercase the string
- split on whitespace
- remove `entityFormatTags` and `entitySelectTags`
- throw an exception if there's anything left: unknown tags

Maybe also throw an exception if there's more than one of each type of tag.

beckyjackson: My only concern was if a label used square brackets for anything, so I was trying to be a bit more strict.

jamesaoverton: I could be overlooking something, but the square bracket stuff only makes sense for logical columns, for which the cell values are entities or expressions. We have a fixed list of those column names, right?

You column name can have be the label of an arbitrary annotation property, but then the cells will contain the annotation value, with no messing around. Annotation values can't be entities. If the annotation value is an IRI, the cell should contain the IRI.

So I think we only need to parse the square bracket stuff if the column name is on our list of logical columns. In which case we can be strict about the contents of the square brackets.

beckyjackson: I updated the method to look at any contents of the square brackets, but I don't think we can have a standard list of logical columns, since you can use properties as the column headers as well.  I could be wrong, though?

jamesaoverton: You're right, I was missing the case where the column header is the label of an object property. I'm not too worried about that. You could check to see if the column name matches an object property label, and if it does don't try to process the `[]`. Alternatives tka make the bracketed string more distinct include: start with `[]`, double it `[[ID]]`, use weirder characters, ... Do you have a preferred alternative?

Your updated code still allows multiple tags of each type, and just quietly accepts the last one it sees.

beckyjackson: I think the square brackets is probably the easiest for users, and I don't think I've ever seen square brackets used for properties so I'm tempted to just ignore it... This will only catch square brackets at the end, as well, so if a property uses them in the middle it's OK.

PR #699, opened 2020-05-27
Add more formats to report, including HTML

### LINKED ISSUE #620: HTML output for `report`
I'm 99% sure that I would like HTML output added to `robot report`. By default it should just output a `<table>`:

```
<table class="table">
  <tr>
    <th>Row</th>
    <th>Level</th>                                                             
    <th>Rule Name</th>
    <th>Subject</th>    
    <th>Property</th>
    <th>Value</b</th>
  </tr>
  <tr class="table-danger">
    <td>1</td>
    <td>ERROR</td>
    <td><a href="http://robot.obolibrary.org/report_queries/missing_label">missing_label</a></td>
    <td><a href="http://purl.obolibrary.org/obo/GO_0008150">GO_0008150 [GO:0008150]</a></td>
    <td><a href="http://www.w3.org/2000/01/rdf-schema#label">label [rdfs:label]</a></td>
    <td></td>
    <td></td>
  </tr>
</table>
```

With a `--standalone true` option it should output a full HTML document. The Dashboard basically does this in Python, but I think it's better in ROBOT.

beckyjackson:
Resolves #620 
Also adds JSON and XLSX formats to `report`.

Add more formats to `report` using the `export` Table object. This will allow us to add even more formats as we add them to `export`. Note that for JSON and YAML formats, we don't use the `toJSON` from the Table object because we want to group by error-level and rule name.

Any non-CSV format will print tab-separated values when using the `--print` option (I don't think it makes sense to print the first N violations of an HTML doc, for example...). For formats HTML, XLSX, JSON, and YAML, an `--output` is required. For TSV and CSV, the output is optional  - if not included, all output will be printed to the console.

This also adds a `--standalone` option to both `report` and `export` when outputting HTML.


beckyjackson:
This could cause some conflicts with #691 - not sure what the best order to merge is.

jamesaoverton:
What's going on with `Violations.java`?

beckyjackson:
I changed to using OWL objects instead of just string representations of the IRIs so that we can render them with our various renderers we use for `export` (like the HTML renderer). Otherwise I'd have to write another method that takes a string IRI, maybe finds a label or makes a CURIE, and turns this into a link.

beckyjackson:
I'd like to wait for #691 before merging this so I can work out any merge conflicts.

robot-core/src/main/java/org/obolibrary/robot/checks/Violation.java
```diff
@@ -1,29 +1,69 @@
...
+      }
     }
     statements.put(property, values);
   }
+
+  @Deprecated public String subject;
+
+  /** @param subject String subject */
+  @Deprecated
+  public Violation(String subject) {
```
jamesaoverton, line 55: We published these public methods, so I'd very much like them to continue to work somehow.

beckyjackson: Makes sense. I've added these back in and added support for these types of violations when printing reports out.

beckyjackson:
The bug that caused the background colors to not display properly in HTML reports is fixed and this is ready to go.

PR #701, opened 2020-05-28
WIP: OBO-Extract method

beckyjackson:
Some things to think about:
- Do we want to include equivalent and disjoint axioms? These are currently included, but only if all classes are in the targets (e.g. between upper and lower)
- Do we want to try and parse contents of `owl-axioms` tags?

I've tested this on some smaller files, but I think this probably needs a bit of testing on various OBO files. I'd like to get some other eyes on it as well, since I'm not super familiar with OBO format. I based a lot of this on the [OBO Flat File Format Guide](https://owlcollab.github.io/oboformat/doc/GO.format.obo-1_2.html).

This is a proposed replacement for the RDFXML extract method #650. It still works in two passes - first to get the hierarchy structure and all the annotation properties, then to add the annotations (and logic, if we decide to keep that).

## OBO Extract

The `OBO-Extract` method is based on MIREOT principles, but streams the contents of an OBO-format file to reduce strain on memory. This is much faster for large files, but only works for files in OBO format (`.obo` or `.obo.gz`). These files must follow the [OBO Flat File Format Guide](https://owlcollab.github.io/oboformat/doc/GO.format.obo-1_2.html).

Like MIREOT, this method uses "upper" and "lower" terms. The "lower" term or terms are required, whereas "upper" terms are optional.

    robot extract --method OBO-Extract \
        --input example.obo \
        --upper-term UBERON:0000465 \
        --lower-term UBERON:0002369 \
        --output obo_extract.owl
        
Unlike MIREOT, equivalence axiom will be included, but *only* if all classes used in the axiom are between the target terms.

For this method, one or more `--annotation-property` options can be provided in order to add annotations to the entities. If not included, all annotations will be included in the output. To provide a text file of annotation properties, use `--annotation-properties`. The `--intermediates` may also be used with this method.

    robot extract --method OBO-Extract \
        --input example.obo \
        --upper-term UBERON:0000465 \
        --lower-term UBERON:0002369 \
        --intermediates none \
        --annotation-property rdfs:label \
        --output obo_extract_small.owl

### Disclaimers

Please note that `[Instance]` tags are not currently supported by OWLAPI's `OBOFormatParser` so they will be lost if you convert an OWL file to OBO, but the `OBO-Extract` method *will* capture instances.

Additionally, anonymous subclass expressions are stored in an `owl-axioms` field in the ontology header. This field is not parsed by the `OBO-Extract` method. This may be updated in future releases.


cmungall:
Thanks!

I would use the 1.4 specification, it is canonical, and hopefully more precise than the 1.2 guide:
https://owlcollab.github.io/oboformat/doc/obo-syntax.html

beckyjackson:
Thanks Chris! I was originally using 1.4, but I saw it was marked as a draft (I guess I had the wrong link). I'll update the link.

jamesaoverton:
Like #650, I'm no longer convinced that this is a good idea, so I'm closing this PR.

PR #703, opened 2020-06-06
Add HTML lists to export

### LINKED ISSUE #633: Export HTML lists
This is one part of #629. Current HTML export is as a table, but I would also like to support 1`--format html-list`. They'll look a bit like #631:

```
<ul>
  <li><em>CURIE</em> <a href="http://purl.obolibrary.org/obo/GO_0005622">GO:0005622</a></li>
  <li><em>LABEL</em> <span>intracellular</span></li>
  <li><em>SubClass Of</em> <span><a href="http://purl.obolibrary.org/obo/GO_0044464">cell part</a></span></li>
</ul>
<ul>
  <li><em>CURIE</em> <a href="http://purl.obolibrary.org/obo/GO_ 044424">GO:0044424 </a></li>
  <li><em>LABEL</em> <span>intracellular part</span></li>
  <li><em>SubClass Of</em> <span><a href="http://purl.obolibrary.org/obo/GO_0044464">cell part</a></span> | <span><a href="http://purl.obolibrary.org/obo/BFO_0000050">part of</a> some <a href="http://purl.obolibrary.org/obo/GO_0005622">intracellular</a></span></li>
</ul>
```

Maybe the values should be a sublist (styled as a Boostrap inline list)?

beckyjackson:
Resolves #633

This should probably wait on #691 and #699 before merging

Adds `html-list` format to `export` to output reports in a simple HTML list:

image

I tried using the Bootstrap inline lists, but it looked kind of weird. This will also support the `--standalone` option that we are adding in #691 - `--standalone false` will not include the `<head>` and `<body>` tags.

Also fixes a minor bug with the HTML renderer in which `&` and `<` symbols in text would cause issues with the HTML. These are now replaced by `&amp;` and `&lt;`.


robot-core/src/main/java/org/obolibrary/robot/ManchesterOWLSyntaxObjectHTMLRenderer.java
```diff
@@ -92,7 +92,8 @@ public void visit(OWLNamedIndividual individual) {
     write(
         String.format(
             "<a href=\"%s\">%s</a>",
-            individual.getIRI().toString(), getShortFormProvider().getShortForm(individual)));
+            individual.getIRI().toString(),
+            escape(getShortFormProvider().getShortForm(individual))));
```
jamesaoverton, line 96: Why is this one not escaped, when all the others are?

robot-core/src/main/java/org/obolibrary/robot/export/Row.java
```diff
@@ -275,6 +275,42 @@ public String toHTML(List<Column> columns, String split) {
...
+    for (Column c : columns) {
+      String columnName = c.getDisplayName();
+      Cell cell = cells.getOrDefault(columnName, null);
+      if (cell != null) {
+        // Only add to HTML list if not null and values is not empty
+        List<String> values = cell.getDisplayValues();
+        if (values.size() == 0) {
+          continue;
+        }
+        sb.append("\t<li>\n");
```
jamesaoverton, line 294: I'm not sure why we're using tabs for the HTML output, when we don't use tabs anywhere else in any of our projects (TSV doesn't count).

robot-core/src/main/java/org/obolibrary/robot/export/Table.java
```diff
@@ -332,6 +332,20 @@ public String toHTML(String split, boolean standalone, boolean includeJS) {
     return sb.toString();
   }
 
+  public String toHTMLList() {
+    StringBuilder sb = new StringBuilder();
+    sb.append("<head>\n")
+        .append(
+            "\t<link rel=\"stylesheet\" href=\"https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css\">\n")
```
jamesaoverton, line 339: I'd like all these versioned links factored out to shared variables. This Bootstrap version doesn't match others in this same file.

jamesaoverton (changes requested):
Sorry for the long delay.

Overall this looks good but I'd like some changes.

robot-core/src/main/java/org/obolibrary/robot/ManchesterOWLSyntaxObjectHTMLRenderer.java
beckyjackson: I think it is escaped? ` escape(getShortFormProvider().getShortForm(individual))`

robot-core/src/main/java/org/obolibrary/robot/export/Row.java
beckyjackson: I guess to make the HTML more legible. Should I take it out?

robot-core/src/main/java/org/obolibrary/robot/ManchesterOWLSyntaxObjectHTMLRenderer.java
jamesaoverton: Ok

robot-core/src/main/java/org/obolibrary/robot/export/Row.java
jamesaoverton: I would like the code to use spaces for indentation, please. Two spaces for each level works well for HTML, which can get quite deep.

jamesaoverton (approved):
Thanks!

PR #710, opened 2020-07-22
Fix for overlapping classes/individuals when running MIREOT extract

### LINKED ISSUE #709: robot doesn't get pH from current UO by MIREOT
the pH unit is http://purl.obolibrary.org/obo/UO_0000196

It appears twice in the hierarchy, with two different parent classes, and it has also been punned to an individual. 

```Bash
 markampa$ grep http://purl.obolibrary.org/obo/UO_0000196 ontology_downloads/uo.owl 
    <!-- http://purl.obolibrary.org/obo/UO_0000196 -->
    <owl:NamedIndividual rdf:about="http://purl.obolibrary.org/obo/UO_0000196">
                    <rdf:Description rdf:about="http://purl.obolibrary.org/obo/UO_0000196"/>
    <!-- http://purl.obolibrary.org/obo/UO_0000196 -->
    <owl:NamedIndividual rdf:about="http://purl.obolibrary.org/obo/UO_0000196"/>
    <rdf:Description rdf:about="http://purl.obolibrary.org/obo/UO_0000196">
```

But ROBOT doesn't seem to find it with MIREOT:

```Bash
markampa$ robot extract --verbose --method MIREOT --input ./ontology_downloads/uo.owl \
  --lower-term http://purl.obolibrary.org/obo/UO_0000196 --output pH.ttl

markampa$ cat pH.ttl 
@prefix : <http://purl.obolibrary.org/obo/uo.owl#> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix xml: <http://www.w3.org/XML/1998/namespace> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@base <http://purl.obolibrary.org/obo/uo.owl> .

<http://purl.obolibrary.org/obo/uo.owl> rdf:type owl:Ontology .

###  Generated by the OWL API (version 4.5.6) https://github.com/owlcs/owlapi
```

beckyjackson:
Resolves #709

When an entity is both a class and an individual, ignore the individual and prefer the class. This required a small hack because if you just call `OntologyHelper.getEntitiy(ontology, iri)` it will not return an entity that occurs as both.


beckyjackson:
Do we want to add a test for this? It works in the example from #709 but that seems like a pretty rare instance (hopefully...).

jamesaoverton:
Because it's rare, an automated test is even more important. :smile:

beckyjackson:
Makes sense @jamesaoverton - I've added a test in for this

beckyjackson:
Conflicts resolved, is this ready to merge @jamesaoverton ?

PR #713, opened 2020-07-28
Add standardized error output to `template`

### LINKED ISSUE #712: Standardized output for lists of problems with an input table
The ROBOT `template` and `validate` commands accept tables as input. When there are problems with those input tables, we log errors to the console, but it can be a pain to mentally map the console lines back to the input table.

The `validate` can also generate HTML and Excel tables with the problem cells highlighted and comments describing the problem. This is much more friendly.

ROBOT `report` outputs lists of problems in a standardized tabular format that I like, but it reports problems with triples not cells of a table.

One things that's missing from all three is an automatically suggested fix for the problem, as discussed in #705.

I'd like an option to output the problems from `template` and `validate` as a table, similar to the tabular output of `report`. This would be an intermediate format, that could then be converted into a highlighted table.

I think the columns should be:

1. problem: a local numeric identifier, start at 1 and count up; this is mainly to help people refer to the problem
2. table: file name (for TSV) or sheet name (for Excel)
3. cell: I'd prefer "A1" format
4. level: ERROR, WARN, INFO, DEBUG
5. rule ID: preferably an IRI or CURIE
6. rule name: preferably something descriptive
7. value: usually the value of the cell
8. fix: (optional) preferably a single suggestion, but may be blank or a pipe-separated ordered list of suggestions

In an HTML version of this table, rule ID and rule name would be combined into a link. Since we don't implement suggestions yet, this column can be blank. But I'd like to settle on a set of columns that we can standardize and reuse for a bunch of validation tools, even one-off Python scripts.

beckyjackson:
Resolves #712

This adds an `--errors <path>` option to `template` to write a table of all errors (as specified in #712) to the provided path (CSV or TSV).


beckyjackson:
I'd like to wait on #691 before merging this so I'm marking it as a WIP.

beckyjackson:
The `template` classes now have a ton of deprecated methods from when we changed the way it works. If we remove these, `make validate` will fail because of removed methods, but I'm wondering if we should consider allowing that since they've been deprecated for quite some time now? @jamesaoverton

Issue #716, opened 2020-07-31
Impossible to create properties with a characteristic set to ''


bjonnh:
**Robot version ** 1.7.0

** Problem ** When trying to create properties from a template, I get a index out of bound error
It seems to work without the CHARACTERISTIC column. When playing around I had another error message that said that the characteristic '' was not valid, but I couldn't find how I got that one.

It works if they are both set to transitive, so it seems that the empty characteristic should be handled?

** Source file **
```
id,type,Label,comment,characteristic
ID,TYPE,A rdfs:label,A rdfs:comment,CHARACTERISTIC SPLIT=|
nav:x,owl:ObjectProperty,x,x,
nav:y,owl:ObjectProperty,y,y,transitive
```

**Output with -vvv ** 

````java -jar ../local/robot.jar template -vvv --template tables/current/properties.csv \
        --ontology-iri "https://onto.nprod.net/nav.owl" \
        --prefix "ero:http://purl.obolibrary.org/obo/ERO_" --prefix "nav:https://onto.nprod.net/NAV_" \
        --output tables/current/properties.owl
2020-07-31 17:24:33,903 DEBUG org.semanticweb.owlapi.utilities.Injector - Loading file META-INF/services/org.semanticweb.owlapi.model.OWLOntologyManager
2020-07-31 17:24:33,905 DEBUG org.semanticweb.owlapi.utilities.Injector - Loading URL for service jar:file:XXX/2020-newnavigator/local/robot.jar!/META-INF/services/org.semanticweb.owlapi.model.OWLOntologyManager
2020-07-31 17:24:33,906 DEBUG org.semanticweb.owlapi.utilities.Injector - Loading URL for service jar:file:XXX/2020-newnavigator/local/robot.jar!/META-INF/services/org.semanticweb.owlapi.model.OWLOntologyManager
2020-07-31 17:24:33,906 DEBUG org.semanticweb.owlapi.utilities.Injector - Loading URL for service jar:file:XXX/2020-newnavigator/local/robot.jar!/META-INF/services/org.semanticweb.owlapi.model.OWLOntologyManager
2020-07-31 17:24:33,943 DEBUG org.semanticweb.owlapi.utilities.Injector - Loading file META-INF/services/org.semanticweb.owlapi.model.OWLDataFactory
[...]

PR #717, opened 2020-07-31
Solve issue #716: empty characteristics for properties

bjonnh:
Resolves [#716 ]

- [NOT APPLICABLE] `docs/` have been added/updated
- [NO] tests have been added/updated : Not done, there doesn't seem to be much tested for templates yet, will revisit if necessary
- [PROBLEM] `mvn verify` says all tests pass : but was there before this modification

````
  [ERROR] Number of foreign imports: 1
  [ERROR] import: Entry[import  from realm ClassRealm[maven.api, parent: null]]
  [ERROR] 
  [ERROR] ------------
````

- [PROBLEM] `mvn site` says all JavaDocs correct  : Failure missing javadoc.

This handle the two cases raised by the issue #716 :
- we need to allow empty characteristics
- in the case of characteristics being the last	column,
  the list of elements in the row is one element short
  when the characteristic is empty, handling that case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Now handles empty characteristics [#716]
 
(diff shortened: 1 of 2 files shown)


beckyjackson:
Thanks for the fix! I replaced this with a branch on this repo so that it builds in Jenkins as well. I'm going to close this in favor of that one - #719

PR #722, opened 2020-08-17
Fix referencing properties by CURIE in export

### LINKED ISSUE #721: prefixes not working with robot export
I tried running this on go.owl:

`robot export -i go.owl --header "ID|LABEL|IAO:0000115|oboInOwl:hasOBONamespace" --add-prefix "IAO: http://purl.obolibrary.org/obo/IAO_" --add-prefix "oboInOwl: http://www.geneontology.org/formats/oboInOwl#" --export table.tsv`

I don't get any output for `IAO:0000115` or `oboInOwl:hasOBONamespace`. If I use the labels for those properties, I do. Am I doing something wrong?

beckyjackson:
Resolves #721

Change to compare IRI with `equals()` instead of `==` in `export` in order to allow both CURIEs and labels in `export` headers. I have no idea why two properties with the same IRI would be considered different IRI objects when we pass CURIEs but not labels, but if we change to this comparison, it's resolved.

Previously:

```
prop1.getIRI() == prop2.getIRI()
```

Now:

```
prop1.getIRI().equals(prop2.getIRI())
```

I also made the process to get the IRI of a header a little bit smarter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Handle empty [`template`] property charactersitics in [#719]
+- Fix referencing properties by CURIE in [`export`] in [#722]
 
 ## [1.7.0] - 2020-07-31
 
@@ -198,6 +199,7 @@ First official release of ROBOT!
 [`template`]: http://robot.obolibrary.org/template
 [`validate`]: http://robot.obolibrary.org/validate
 
+[#722]: https://github.com/ontodev/robot/pull/722
 [#719]: https://github.com/ontodev/robot/pull/716
 [#715]: https://github.com/ontodev/robot/pull/715
 [#710]: https://github.com/ontodev/robot/pull/710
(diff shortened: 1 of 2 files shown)


balhoff:
@beckyjackson how about `prop1.getIRI().equals(prop2.getIRI())`? `.equals` vs. `==`

beckyjackson:
I should have thought to try that first. It seems to work just as well.

It's strange that `==` works for by-label but not by-CURIE, but `equals()` works for all so... I won't question it.

balhoff:
> It's strange that == works for by-label but not by-CURIE, but equals() works for all so... I won't question it.

😄 strings may be reused internally I think (interned). Also OWL API uses a cache for the strings in IRI construction: https://github.com/owlcs/owlapi/blob/23593593c2853006bfe3bb61ec03fae2e3736461/api/src/main/java/org/semanticweb/owlapi/model/IRI.java#L63

I would just stick to `.equals` 99.999% of the time in Java. One of the little benefits of Scala is that the've swapped this and use `==` as value equality, and `eq` for object identity.

PR #730, opened 2020-09-09
Add Whelk reasoner.

balhoff:
Add Whelk as a reasoner choice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Add Whelk OWL reasoner in [#730]
 - Add [`validate`] command [#691]
 - Add `--errors <path>` option to [`template`] in [#713]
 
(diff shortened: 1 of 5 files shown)


jamesaoverton:
My only concern is increasing the size of the binary by ~15%:

branch | link | size
-- | -- | --
master | https://build.obolibrary.io/job/ontodev/job/robot/job/master/ | 62.36 MB
whelk | https://build.obolibrary.io/job/ontodev/job/robot/job/whelk/ | 72.39 MB

What are the key benefits of Whelk for ROBOT?

balhoff:
> My only concern is increasing the size of the binary by ~15%

This was surprising to me, since Whelk is tiny and I have tried to keep the dependencies minimal. The core has only one dependency (scalaz-core) which is used for a couple small things; it turns out it's a lot bigger than I realized! scalaz-core accounts for pretty much all of that. I'll just re-implement the one method I'm using—hold off until I update this PR.

> What are the key benefits of Whelk for ROBOT?

Whelk provides a combination of OWL features not available together in any of the other reasoners:
- EL scalability (albeit slower than ELK)
- better support for Self restrictions than ELK, supporting rolification
- object property assertion materialization
- OWL RL features for Abox
- SWRL rules (class and object property atoms)

Whelk is also much faster than ELK at answering DL queries, but I don't think this will be apparent in the current ROBOT commands.

jamesaoverton:
Thanks. Now I see 63.00 MB.

The changes look good. I want to try it out on a couple of things before merging.

PR #761, opened 2020-11-04
Suggestion to allow single quotes in C template values

beckyjackson:

Currently, when a term label contains a single quote, we need to use CURIEs instead of labels in the parent column, or else the term cannot be parsed.

For example, the label:
> Streptococcus sp. 'group B'

Results in:
```
MANCHESTER PARSE ERROR the expression ''Streptococcus sp. 'group B'''  at row 1976, column 2 
in table "src/ontology/templates/taxon.tsv" cannot be parsed: encountered unknown 'Streptococcus sp. '
```

Typically, I would just use `NCBITaxon:1319` instead and the template would be OK, but now that we are using labels with [VALVE](https://github.com/ontodev/valve.py), we need to be able to reference all labels in the "Parent" columns or else certain functions fail.

This is a proposed fix that first attempts to use a `QuotedEntityChecker` to retrieve a class by label. If a class cannot be found, then the `ManchesterOWLSyntaxClassExpressionParser` is used to try to parse an expression. My one concern is that for large templates with a bunch of class expressions, this double-step may be less efficient. At the same time, it may actually be quicker in some cases where only labels are used. The relevant lines are here:
https://github.com/ontodev/robot/blob/e486ce9c009334be9b21a1c8a6c212ecba8dbb6c/robot-core/src/main/java/org/obolibrary/robot/TemplateHelper.java#L987-L996

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/DiffOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/DiffOperation.java
@@ -153,9 +153,7 @@ public class DiffOperation {
     return false;
   }
 
-  /**
-   * OWLOntologySetProvider for two (left and right) ontologies.
-   */
+  /** OWLOntologySetProvider for two (left and right) ontologies. */
   private static class DualOntologySetProvider implements OWLOntologySetProvider {
 
     private static final long serialVersionUID = -8942374248162307075L;
--- a/robot-core/src/main/java/org/obolibrary/robot/Template.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/Template.java
@@ -870,34 +870,40 @@ public class Template {
         // Subclass expression
         subclassExpressionColumns.put(
             column,
-            TemplateHelper.getClassExpressions(name, parser, template, value, rowNum, column));
+            TemplateHelper.getClassExpressions(
+                name, checker, parser, template, value, rowNum, column));
       } else if (template.startsWith("EC")) {
         // Equivalent expression
         equivalentExpressionColumns.put(
             column,
-            TemplateHelper.getClassExpressions(name, parser, template, value, rowNum, column));
+            TemplateHelper.getClassExpressions(
+                name, checker, parser, template, value, rowNum, column));
       } else if (template.startsWith("DC")) {
         // Disjoint expression
         disjointExpressionColumns.put(
             column,
-            TemplateHelper.getClassExpressions(name, parser, template, value, rowNum, column));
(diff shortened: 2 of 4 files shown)


beckyjackson:
We still run into problems when using a label with a single quote in expressions...

```
MANCHESTER PARSE ERROR the expression ''in taxon' some 'Theiler's encephalomyelitis virus'' at row 494,
column 6 in table "src/ontology/templates/protein.tsv" cannot be parsed: encountered unknown 'Theiler'
```

I can work around this in VALVE since it's not a tree, but it's still annoying

jamesaoverton:
I guess ROBOT should be escaping before insertion: `Theiler\'s encephalomyelitis virus`.

These changes have a fair chance of to breaking somebody's stuff, somewhere. I'll want to unit test them thoroughly, then test against a bunch of real ontologies. If we see significant breakage, then I'll want to hide this improved string handling behind an option.

beckyjackson:
But you can plug in expressions into templates, e.g. I can provide this string as a value in a cell (not the substituted template string):
> part_of some 'test one'

If we escape single quotes before passing it to the parser, this parsing will fail (the unit tests fail):
```
testTemplateCSV(org.obolibrary.robot.TemplateTest): template#MANCHESTER PARSE ERROR
the expression 'part_of some 'test one'' at row 3, column 8 in table "/template.csv" cannot be parsed:
encountered unknown 'test
```

I don't think there's a good way to know if something is an expression vs. a label, unless we run it through the `QuotedEntityChecker`. Thoughts?

jamesaoverton:
Yes, using the QuotedEntityChecker is the right way to determine whether a cell contains a label. I didn't think I was disagreeing, just mentioning one case. I guess we need a list of cases.

beckyjackson:
Oh I misunderstood, sorry!

jamesaoverton (changes requested):
I'd like to move forward with this, but I need to be convinced that it won't break things. @beckyjackson Please try this on OBI, MRO, ONTIE, and maybe other projects and report on the results.

beckyjackson:
I think ONTIE is the only one I can think of that uses labels with single quotes in template patterns. I tested it on ONTIE (using the labels instead of the CURIEs) and it worked exactly as expected.

I also built MRO and OBI using the JAR from Jenkins and it completed just fine!

jamesaoverton:
Ok. Please add a few unit tests, and then we'll merge this.

beckyjackson:
Test added!

jamesaoverton (approved):

PR #773, opened 2020-11-24
Make Table more widely useable

beckyjackson:
I made a couple of changes to `Column` and `Table` that should help users use these classes from the library more easily. I also wrote a doc page on how to use the `Table` to create various formats of table outputs.

I'm not sure where to put this doc, though, so I just created a new `docs/developer/` folder. Thoughts?


jamesaoverton:
@matentzn Can you please see if these docs suit your current needs?

@beckyjackson I'm fine with `docs/developers/`. The only other place I can think of would be `CONTRUBTING.md` but that doesn't seem like a good fit.

beckyjackson:
@matentzn said these changes worked for him, I'm marking this ready for review!

matentzn:
I think the docs and changes all make sense to me.. I need this change for my branch, so I would be happy for it to be merged! :)

jamesaoverton (approved):

PR #774, opened 2020-11-24
Add metrics command first pass

matentzn:
Resolves [#627]

- [x] `docs/` have been added/updated
- [x] tests have been added/updated
- [x] `mvn verify` says all tests pass
- [x] `mvn site` says all JavaDocs correct
- [x] `CHANGELOG.md` has been updated

ROBOT jar for testing [here](https://www.dropbox.com/s/3wr8lcuz3pkopoo/robot.jar?dl=0).

For the community, the most relevant question will be which of the metrics should go in which bin (essential, extended, advanced):

#### Entity counts
- SIGNATURE_SIZE_INCL
- SIGNATURE_SIZE
- CLASS_COUNT_INCL
- CLASS_COUNT
- OBJPROPERTY_COUNT_INCL
- OBJPROPERTY_COUNT
- INDIVIDUAL_COUNT_INCL
- INDIVIDUAL_COUNT
- DATAPROPERTY_COUNT_INCL
- DATAPROPERTY_COUNT
- ANNOTATION_PROP_COUNT
- ANNOTATION_PROP_COUNT_INCL
- DATATYPE_COUNT_INCL
- DATATYPE_COUNT

#### Axiom counts
- LOGICAL_AXIOM_COUNT
- LOGICAL_AXIOM_COUNT_INCL
- AXIOM_COUNT
- AXIOM_COUNT_INCL
- TBOX_SIZE
- TBOX_SIZE_INCL
- TBOXRBOX_SIZE
- TBOXRBOX_SIZE_INCL
- RULE_CT
- RULE_CT_INCL
- RBOX_SIZE
- RBOX_SIZE_INCL
- ABOX_SIZE
- ABOX_SIZE_INCL

#### OWL Profile metrics
- BOOL_PROFILE_OWL
- BOOL_PROFILE_OWL
- BOOL_PROFILE_OWL
- BOOL_PROFILE_OWL
- BOOL_PROFILE_OWL
- BOOL_PROFILE_RDFS
- VIOLATION_PROFILE_OWL

#### Other metrics
- OWLAPI_VERSION
- ONTOLOGY_ID
- ONTOLOGY_ANNOTATIONS_COUNT
- VALID_IMPORTS
- VALID_IMPORTS_INCL

#### Extended entity metrics
- DATATYPE_BUILTIN_COUNT_INCL
- DATATYPE_BUILTIN_COUNT
- DATATYPE_NOTBUILTIN_COUNT_INCL
- DATATYPE_NOTBUILTIN_COUNT

[...]

Issue #627, opened 2020-02-21
New command: statistics/metrics (or add option to export/report)


cmungall:
Was mentioned in a comment in #205 (now closed)

It would be useful to be able to generate aggregate stats in a standardized form in robot

This could include the axiom counts from as seen in Protege

![image](https://user-images.githubusercontent.com/50745/75079177-bbd36680-54bc-11ea-815a-0076f4f6baf0.png)

This would be useful for ontogeeks. But for others it would also be useful to see more basic stuff like

 - num classes/relations
    - incl/excl obsolete
    - all / main ID space only
 - num test definitions/synonyms/xrefs/other APs
 - num of existential restrictions broken down by OP
 - num of defining equivalence axioms (NamedC to AnonC)

A simple TSV would be best for most people, we could also consider an rdf export using vocabs like void/dcat

This could be a new command like `statistics` (Protege says "metrics" but this sounds like OBO evaluation, whereas these are just for informative purposes)

Or it could be piggy backed as an option on `report`. This is really more like a "report card" so not clear if aggregate stats figure here, and what category (INFO?) we would use.

jamesaoverton:
Good idea. I prefer a new command (I like `statistics`), since this would have a different implementation and goals from `report`. But I could be convinced otherwise.

I don't have a pressing need for this, though. It would probably be a good feature for a new ROBOT contributor. Protege's implementation would be a starting point.

cmungall:
Another option here is to add an option `--stats FILE` to `export` that will produce summary statistics and write these to FILE

The stats can be very basic. One line per header, listing total number of entries and number of distinct entries. E.g. if I specify `ID|label|synonym`

E.g.

```
ID count: 13000 13000
label count: 13000 13000
synonym count: 20000 17000
```

Anything more advances should be done using R/pandas/etc

I like this because it standardizes some of the summary stats we see published with ontologies. Sometimes I will see a table in a paper with headings like 'num relationships' or  'num terms' with ambiguous meaning. We have already thrashed out precise semantics of the column headings within export, including what should be done with anons etc.

matentzn:
I would like to take this on! I would prefer `metrics` over `statistics`, because its broader, and I could include things like "owl profile", "serialisation format", which are not statistics.. I was thinking of something simple like this:

```
robot metrics -i ont.owl $(OPTS) -o metrics.yml
```

Which produces a file like this:

```
metrics:
  - metric: term_count
     metric_type: count
     value: 566 
[...]

PR #778, opened 2020-11-26
New command: measure

matentzn:
Resolves [#627]
The original pull request was here: https://github.com/ontodev/robot/pull/774

I replaced it to trigger jenkins etc.

[robot.jar](https://build.obolibrary.io/job/ontodev/job/robot/job/cmd-metrics/lastSuccessfulBuild/artifact/bin/robot.jar) for testing.

For the community, the most relevant question will be which of the metrics should go in which bin (essential, extended, advanced):

#### Entity counts
- SIGNATURE_SIZE_INCL
- SIGNATURE_SIZE
- CLASS_COUNT_INCL
- CLASS_COUNT
- OBJPROPERTY_COUNT_INCL
- OBJPROPERTY_COUNT
- INDIVIDUAL_COUNT_INCL
- INDIVIDUAL_COUNT
- DATAPROPERTY_COUNT_INCL
- DATAPROPERTY_COUNT
- ANNOTATION_PROP_COUNT
- ANNOTATION_PROP_COUNT_INCL
- DATATYPE_COUNT_INCL
- DATATYPE_COUNT

#### Axiom counts
- LOGICAL_AXIOM_COUNT
- LOGICAL_AXIOM_COUNT_INCL
- AXIOM_COUNT
- AXIOM_COUNT_INCL
- TBOX_SIZE
- TBOX_SIZE_INCL
- TBOXRBOX_SIZE
- TBOXRBOX_SIZE_INCL
- RULE_CT
- RULE_CT_INCL
- RBOX_SIZE
- RBOX_SIZE_INCL
- ABOX_SIZE
- ABOX_SIZE_INCL

#### OWL Profile metrics
- BOOL_PROFILE_OWL
- BOOL_PROFILE_OWL
- BOOL_PROFILE_OWL
- BOOL_PROFILE_OWL
- BOOL_PROFILE_OWL
- BOOL_PROFILE_RDFS
- VIOLATION_PROFILE_OWL

#### Other metrics
- OWLAPI_VERSION
- ONTOLOGY_ID
- ONTOLOGY_ANNOTATIONS_COUNT
- VALID_IMPORTS
- VALID_IMPORTS_INCL

#### Extended entity metrics
- DATATYPE_BUILTIN_COUNT_INCL
- DATATYPE_BUILTIN_COUNT
- DATATYPE_NOTBUILTIN_COUNT_INCL
- DATATYPE_NOTBUILTIN_COUNT

#### Logical expressivity
- EXPRESSIVITY
- EXPRESSIVITY_INCL
- CONSTRUCTS_INCL
- CONSTRUCTS
- AXIOM_TYPES
- AXIOM_TYPES_INCL
- AXIOMTYPE_COUNT
- AXIOMTYPE_COUNT_INCL
- SYNTAX

#### Entity usage
- NS_USE_AXIOMS
- NS_USE_AXIOMS_INCL
- NS_USE_SIGNATURE
- NS_USE_SIGNATURE_INCL

### Advanced metrics (all)
- MAX_AXIOMLENGTH
- MAX_AXIOMLENGTH_INCL
- AVG_ASSERT_N_SUBCLASS_INCL
- AVG_ASSERT_N_SUBCLASS
- AVG_ASSERT_N_SUPERCLASS_INCL
- AVG_ASSERT_N_SUPERCLASS
- AVG_INSTANCE_PER_CLASS_INCL
- AVG_INSTANCE_PER_CLASS
- CLASS_SGL_SUBCLASS_COUNT_INCL
- CLASS_SGL_SUBCLASS_COUNT
- AXIOM_COMPLEXRHS_COUNT_INCL
- AXIOM_COMPLEXRHS_COUNT
- TBOX_CONTAINS_NOMINALS_INCL
- TBOX_CONTAINS_NOMINALS
- ABOX_CONTAINS_NOMINALS_INCL
- ABOX_CONTAINS_NOMINALS
- GCI_COUNT_INCL
- GCI_COUNT
- GCI_HIDDEN_COUNT_INCL
- GCI_HIDDEN_COUNT
- MAX_NUM_NAMED_SUPERCLASS_INCL
- MAX_NUM_NAMED_SUPERCLASS
- MULTI_INHERITANCE_COUNT_INCL
- MULTI_INHERITANCE_COUNT
- REF_CLASS_COUNT
- REF_DATAPROP_COUNT_INCL
- REF_DATAPROP_COUNT
- REF_INDIV_COUNT_INCL
- REF_INDIV_COUNT
- REF_OBJPROP_COUNT_INCL
- REF_OBJPROP_COUNT
- UNDECLARED_ENTITY_COUNT_INCL
- UNDECLARED_ENTITY_COUNT
- TAUTOLOGYCOUNT
- TAUTOLOGYCOUNT_INCL
- CYCLE
- CYCLE_INCL
- CYCLE
- CYCLE
- DATATYPE_AXIOMCOUNT
- DATATYPE_AXIOMCOUNT_INCL
- DATATYPES
- DATATYPES_INCL
- DATATYPES_NOT_BUILT_IN
- DATATYPES_NOT_BUILT_IN_INCL
- MOST_FRQUENTLY_USED_CONCEPT_INCL
- MOST_FRQUENTLY_USED_CONCEPT


matentzn:
I have re-created this pull request in a branch local to robot so that the jenkins jobs etc are run properly. Original PR with some interesting comments here:
https://github.com/ontodev/robot/pull/774

matentzn:
I walked @cmungall through this PR and he is happy. he suggests the write a biolink ML schema for the metrics, which I will do, just because I like it. This will allow us to link specific metrics to documentation.

robot-core/src/main/java/org/obolibrary/robot/MeasureOperation.java
```diff
@@ -0,0 +1,330 @@
...
+      throws IOException {
+    if (!result.isEmpty()) {
+      writeResult(result, format, output);
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  public static void executeMetrics(
```
beckyjackson, line 62: I think this should be renamed to `measure` to match other core operations (e.g. `MaterializeOperation.materialize` and `MergeOperation.merge`). Also, please add Java docs to this method.

robot-core/src/main/java/org/obolibrary/robot/MeasureOperation.java
```diff
@@ -0,0 +1,330 @@
...
+      OWLOntology o,
```
beckyjackson, line 63: This is super minor and I'm not sure if it's important, but the `OWLOntology` is called `ontology` everywhere else (`OWLOntology ontology`).

robot-core/src/main/java/org/obolibrary/robot/MeasureOperation.java
```diff
@@ -0,0 +1,330 @@
...
+      OWLReasonerFactory rf,
+      String metrics_type,
```
beckyjackson, line 65: Please use camel case for variables - I think `metric_type` and `metric_value` are the only vars like this.

robot-core/src/main/java/org/obolibrary/robot/MeasureOperation.java
```diff
@@ -0,0 +1,330 @@
...
+
+  /**
+   * Given a dataset, a query string, a format name, and an output stream, run the SPARQL query over
+   * the named graphs and write the output to the stream.
+   *
+   * @param ontology Ontology to run metrics
+   * @param metrics_type what kind of metrics to harvest
+   * @return Metrics, if successful
+   */
+  public static MeasureResult runMeasures(
```
beckyjackson, line 91: I think `getMetrics` would be a better name here? Since we're returning the metrics in this step?

robot-core/src/main/java/org/obolibrary/robot/MeasureOperation.java
```diff
@@ -0,0 +1,330 @@
...
+    return metrics;
+  }
+
+  /**
+   * Write a model to an output stream.
+   *
+   * @param result results of the metrics operation
+   * @param format the language to write in (if null, TTL)
+   * @param output the output stream to write to
+   */
```
beckyjackson, line 157: Java docs need a `@throws IOException` line

robot-core/src/main/java/org/obolibrary/robot/MeasureOperation.java
```diff
@@ -0,0 +1,330 @@
...
+    ObjectMapper mapper =
+        new ObjectMapper(new YAMLFactory().disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER));
+    return mapper.writeValueAsString(jsonNodeTree);
+  }
+
+  private static String escapeTSV(String s) {
+    return s.replaceAll("\t", " ");
+  }
+
+  private static final Column cl_metric = new Column("metric");
```
beckyjackson, line 261: Final vars should be all uppercase (underscores OK here), e.g. `CL_METRIC`

robot-core/src/main/java/org/obolibrary/robot/metrics/MeasureResult.java
```diff
@@ -0,0 +1,56 @@
+package org.obolibrary.robot.metrics;
+
+import java.util.*;
+
+public class MeasureResult {
+
+  Map<String, Object> data = new HashMap<>();
+  Map<String, List<Object>> data_listvalues = new HashMap<>();
```
beckyjackson, line 8: Please use camel case for non-final variables (e.g., `dataListValues`)

beckyjackson (changes requested):
The code looks good! I ran it a couple times and the results look as expected. Super cool! I love the namespace metrics.

My only comments are about code style, so they're pretty picky (I made a couple of comments on examples of these, but I stopped after the first one or two):
- [ ] Use camel case for non-final variables
- [ ] Final variables should be all uppercase
- [ ] All `public` methods and classes need Javadoc strings attached to them (e.g, in `MeasureResult`, `MetricsUtils`, etc.)
- [x] The tests look thorough, but it looks like there's no tests for the `-reasoner` metrics. Can you please add that?

matentzn:
Merci bien @beckyjackson will take care of all of that! Very much appreciate the effort!

beckyjackson:
This isn't something super important, but have you considered adding support for a custom set of metrics? For example, what if I only want `axiom_count` and `namespace_entity_count`? Perhaps we could supply a string list of the metrics we want to `-m`, or make it a different option... I'm just thinking of this because we have some custom queries for DO right now that generate similar data, but there's a lot of data in the `measure` command that we don't need. I could always filter it out, though.

No need to delay merging because of this, just a thought for another feature.

matentzn:
I thought of that as well; I think we could extend the --metrics parameter to be n-ary instead of unary to do all that in a second step; I would carefully suggest to make a ticket for that and wait for some funding -> this can be done without breaking the interfaces that exist, but a huge amount of "if else" is going to be required in the metrics class, which now distinguishes only between extended, essential and all..

beckyjackson:
Some minor tweaks: I ran `mvn clean package` to auto-format the code and placed the public methods first for all classes. Just make sure to run `mvn clean package` before committing changes, otherwise we end up with some spurious diffs.

I also noticed this which I didn't noticed last time in `MeasureCommand.java`:
https://github.com/ontodev/robot/blob/8486165b1554d593dca9db95d76c9d4a43360de2/robot-command/src/main/java/org/obolibrary/robot/MeasureCommand.java#L114-L118

Instead of getting an optional value and checking that it's not null, you can just use:
```
 String output = CommandLineHelper.getRequiredValue(line, "output", "an output file must be specified");
```

It was a minor thing so I just included it in my last commit 😄

robot-core/src/main/java/org/obolibrary/robot/metrics/MetricsUtils.java
```diff
@@ -0,0 +1,239 @@
...
+
+  private static boolean isNumeric(String id) {
+    return (isIntegerParsable(id) || isLongParsable(id) || isDoubleParsable(id));
+  }
+
+  private static Set<OWLAxiom> getLogicalAxioms(
+      OWLOntology ontology,
+      Imports includeImportsClosure,
+      boolean skiprules,
+      boolean leavedeclarations) {
```
beckyjackson, line 118: > Actual value of parameter 'leavedeclarations' is always 'false'

Is this something that we need to add an option to the command for? Since this is private, it's never accessed anywhere else.

robot-core/src/main/java/org/obolibrary/robot/metrics/MetricsUtils.java
```diff
@@ -0,0 +1,239 @@
...
+  private static Set<OWLAxiom> getLogicalAxioms(
+      OWLOntology ontology, Imports includeImportsClosure, boolean skiprules) {
+    return getLogicalAxioms(ontology, includeImportsClosure, skiprules, false);
+  }
+
+  private static Set<OWLAxiom> getLogicalAxioms(
+      OWLOntology ontology,
+      Imports includeImportsClosure,
+      boolean skiprules,
+      boolean stripaxiomanno,
```
beckyjackson, line 138: > Actual value of parameter 'stripaxiomanno' is always 'false'

Same as above.

beckyjackson (approved):
Looks great! There are two questionable params in `MetricsUtils` that I commented on, but they aren't critical.

matentzn:
Thank you so much @beckyjackson ! :) 

I will be sure to apply your feedback to my next features before I give it up for review! Very helpful!!

matentzn:
@jamesaoverton Despite the approval, can I get until Monday to fix @beckyjackson remaining suggestions? I wont set it to "draft" until then; I will either have done it by Monday morning or not! :)

robot-core/src/main/java/org/obolibrary/robot/metrics/MetricsUtils.java
matentzn: I think in this case, I would just like to keep the parameter, as I may want to change this behaviour in the future, and this reminds me that this actually was a _thing_. So each time I write code and use this method, I will be reminded to consider this stupid problem: 

Declaration(A) -> kinda like A subclass of owl:Thing, but its not strictly speaking logical - so you may lose signature if you look at the logical axioms without considering the Declarations.

matentzn:
@beckyjackson thanks, I drastically reduced the code in MetricsUtils to the bare minimum now; I left one of these "always same" parameters, but added an annotation to suppress the warning; this parameter is now mainly documentation. @jamesaoverton its all done then for whenever you want to release this!

robot-command/src/main/java/org/obolibrary/robot/MeasureCommand.java
```diff
@@ -0,0 +1,147 @@
...
+      }
+    }
+
+    MeasureOperation.measure(
+        state.getOntology(),
+        reasonerFactory,
+        metricsType,
+        format,
+        output_file,
+        ioHelper.getPrefixes());
```
matentzn, line 143: @beckyjackson Somehow this does not seem to take into account prefixes I add with `-- prefix` to the command; I there must be something wrong with my command perhaps? I though the global properties are somehow handled elsewhere?

matentzn: Maybe it works, it must.. What could be wrong with this:

```
--prefix 'DDPHENOALT: http://purl.obolibrary.org/obo/ddpheno#'
```
For some reason this is not picked up. In any case. I am rambling. Forget about it! :P Sorry.

jamesaoverton:
I'd like to take one more look at this, once the other PRs for https://github.com/ontodev/robot/milestone/7 are merged.

robot-command/src/main/java/org/obolibrary/robot/MeasureCommand.java
jamesaoverton: It should be handled at line 109 where you use the `line` to get a new `IOHelper`.

robot-core/src/main/java/org/obolibrary/robot/metrics/MetricsLabels.java
```diff
@@ -0,0 +1,201 @@
+package org.obolibrary.robot.metrics;
+
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+
+@SuppressWarnings("unused")
```
jamesaoverton, line 7: Can we remove this now?

matentzn: I don't mind - it annoyed me to see the warnings for unused methods - which I left purely because I think I might need them going forward, but I don't have another place for them anymore. But yeah, feel free to delete!

jamesaoverton (approved):

Issue #688, opened 2020-05-07
Robot raise a ONTOLOGY STORAGE ERROR when template gets an rdfxml error


bjonnh:
The error is a bit misleading, as I was just missing a prefix
And it is solved in that case by adding the ero prefix.

(of course Robot has write access to /tmp (it works with merge and extract))

java.io.IOException: errors#ONTOLOGY STORAGE ERROR Could not save ontology to IRI: file:/tmp/tables.owl
        at org.obolibrary.robot.IOHelper.saveOntologyFile(IOHelper.java:1406)
        at org.obolibrary.robot.IOHelper.saveOntology(IOHelper.java:716)
        at org.obolibrary.robot.CommandLineHelper.maybeSaveOutput(CommandLineHelper.java:669)
        at org.obolibrary.robot.TemplateCommand.execute(TemplateCommand.java:190)
        at org.obolibrary.robot.CommandManager.executeCommand(CommandManager.java:248)
        at org.obolibrary.robot.CommandManager.execute(CommandManager.java:192)
        at org.obolibrary.robot.CommandManager.main(CommandManager.java:139)
        at org.obolibrary.robot.CommandLineInterface.main(CommandLineInterface.java:58)
Caused by: org.semanticweb.owlapi.model.OWLOntologyStorageException: ero:0000480
ero:0000041
        at org.semanticweb.owlapi.rdf.rdfxml.renderer.RDFXMLStorer.storeOntology(RDFXMLStorer.java:57)
        at org.semanticweb.owlapi.util.AbstractOWLStorer.store(AbstractOWLStorer.java:99)
        at org.semanticweb.owlapi.util.AbstractOWLStorer.storeOntology(AbstractOWLStorer.java:64)
        at uk.ac.manchester.cs.owl.owlapi.OWLOntologyManagerImpl.saveOntology(OWLOntologyManagerImpl.java:1478)
        at org.obolibrary.robot.IOHelper.saveOntologyFile(IOHelper.java:1398)
        ... 7 more
Caused by: org.semanticweb.owlapi.rdf.rdfxml.renderer.IllegalElementNameException: Illegal Element Name (Element Is Not A QName): ero:0000480
ero:0000041

beckyjackson:
Hi @bjonnh - thanks for bringing this up! `ONTOLOGY STORAGE ERROR` is a bit misleading here, I think it would be more helpful to look at the `IllegalElementNameException` and provide a message related to that.

[...]

PR #783, opened 2020-12-02
Fix template handling of full IRIs

### LINKED ISSUE #782: Template: IRIs, used as IDs, can't end with digits?
I just reran a workflow that I have been using for many months. ROBOT templating is one part of it.

Here's a made-up example of my template:

| class.col                                   | source_med_id                 | source_full_name                             |
|---------------------------------------------|-------------------------------|----------------------------------------------|
| TYPE                                        | ID                            | A http://www.w3.org/2000/01/rdf-schema#label |
| http://purl.obolibrary.org/obo/PDRO_0000024 | http://example.com/resource/1 | aspirin                                      |
| http://purl.obolibrary.org/obo/PDRO_0000024 | http://example.com/resource/2 | tylenol                                      |
| http://purl.obolibrary.org/obo/PDRO_0000024 | http://example.com/resource/3 | lithotabs                                    |

Which I would execute with

```Bash
robot -vvv template  --template numerical_rhs.tsv --output numerical_rhs.ttl
```
That used to work (v?), but now (ROBOT 1.7.2) I get

> 2020-12-01 18:02:36,117 DEBUG org.semanticweb.owlapi.utilities.Injector - Injecting values [[]] on method public void uk.ac.manchester.cs.owl.owlapi.OWLOntologyManagerImpl.setIRIMappers(java.util.Set).
UNKNOWN ENTITY ERROR could not interpret 'http://example.com/resource/1' in row 3, column 2 ("ID") in table "numerical_rhs.tsv".
For details see: http://robot.obolibrary.org/template#unknown-entity-error
org.obolibrary.robot.exceptions.RowParseException: template#UNKNOWN ENTITY ERROR could not interpret 'http://example.com/resource/1' in row 3, column 2 ("ID") in table "numerical_rhs.tsv".
	at org.obolibrary.robot.Template.processRow(Template.java:731)
	at org.obolibrary.robot.Template.generateOutputOntology(Template.java:377)
	at org.obolibrary.robot.TemplateOperation.template(TemplateOperation.java:184)
	at org.obolibrary.robot.TemplateCommand.execute(TemplateCommand.java:148)
	at org.obolibrary.robot.CommandManager.executeCommand(CommandManager.java:248)
	at org.obolibrary.robot.CommandManager.execute(CommandManager.java:192)
	at org.obolibrary.robot.CommandManager.main(CommandManager.java:139)
	at org.obolibrary.robot.CommandLineInterface.main(CommandLineInterface.java:59)

Replacing the terminal digits in the IDs with alphabetical characters solves the problem

| class.col                                   | source_med_id                 | source_full_name                             |
|---------------------------------------------|-------------------------------|----------------------------------------------|
| TYPE                                        | ID                            | A http://www.w3.org/2000/01/rdf-schema#label |
| http://purl.obolibrary.org/obo/PDRO_0000024 | http://example.com/resource/one | aspirin                                      |
| http://purl.obolibrary.org/obo/PDRO_0000024 | http://example.com/resource/two | tylenol                                      |
| http://purl.obolibrary.org/obo/PDRO_0000024 | http://example.com/resource/three | lithotabs                                    |

Does this seem plausible?

I get the same error when using CURIEs like "mydata:1" and  `--prefix  "mydata: http://example.com/resource/"`

But IRIs like http://example.com/resource/1___ are tolerated.

Also, it doesn't appear that the help URL http://robot.obolibrary.org/template#unknown-entity-error resolves to anything beside the top of the template page.

beckyjackson:
Resolves #782

In the issue, `http://example.com/resource/1` is not a *technically* a valid QName, and we updated the processing of IDs/IRIs in templates last May to make sure all IRIs were also valid QNames. This is because the writer would fail on invalid QNames, but it seems like it's only *certain* invalid QNames... Even though it's not valid, OWLAPI seems to still handle it OK as opposed to the CURIEs that were never expanded to IRIs (see https://github.com/ontodev/robot/issues/688).

The QName check in OWLAPI fails every time for `http://example.com/resource/1` so I made a hack around it with a new `isQName` method in IOHelper. It will still fail for CURIEs with undefined namespaces, as we expect.

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
@@ -821,6 +821,24 @@ public class IOHelper {
   }
 
   /**
+   * Given a term string, use the current prefixes to create an IRI.
+   *
+   * @deprecated replaced by {@link #createIRI(String)}
+   * @param term the term to convert to an IRI
+   * @param qName if true, validate that the IRI expands to a QName
+   * @return the new IRI or null
+   */
+  @Deprecated
+  public IRI createIRI(String term, boolean qName) {
+    IRI iri = createIRI(term);
+    // Check that this is a valid QName
+    if (qName && !iri.getRemainder().isPresent()) {
+      return null;
+    }
+    return iri;
+  }
+
+  /**
    * Given a set of term identifier strings, return a set of IRIs.
    *
    * @param terms the set of term identifier strings
@@ -890,29 +908,6 @@ public class IOHelper {
     OWLDataFactory df = manager.getOWLDataFactory();
     OWLDatatype datatype = df.getOWLDatatype(type);
     return df.getOWLLiteral(value, datatype);
-  }
-
-  /**
-   * Determine if a string is a valid QName. Adapted from:
-   *
-   * @see org.semanticweb.owlapi.io.XMLUtils#isQName(CharSequence)
-   * @param iri IRI to check
-   * @return true if valid QName
-   */
-  public boolean hasValidLocalID(IRI iri) {
-    String s = iri.getShortForm();
-    if (0 >= s.length()) {
-      // local ID is empty
-      return false;
-    }
-    for (int i = 0; i < s.length(); ) {
-      int codePoint = Character.codePointAt(s, i);
-      if (!XMLUtils.isXMLNameChar(codePoint)) {
-        return false;
-      }
-      i += Character.charCount(codePoint);
-    }
-    return true;
   }
 
   /**
@@ -1222,9 +1217,9 @@ public class IOHelper {
    *
    * @see org.semanticweb.owlapi.io.XMLUtils#isQName(CharSequence)
    * @param s Character sequence to check
-   * @return true if valid QName
+   * @return true if valid CURIE
    */
-  private static boolean isValidCURIE(CharSequence s) {
+  public static boolean isValidCURIE(CharSequence s) {
     if (s == null || 0 >= s.length()) {
       // string is null or empty
       return false;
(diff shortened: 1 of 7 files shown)


jamesaoverton:
The code looks good, but you know what I'm going to say! :smile: We need some unit tests, please.

beckyjackson:
Done!

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
```diff
@@ -1198,6 +1211,64 @@ public void saveContext(File file) throws IOException {
...
+    return true;
+  }
+
+  /**
+   * Determine if the short form of an IRI contains invalid characters.
+   *
+   * @param iri IRI to check
+   * @return true if no invalid characters found
+   */
+  public static boolean shortFormIsValid(IRI iri) {
```
jamesaoverton, line 1260: Is valid what? CURIE? QName? String?

beckyjackson: It validates that the short form does not contain invalid characters. I guess `shortFormContainsInvalidCharacters` (returns true on invalid) or something like that sounded too long to me but I can change it.

jamesaoverton: My point is that there are no invalid characters in general, without saying what the standard is. There are invalid characters for a CURIE, or a QName, or an IRI, or a Python function name.

I'll think about it a little more and make changes myself.

jamesaoverton:
I asked for three different error messages, but when I went to review this closely only two made sense to me, so I removed one.

PR #796, opened 2021-01-12
Improve template parse error messages

### LINKED ISSUE #795: Better error messages for "encountered unknown..." template errors
See #794 

Whenever the parser throws an "encountered unknown X" where X is one of (I think this is everything):
* some
* only
* min
* max
* exactly 
* value

... we should catch the error and write a better message with a hint on how to fix it, e.g.:

> encountered unknown "some"
hint: you may be missing a property before an expression

This would also provide a link to the docs for the error message explaining that you must have a property before a restriction keyword.

beckyjackson:
Resolves #795

When using a restriction with a non-property entity, you will see:
```
MANCHESTER PARSE ERROR the expression ('test 1' some 'test 4') at row 7, column 4 in table "test.tsv" cannot be parsed: encountered unexpected 'some'
	hint: the term before 'some' must be a property
For details see: http://robot.obolibrary.org/template#manchester-parse-error
```

For all other types of parse errors, you will now see some expected suggestions. These come directly from the parser, but I tried to format them as much as possible to make sense.

e.g., using a class in a restriction for a data property:
```
MANCHESTER PARSE ERROR the expression ('test 1' some 'test 3') at row 5, column 4 in table "test.tsv" cannot be parsed: encountered 'test 3', but expected one of:
	Datatype name
	not
	{ ... }
For details see: http://robot.obolibrary.org/template#manchester-parse-error
```

Or using an undeclared class:
```
MANCHESTER PARSE ERROR the expression ('test 1' some 'test 4') at row 5, column 4 in table "test.tsv" cannot be parsed: encountered 'test 4', but expected one of:
 name
	( ... )
	{ ... }
	Self
For details see: http://robot.obolibrary.org/template#manchester-parse-error
```

### DIFF
--- a/docs/template.md
+++ b/docs/template.md
@@ -310,6 +310,10 @@ AL rdfs:label@en
 
 The provided value cannot be parsed and may not be in proper Manchester syntax. See [Manchester Syntax](http://www.w3.org/2007/OWL/wiki/ManchesterSyntax) for more details. If you are using labels, make sure the labels are defined in the `--input` ontology or using the `LABEL` column. Also ensure that all properties use a label instead of a CURIE or IRI.
 
+When using a restriction (`some`, `only`, `min`, `max`, `exactly`, or `value`) the term that preceeds the restriction must be a property. 
+
+Terms joined using `and` or `or` must be of the same entity type, e.g., you cannot join an object property and a class in an expression.
+
 ### Merge Error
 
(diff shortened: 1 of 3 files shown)


beckyjackson:
I'm not really sure why that first test failed, but it's fixed now...

jamesaoverton (approved):

PR #803, opened 2021-01-25
Switch to GitHub Actions

### LINKED ISSUE #800: Switch to GitHub Actions for testing, and ideally for compiled JARs
My ontodev org has already blown through 75% of our free Travis credits. I like GitHub Actions better than Travis, so let's plan to migrate to GitHub Actions.

We have also relied on build.obolibrary.io Jenkins for JARs from development branches. It would be better to tell GitHub Actions to upload those JARs somewhere.

beckyjackson:
Resolves #800

So we can attach the JAR as an artifact, but unfortunately, when you download it, you get a zip file... which is slightly annoying (clicking this link will download the artifact): https://github.com/ontodev/robot/suites/1896473600/artifacts/37141639

### DIFF
--- a/.github/workflows/java-ci.yml
+++ b/.github/workflows/java-ci.yml
@@ -1,6 +1,8 @@
 name: Java CI
 
-on: [push]
+on:
+  pull_request:
+    branches: [ master ]
 
 jobs:
   build:
--- a/.github/workflows/java-ci.yml
+++ b/.github/workflows/java-ci.yml
@@ -1,0 +1,22 @@
+name: Java CI
+
+on: [push]
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


beckyjackson:
Another question - if we are building an artifact on each push, is that going to be too much?

I was thinking of only doing it on PRs, but then we run into a problem if we update a PR.

balhoff:
@beckyjackson I am using this:

```yaml
on:
  # Triggers the workflow on pull request events for the master branch
  pull_request:
    branches: [ master ]
```

and it runs again whenever a PR branch has new commits pushed to it.

jamesaoverton:
We're really running low on credits, and this seems to be working, so I'm just going to merge.

PR #816, opened 2021-02-02
Do not span all gaps for external/internal

### LINKED ISSUE #693: remove external axioms has different behavior than filter internal axioms
This is probably an error in my understanding. Happy to add to docs if the case.

This gives me the expected behavior:

` robot remove --base-iri PCO --axioms external -i imports/pco_import.owl -o z.owl`

a nice PCO base module

I would expect this to do the same:

`robot filter --base-iri PCO --axioms internal -i imports/pco_import.owl -o z.owl`

however, this preserves some foreign *logical* axioms such as subClassOf

e.g.

```
   <!-- http://purl.obolibrary.org/obo/BFO_0000002 -->

    <owl:Class rdf:about="http://purl.obolibrary.org/obo/BFO_0000002">
        <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/BFO_0000001"/>
    </owl:Class>
```

the input file is attached

[pco_test.owl.txt](https://github.com/ontodev/robot/files/4625713/pco_test.owl.txt)

beckyjackson:
Resolves #693

`filter` was preserving structure in the non-internal axioms. This PR fixes handling external/internal axioms with the `--preserve-structure` option.

I removed the `--preserve-structure false` from the integration test and the result is thee same. With the bug, the result would have extra axioms.


jamesaoverton (approved):
I don't quite get this, but I'll trust the tests.

PR #821, opened 2021-02-04
Update owl-diff version.

balhoff:
This version of owl-diff does not emit a difference when comparing two anonymous ontology IDs. (cc @beckyjackson)


beckyjackson:
@jamesaoverton I can confirm that this fixes the diff issue in #810

jamesaoverton (approved):

PR #837, opened 2021-03-11
Maven updates & clean up

### LINKED ISSUE #805: Clean up Maven output
The output of `mvn` is so noisy that it's hard to see any important messages. I'd like us to spend at least a few hours trying to clean that up, before we inevitably give up in despair.

beckyjackson:
This PR updates Maven plugins and project dependencies to their latest versions. It also adds in a bunch of exclusions to remove the warning messages from packaging. The `mvn` output is still pretty noisy, but I don't think there's anything we can do about those INFO messages.

Resolves #805 #839

Notes:
* `fmt-maven-plugin` 2.9.x is the latest version that supports Java 8. Java 11 is required for 2.10.x and up.
* I had to update a couple of classes to replace deprecated methods
* We have one warning left coming from `ExtendedPosixParser.java` - this class is never used, and even though I deprecated it, it still shows as using a deprecated API. Ideally we can remove this entirely?

**Problem with using classes from `robot-core` in command line tests:**

Java 8 passes, but later versions fail on `CommandLineIT` with
```
java.lang.NoClassDefFoundError: org/obolibrary/robot/IOHelper
```
It seems like the tests fail to include workspace dependencies... Looking at the logs, it looks like all the tests actually run (which is good), and then `compareResults()` fails because the test module cannot find `IOHelper`. We can resolve this by not using `robot-core` classes, and the JAR still works fine. I'm not sure what changed between Java versions to cause this issue, though.

### DIFF
--- a/robot-core/src/test/java/org/obolibrary/robot/DiffOperationTest.java
+++ b/robot-core/src/test/java/org/obolibrary/robot/DiffOperationTest.java
@@ -5,6 +5,7 @@ import static org.junit.Assert.assertFalse;
 
 import java.io.IOException;
 import java.io.StringWriter;
+import java.nio.charset.Charset;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -53,7 +54,9 @@ public class DiffOperationTest extends CoreTest {
     boolean actual = DiffOperation.compare(simple, simple1, writer);
     System.out.println(writer.toString());
     assertFalse(actual);
-    String expected = IOUtils.toString(this.getClass().getResourceAsStream("/simple1.diff"));
+    String expected =
+        IOUtils.toString(
+            this.getClass().getResourceAsStream("/simple1.diff"), Charset.defaultCharset());
     assertEquals(expected, writer.toString());
   }
 
@@ -61,10 +64,9 @@ public class DiffOperationTest extends CoreTest {
    * Compare one ontology to a modified copy with labels in output.
    *
    * @throws IOException on file problem
-   * @throws OWLOntologyCreationException on ontology problem
    */
   @Test
-  public void testCompareModifiedWithLabels() throws IOException, OWLOntologyCreationException {
+  public void testCompareModifiedWithLabels() throws IOException {
     OWLOntology simple = loadOntology("/simple.owl");
     OWLOntology elk = loadOntology("/simple_elk.owl");
 
@@ -74,7 +76,9 @@ public class DiffOperationTest extends CoreTest {
     boolean actual = DiffOperation.compare(simple, elk, new IOHelper(), writer, options);
     System.out.println(writer.toString());
     assertFalse(actual);
-    String expected = IOUtils.toString(this.getClass().getResourceAsStream("/simple.diff"));
+    String expected =
+        IOUtils.toString(
+            this.getClass().getResourceAsStream("/simple.diff"), Charset.defaultCharset());
     assertEquals(expected, writer.toString());
   }
 }
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
           <plugin>
             <groupId>org.sonatype.plugins</groupId>
             <artifactId>nexus-staging-maven-plugin</artifactId>
-            <version>1.6.7</version>
+            <version>1.6.8</version>
             <extensions>true</extensions>
             <configuration>
               <serverId>ossrh</serverId>
@@ -105,7 +105,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-source-plugin</artifactId>
-            <version>2.2.1</version>
+            <version>3.2.0</version>
             <executions>
               <execution>
                 <id>attach-sources</id>
@@ -118,7 +118,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-javadoc-plugin</artifactId>
-            <version>3.0.1</version>
+            <version>3.2.0</version>
             <executions>
               <execution>
                 <id>attach-javadocs</id>
@@ -131,7 +131,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-gpg-plugin</artifactId>
-            <version>1.5</version>
+            <version>1.6</version>
             <executions>
               <execution>
                 <id>sign-artifacts</id>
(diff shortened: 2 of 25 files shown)


beckyjackson:
Uploading the artifact keeps failing. The tests pass locally for me, using Java 13. It looks like the actual test part of the GH Actions passes too.

jamesaoverton:
First of all: This is amazing! Way better than I hoped for!

I tried building all packages locally with Apache Maven 3.6.3 under OpenJDK 8, 11, and 15: `mvn clean verify`. I just pushed  a few small changes to get rid of the last of the warnings that I saw.

Under OpenJDK 8 (but not 11 or 15) I got a few overlapping classes warnings. I excluded these from the HermiT import  in the top-level `pom.xml` to get rid of the warnings.

```
[WARNING] geronimo-stax-api_1.0_spec-1.0.1.jar, xml-apis-1.4.01.jar define 33 overlapping classes: 
[WARNING]   - javax.xml.stream.EventFilter
[WARNING]   - javax.xml.stream.FactoryConfigurationError
[WARNING]   - javax.xml.stream.Location
[WARNING]   - javax.xml.stream.StreamFilter
[WARNING]   - javax.xml.stream.XMLEventFactory
[WARNING]   - javax.xml.stream.XMLEventReader
[WARNING]   - javax.xml.stream.XMLEventWriter
[WARNING]   - javax.xml.stream.XMLInputFactory
[WARNING]   - javax.xml.stream.XMLOutputFactory
[WARNING]   - javax.xml.stream.XMLReporter
[WARNING]   - 23 more...
[WARNING] jaxen-1.1.4.jar, xml-apis-1.4.01.jar define 1 overlapping classes: 
[WARNING]   - org.w3c.dom.UserDataHandler
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See http://maven.apache.org/plugins/maven-shade-plugin/
```

Under all versions I get this warning about missing tests. You can't store an empty directory with git. We could make an empty test. But I think we can just ignore this -- I didn't do anything about it.

```
[warn] Test source directory '/Users/james/Repositories/github/ontodev/robot/robot-maven-plugin/src/test/java' does not exist, ignoring.
```

Under all versions I saw a few last deprecation warnings. Since ExtendedPosixParser has now been deprecated, I decided to `@SuppressWarnings("deprecation")` on it.

```
[WARNING] /Users/james/Repositories/github/ontodev/robot/robot-command/src/main/java/org/obolibrary/robot/ExtendedPosixParser.java: Some input files use or override a deprecated API.
[WARNING] /Users/james/Repositories/github/ontodev/robot/robot-command/src/main/java/org/obolibrary/robot/ExtendedPosixParser.java: Recompile with -Xlint:deprecation for details.
```

Then I saw three more `readFileToString(File)` deprecation warnings, so I fixed those like @beckyjackson did elsewhere.

I think we should enable these deprecation warnings by default, so I did this: http://frequal.com/java/EnableWarningsInMaven.html

Finally, I ran in a clean Docker container (Maven 3.6.3 and OpenJDK 15), and the output was good:

```
docker run -it --rm -v "$(pwd)":/usr/src/mymaven -w /usr/src/mymaven maven:latest mvn clean verify
```

balhoff:
I don't understand why we need the owlapi-distribution exclusions. I thought that if a bunch of dependencies relied on the same transitive dependency, Maven would pick one using an algorithm and you would just have one version of that dependency. What am I getting wrong?

jamesaoverton:
We've tried over and over to get a clean run of the GitHub Actions on this PR, and every time all the test pass but the artifact upload step fails for one or two of the three checks (at random). I don't get it. This is the first significant problem I've hit with GitHub Actions.

matentzn:
see slack, is suspect a race condition..

jamesaoverton:
@matentzn Points out that the error messages point to the same upload URL, e.g. `https://pipelines.actions.githubusercontent.com/TECf0E0FzfDEFJS3tXPjdwD8k5lGC2hgL6JNt2zOfoUtVRCtp4/_apis/resources/Containers/8558120?itemPath=robot.jar%2Frobot.jar`. So it seems like all three checks are trying to upload to the same place, resulting in a race condition. That's not how I thought it would work, but it makes sense. We don't actually need three artifacts uploaded, just one.

@beckyjackson When you get a chance, can you please try uploading just the Java 8 artifact?

robot-core/src/test/java/org/obolibrary/robot/DiffOperationTest.java
```diff
@@ -53,18 +54,19 @@ public void testCompareModified() throws IOException, OWLOntologyCreationExcepti
     boolean actual = DiffOperation.compare(simple, simple1, writer);
     System.out.println(writer.toString());
     assertFalse(actual);
-    String expected = IOUtils.toString(this.getClass().getResourceAsStream("/simple1.diff"));
+    String expected =
+        IOUtils.toString(
+            this.getClass().getResourceAsStream("/simple1.diff"), Charset.defaultCharset());
```
lillieMaiBauer, line 59: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [UI: Usage of GetResource may be unsafe if class is extended (UI_INHERITANCE_UNSAFE_GETRESOURCE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#ui-usage-of-getresource-may-be-unsafe-if-class-is-extended-ui-inheritance-unsafe-getresource).
Calling this.getClass().getResource(...) could give results other than expected if this class is extended by a class in another package.

jamesaoverton: @lillieMaiBauer Are you a person or a bot? This looks like an automated message, and your GitHub profile is pretty bare, so I can't tell.

This code is only used in our test suite to work around annoyances loading test files. There may be a better way to do this. But it should never be extended by any other class.

lillieMaiBauer: Thank you for your quick reply. Since this is a test class, this problem is inapplicable.

jamesaoverton: Ok, thanks.

jamesaoverton:
I've tested this on several of my fancier ontologies with no problems. I'd appreciate some help testing. You can get a ZIP file of the build result here, but you have to download it from your browser, `curl` or `wget` won't work.

https://github.com/ontodev/robot/suites/2408201260/artifacts/51287257

jamesaoverton:
It's been a long time now, and @beckyjackson and @matentzn and I have done all the testing we can, so let's merge this.

PR #841, opened 2021-04-02
WIP - upload artifact only on Java 8

beckyjackson:
Surprisingly, this worked perfectly!


jamesaoverton (approved):
That's good, but now I don't understand why we didn't see the problem when we added the test matrix in #810.

@beckyjackson Please merge this to `master` and then update #837 and we'll see what happens.

PR #852, opened 2021-05-13
Try GitHub Actions with different runner

jamesaoverton:


jamesaoverton:
I tried replacing ubuntu-latest with ubuntu-18.04 and ubuntu-20.04 but they all fail in the same way.

jamesaoverton:
I also tried using the `actions/setup-java@v2` and that also failed.

PR #856, opened 2021-05-14
Report: loosen multiple_equivalent_classes check

### LINKED ISSUE #733: ROBOT report multiple_equivalent_classes perhaps too strict?
ROBOT v. 1.7.0

I am a huge advocate of `multiple_equivalent_classes` -> having multiple definitions is very likely to be a huge design flaw. But may be an exception to this perhaps, which I want to debate until we are all agreed on the pros and cons (@jamesaoverton @beckyjackson @cmungall @balhoff @dosumis).

We recently came across a few of these cases:

```
ax1: A = B and R some C
ax2: A = C or D
```

To start, this case:

```
ax1: A = B and R some C
ax2: A = C and S some D
```

is clearly bad; I think @cmungall would say that you would apply multiple "genus differentinae" to the same class. I don't want to discuss all the ways this is bad, I think we all agree -> hence the `multiple_equivalent_classes` in the first place.

Now about

```
ax1: A = B and R some C
ax2: A = C or D
```

I am wondering whether quite the same concerns apply here, or whether this case should be down-regulated to a warning (I am actually wondering, not trying to convince anyone). To make it explicit, I would like to suggest that 

- for the `multiple_equivalent_classes` check, we ignore equivalent class expressions that are pure _unions of names_, or in OWL API speech: where the expression is an `OWLObjectUnionOf` whose operands are all of type `OWLClass`. (I would _not want to go further in this ticket and think about unions or complex expressions_; lets just not discuss this because it can become much more hairy - this case here is much more frequent). 
- we keep the error level as ERROR (as it is right now)
- we introduce a new WARN check `combined_union_and_equivalent_classes` to cover the case we just removed from `multiple_equivalent_classes`. 

So, to reduce the debate to the key point:

What are the _actual_ risks of asserting:

```
ax1: A = B and R some C
ax2: A = C or D
```

matentzn:
Resolves #733 

This is a proposal based on what we started discussing in #733 but I modified it slightly to follow my personal ideal. Of course this was never discussed, and I am happy to revise:
1. relax the multiple_equivalent_class check to WARN level and exclude "covering axioms" (OR OR OR) from the check altogether. Reducing the WARN level is because I think that asserting logical equivalence between named classes should either be discouraged altogether (this is then tested with ROBOT reason though), or we should be more permissive when there are two of them. The covering axiom argument is discussed in the ticket.
2.  Add a new check, `multiple_logical_definitions` to ensure that no one has two equivalent class axioms with intersections in it. That should _always_ be discouraged.


docs/report_queries/multiple_logical_definitions.md
```diff
@@ -0,0 +1,18 @@
+# Multiple Logical Definitions
+
+**Problem:** A class has more than one logical definitions. This should be avoided, as logical definitions can interact adversely.
```
jamesaoverton, line 3: I don't like "more than one logical definitions", because we allow multiple subClassOf assertions.

What about "A class has more than one equivalent class definition." ?

beckyjackson:
It looks like you only updated the docs, not the queries. Can you please:

* Update `robot-core/src/main/resources/report_queries/multiple_equivalent_classes.rq` to match [`multiple_equivalent_classes.md`](https://github.com/ontodev/robot/blob/5596e22396b9b8f9e65567237f49cd694d58b5f3/docs/report_queries/multiple_equivalent_classes.md)
* Add `robot-core/src/main/resources/report_queries/multiple_logical_definitions.rq` to match [`multiple_logical_definitions.md`](https://github.com/ontodev/robot/blob/5596e22396b9b8f9e65567237f49cd694d58b5f3/docs/report_queries/multiple_logical_definitions.md)?

It seems like this test should have failed because of the missing `multiple_logical_definitons.rq`. That might be another issue to open.

matentzn:
Oops, sorry! I guess I thought the queries would magically materialise themselves from the docs. :D

jamesaoverton:
It's almost magic. There's this Makefile: https://github.com/ontodev/robot/blob/master/docs/Makefile

robot-core/src/main/resources/report_queries/multiple_logical_definitions.rq
```diff
@@ -0,0 +1,16 @@
+# Multiple Logical Definitions
+
+**Problem:** A class has more than one logical definitions. This should be avoided, as logical definitions can interact adversely.
```
beckyjackson, line 3: These should all be commented out (see diff on multiple_equivalent_classes)

docs/report_queries/multiple_logical_definitions.md
matentzn: This makes sense, probably we should also rename the files as well 🗡️

matentzn: Name for that check maybe `multiple_equivalent_class_definitions`

matentzn: Renamed.

beckyjackson:
I think these two names are a bit too close to each other and might cause confusion:

`multiple_equivalent_classes` vs. `multiple_equivalent_class_definitions`

I don't have a better proposal though so... maybe I'll just have to live with it.

Two more things and I think this is in good shape:

1. Make sure to update [this line](https://github.com/ontodev/robot/pull/856/files#diff-55743558c78c5be167d235708bd9c763cb3487bc5c86486e32e8dd4c4d3a19ebR34) with the correct rule name
2. I've kept the [report profile](https://github.com/ontodev/robot/blob/66b930655b94642e3572069efb57db972ca6e154/robot-core/src/main/resources/report_profile.txt) in alphabetical order, so can we move this new check to above `multiple_equivalent_classes`?
Thank you!

matentzn:
Thanks @beckyjackson 

Ideally the other rule would have been renamed to something more specific, but its better to not rename rules I think. Its ok! 80%, good enough I think :D

beckyjackson (approved):

PR #854, opened 2021-05-13
Update deprecated_boolean_datatype.md

matentzn:
One of the sparql queries was missing the correct bindings for the report table.

PR #859, opened 2021-05-14
Added rq file, not just docs (for deprecated datatype query).

matentzn:
In here, I managed to only update the docs: https://github.com/ontodev/robot/pull/854/files

As Becky explained to me [here](https://github.com/ontodev/robot/pull/856#issuecomment-841300359), that did not do the trick :D


beckyjackson (changes requested):
Just make sure the problem and solutions are still commented out 😄

matentzn:
My shame knows no bounds

beckyjackson (approved):

PR #863, opened 2021-05-16
Report: Allowing dcterms for description and title

### LINKED ISSUE #741: dcterms namespace and ROBOT report
Currently this is an ERROR in ROBOT report:

```
property_value: http://purl.org/dc/terms/description "A semi-automatically constructed ontology that merges in multiple disease resources to yield a coherent merged ontology." xsd:string
```

While this isnt:

```
property_value: http://purl.org/dc/elements/1.1/description "A semi-automatically constructed ontology that merges in multiple disease resources to yield a coherent merged ontology." xsd:string
```

The [description relation appears in both](https://dublincore.org/specifications/dublin-core/dcmi-terms/#description) of these legally speaking. Was there ever a discussion on why the one is allowed and the other one is not?

matentzn:
fixes #741

As part of a multiphase process, we allow, at first, both dcterms and dc elements, and will phase out elements next year.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Split equivalent class check [#856]
+- Allow Dublin Core "terms" namespace (`http://purl.org/dc/terms/`) for `description` and `title` properties on ontology. [#741]
 
 ### Fixed
 - Fix printing violations in [`report`] [#823]
--- a/docs/examples/report.tsv
+++ b/docs/examples/report.tsv
@@ -1,18 +1,18 @@
 Level	Rule Name	Subject	Property	Value
 ERROR	missing_ontology_license	https://github.com/ontodev/robot/examples/edit.owl	dc:license	
-ERROR	missing_ontology_title	https://github.com/ontodev/robot/examples/edit.owl	dc11:title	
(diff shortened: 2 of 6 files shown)


beckyjackson:
Sorry, I totally missed this issue.

If it's not too much trouble to implement, I prefer this to be a custom check for the OBO dashboard. These namespaces are technically not correct and we plan to phase them out anyway, so we will just have to change this back later. @jamesaoverton may have a different opinion, but I think it makes sense for ROBOT to keep the more-strict version of this check.

matentzn:
I don't mind either way but this would mean terms, which would mean everything goes red again..

jamesaoverton:
The logic of these revised queries isn't right. We don't want to require *both* predicates, just one, and preferably the `dcterms:description`. The automated tests are failing with rows for all the predicates:

```
ERROR	missing_ontology_description	https://github.com/ontodev/robot/examples/edit.owl	dc11:description	
ERROR	missing_ontology_description	https://github.com/ontodev/robot/examples/edit.owl	dc:description	
```

jamesaoverton (changes requested):
Please fix this so to match either `dce` or `dcterms` both not both.

matentzn:
Ohhh thank you @jamesaoverton 

I have tested all four cases:

1. If there is no title/descriptions the test fail
2. If there is a misspelled title (dc:titles) the test fail (obviously)
3. If there is a terms:title, the test passes
4. If there is a elements:title, the test passes

jamesaoverton (approved):

PR #864, opened 2021-05-16
Report: Fixing duplicate_label_synonym

### LINKED ISSUE #747: Add a check for clashes between an exact synonym and a label
We have checks for duplicate label, duplicate exact syn, but not label=exact_syn

matentzn:
Resolves #747 

This test is a bit hard to wrap around, but the goal of this test, unchanged by this PR, is:

```
An entity shares a label with an exact synonym. This causes ambiguity. Excludes deprecated entities.
```

Hence, this PR:

1. Reduces the scope of the test to comparing labels with *exact synonyms and alternative terms*, excluding now all other synonym types
2. Looking not only at the entity itself (`entity`), but also checking that there is no other entity (`entity2`) that has an exact synonym that is the same as the label of `entity`.


matentzn:
This merits a more careful review probably.

beckyjackson (approved):

beckyjackson:
The query looks correct to me. I just requested a review from @jamesaoverton.

PR #865, opened 2021-05-16
Report: Equivalent class with no genus check

### LINKED ISSUE #844: Suggestion: new ROBOT report check for logical definitions without a genus
I quite often see this:

A = R some C

This nearly always, if not always, points to an error. I suggest a new test:
Severity: WARN
Test:

```


SELECT DISTINCT ?entity ?property ?value ?y WHERE {
 ?entity owl:equivalentClass ?ec .
  ?ec rdf:type owl:Restriction .
  ?ec owl:someValuesFrom ?y .
  ?ec owl:onProperty ?property .
  FILTER (!isBlank(?entity))
  FILTER (!isBlank(?y))
  BIND("No genus in definition" as ?value)
ORDER BY ?entity
```

matentzn:
Resolves #844

See issue. What needs to be determined is whether we all agree this should be `WARN` or `ERROR`; there may be even some that would argue for `INFO`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add links to query documentation for default rules in [`report`] [#879]
 - Ability to restrict [`report`] to base ontology [#872]
+- Add check for equivalent class with no genus to [`report`] [#865]
 
 ### Changed
 - Split equivalent class check [#856]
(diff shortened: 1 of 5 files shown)


beckyjackson:
I think this is a good check and the query works well. I'm not sure I like the word "genus" in the name of it though... is there a different word we can use?

Two (style) notes on the query:

1. You don't need to `BIND` an empty string to value. If you remove [this line](https://github.com/ontodev/robot/blob/de8d2b0d80e5f626a7b6df8d0b12a8e655654c70/robot-core/src/main/resources/report_queries/equivalent_class_axiom_no_genus.rq), the query will still return nothing for `?value`.
2. You don't need to name `?ec`, it can be rewritten as below. This isn't really a big deal, it's just a style thing that I like 🙂 
```
?entity owl:equivalentClass [ rdf:type owl:Restriction ;
                              owl:someValuesFrom ?y ;
                              owl:onProperty ?property ] .
```

beckyjackson:
Oh, also please put the query in the report profile in alphabetical order. Thanks!

matentzn:
Ok, apart from the name, I think I dealt what you suggested! We also forget index.md again, but added it now.

PR #866, opened 2021-05-16
Check lowercase_definition_ should allow numbers

### LINKED ISSUE #842: Add exception to "lowercase_definition" check if it starts with number?
Could we have the "lowercase_definition" [check](http://robot.obolibrary.org/report_queries/lowercase_definition)  to include numbers?

```
PREFIX obo: <http://purl.obolibrary.org/obo/>

 VALUES ?property { obo:IAO_0000115
                    obo:IAO_0000600 }
 ?entity ?property ?value .
 FILTER (!regex(?value, "^[A-Z0-9]"))
}
ORDER BY ?entity
```

Use case would be if the term in question starts with a number (many chemicals and diseases).

matentzn:
Fixes #842

See issue for discussion.

### DIFF
--- a/docs/report_queries/lowercase_definition.md
+++ b/docs/report_queries/lowercase_definition.md
@@ -13,7 +13,7 @@ SELECT DISTINCT ?entity ?property ?value WHERE {
  VALUES ?property { obo:IAO_0000115
                     obo:IAO_0000600 }
  ?entity ?property ?value .
- FILTER (!regex(?value, "^[A-Z]"))
+ FILTER (!regex(?value, "^[A-Z0-9]"))
 }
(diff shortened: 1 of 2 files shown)


beckyjackson:
I don't mind this check and the changes to the query look good. But even if the term in question starts with a number, shouldn't the label not be reused in the definition anyway? I know some people do this anyway so... oh well.

matentzn:
I think this pattern 

```
{label} is a {genus} which {differentiae}
```

is so widely used, like:

```
Alzheimer's is a neurodegenerative disease that usually starts slowly and progressively worsens.
```

beckyjackson (approved):

PR #867, opened 2021-05-16
Report: Adding first pass of built in vocab check

### LINKED ISSUE #835: New ROBOT report checks for problematic assertions about built-ins
In https://github.com/ontodev/robot/issues/833 we found a case where `rdf:type` was being asserted to be an OWL Annotation Property. This can lead to bad behaviour in any OWL processing tool. I don't want ROBOT to fail in that case, but we should let users know that this could be a problem.

So I think we should add a ROBOT report check for this sort of thing.

But beyond this specific case, I'm not sure what to check. Suggestions?

matentzn:
Resolves #835 

Adding a first check to prevent use of rdf:type in an axiom as a subject. Its less than what we wanted, but maybe lets just build the check up slowly slowly.

### DIFF
--- a/docs/report_queries/illegal_use_of_built_in_vocabulary.md
+++ b/docs/report_queries/illegal_use_of_built_in_vocabulary.md
@@ -1,0 +1,18 @@
+# Illegal annotation of built-in vocabulary
+
+**Problem:** Redefining built-in vocabulary should never be done.
+
+**Solution:** Remove any statements about build-in vocabulary
+
+```
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+
+SELECT DISTINCT ?entity ?property ?value WHERE {
+  VALUES ?entity {
+      rdf:type
+    }
+  ?entity ?property ?value
+}
+ORDER BY ?entity
+```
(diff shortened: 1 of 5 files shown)


docs/report_queries/illegal_use_of_built_in_vocabulary.md
```diff
@@ -0,0 +1,18 @@
+# Illegal annotation of built-in vocabulary
```
beckyjackson, line 1: This should match the file name:

> Illegal Use of Built-In Vocabulary

robot-core/src/main/resources/report_profile.txt
```diff
@@ -23,3 +23,4 @@ ERROR	misused_obsolete_label
 ERROR	multiple_definitions
 ERROR	multiple_equivalent_classes
 ERROR	multiple_labels
+ERROR	illegal_use_of_built_in_vocabulary
```
beckyjackson, line 26: Please put in alphabetical order

beckyjackson (changes requested):
Query looks good and this is definitely a needed check in my opinion.

matentzn:
@beckyjackson thanks! PR updated!

beckyjackson:
Everything looks good. Are there any other builtins we want to include before merging this? Maybe `rdfs:subClassOf`, etc?

beckyjackson (approved):
Looks good. I also want @jamesaoverton to take a look before merging.

PR #868, opened 2021-05-16
Report: Adding check for invalid entity URIs

### LINKED ISSUE #790: Add ROBOT report warning for typical IRI typos.
The general architecture of an OBO PURL is:

http://purl.obolibrary.org/obo/CARO_0000008

So we have the OBO URL namespace (http://purl.obolibrary.org/obo/) succeeded by the short form: (in this example CARO_0000008).

These are the most typical faulty variants we have identified as part of @jamesaoverton and my work on the dashboard:
- [ ] https://purl.obolibrary.org/obo/CARO_0000008 (https instead of http)
- [ ] http://purl.obolibrary.org/CARO_0000008 (missing /obo in purl, example)
- [ ] http://purl.org/obo/owl/PATO#PATO_0001025 (not having the appropriate structure at all, usually a problem for relations, example)
- [ ] http://www.obofoundry.org/ro/ro.owl#xyz for RO relations
- [ ] http://purl.org/obo/owl/OBO_REL#role_of

Wouldnt it be cool to have a robot report warning for these?

matentzn:
Resolves #790

This PR adds a check for typical mistakes in entity URIs. Careful review please, I tested the query, but not 1000% sure it will not fire in unintended situations.

### DIFF
--- a/docs/report_queries/invalid_entity_uri.md
+++ b/docs/report_queries/invalid_entity_uri.md
@@ -1,6 +1,6 @@
 # Invalid Entity URI
 
-**Problem:** OBO entities are formatted http://purl.obolibrary.org/obo/IDSPACE_0000000. This format is assumed by many OBO tools. Often, accidentally typos cause entity to be ignored during processing.
+**Problem:** An entity's URI is not formatted correctly. OBO entities are formatted http://purl.obolibrary.org/obo/IDSPACE_0000000. This format is assumed by many OBO tools. Often, accidental typos cause an entity to be malformed, which can cause problems for tools that deal with OBO ontologies.
 
 **Solution:** Fix the entity OBO URI.
 
--- a/docs/report_queries/invalid_entity_uri.md
+++ b/docs/report_queries/invalid_entity_uri.md
@@ -1,0 +1,21 @@
+# Invalid Entity URI
+
+**Problem:** OBO entities are formatted http://purl.obolibrary.org/obo/IDSPACE_0000000. This format is assumed by many OBO tools. Often, accidentally typos cause entity to be ignored during processing.
+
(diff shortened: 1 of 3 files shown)


docs/report_queries/invalid_entity_uri.md
```diff
@@ -0,0 +1,21 @@
+# Invalid Entity URI
+
+**Problem:** OBO entities are formatted http://purl.obolibrary.org/obo/IDSPACE_0000000. This format is assumed by many OBO tools. Often, accidentally typos cause entity to be ignored during processing.
```
beckyjackson, line 3: In other query docs, we've typically formatted the "problem" section as stating what exactly was done wrong. This is minor, but it would be nice to keep it consistent:

> An entity's URI is not formatted correctly ... etc. etc. etc.

beckyjackson (commented):
I worry that this is a little bit too OBO-Foundry-centric for ROBOT. Of course our major users are in the OBO Foundry, but I don't want to discourage people from using this tool if they aren't.

This could be better suited as a custom dashboard check, but I'd like to hear @jamesaoverton 's opinion as well.

matentzn:
This query is broken if it would not work for non OBO ontologies! I thought it was working for everything, basically identifying typical mistakes. So a non obo ontology would not reveal any results here? Right?

beckyjackson:
Ah, my mistake, I totally misread the query! Thanks for correcting me.

matentzn:
Thanks @beckyjackson 

I don't think there is a problem with too many checks. In fact, I hope the number of checks will double over the next year - the more, the better. The only thing to keep in mind is performance!

beckyjackson (approved):

jamesaoverton (approved):
Let's bite the bullet and merge this.

PR #869, opened 2021-05-16
Report: Adding replacedBy check

### LINKED ISSUE #288: Proposed new check: replaced_by should only be used on deprecated classes
I will do this when #267 is done

matentzn:
Resolves #288 

This introduces a check for illegal replaced-by annotation involving non-obsolete terms.

### DIFF
--- a/docs/report_queries/misused_replaced_by.md
+++ b/docs/report_queries/misused_replaced_by.md
@@ -1,0 +1,18 @@
+# Misused Replaced-by Annotation
+
+**Problem:** A replaced-by annotation was used on a non-obsolete class.
+
+**Solution:** Remove the replaced-by annotation or obsolete the class.
+
+```sparql
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+
+SELECT DISTINCT ?entity ?property ?value WHERE {
+ VALUES ?property { <http://purl.obolibrary.org/obo/IAO_0100001> }
+ ?entity ?property ?value .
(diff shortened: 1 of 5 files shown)


docs/report_queries/misused_replaced_by.md
```diff
@@ -0,0 +1,18 @@
...
+
+```sparql
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+
+SELECT DISTINCT ?entity ?property ?value WHERE {
+ VALUES ?property { <http://purl.obolibrary.org/obo/IAO_0100001> }
+ ?entity ?property ?value .
+ FILTER NOT EXISTS { ?entity owl:deprecated true }
+ FILTER (?entity != oboInOwl:ObsoleteClass)
```
beckyjackson, line 15: We don't want to use the replaced by annotation on `oboInOwl:ObsoleteClass` either though, correct?

robot-core/src/main/resources/report_profile.txt
```diff
@@ -23,3 +23,4 @@ ERROR	misused_obsolete_label
 ERROR	multiple_definitions
 ERROR	multiple_equivalent_classes
 ERROR	multiple_labels
+ERROR	misused_replaced_by
```
beckyjackson, line 26: Please put this in alphabetical order

beckyjackson (changes requested):

docs/report_queries/misused_replaced_by.md
matentzn: Of course :D

beckyjackson (approved):

PR #870, opened 2021-05-16
Report: Adding undeclared synonymtype and subsettype checks

### LINKED ISSUE #861: Proposed check: Undeclared subset
This happens all the time because ROBOT extract strips these out, so its good to add a check to ensure they werk correctly.

```
PREFIX oio: <http://www.geneontology.org/formats/oboInOwl#>

   ?c oio:inSubset ?entity .
   FILTER NOT EXISTS { ?entity rdfs:subPropertyOf oio:SubsetProperty } 
   BIND(oio:inSubset AS ?property)
   BIND("Subset not declared." as ?value)
}
ORDER BY ?entity
```

### LINKED ISSUE #862: Proposed check: Undeclared synonym type
As ROBOT extract (slme) excludes the synonym type declarations, its good to check for this to ensure they work correctly.

```
PREFIX oio: <http://www.geneontology.org/formats/oboInOwl#>

       ?axiom oio:hasSynonymType ?entity .
       FILTER NOT EXISTS { ?entity rdfs:subPropertyOf oio:SynonymTypeProperty } 
       BIND(oio:hasSynonymType AS ?property)
}
ORDER BY ?entity
```

matentzn:
Resolves #862 
Resolves #861 

This pull request adds two new checks, for undeclared synonym types and undeclared subsets. These can cause problems in OBO serialisations and are easily removed accidentally because of ROBOT slme extract, for example.

### DIFF
--- a/docs/report_queries/missing_subset_declaration.md
+++ b/docs/report_queries/missing_subset_declaration.md
@@ -1,0 +1,17 @@
+# Missing Subset Declaration
+
+**Problem:** A subset is used in an annotation, but is not properly declared as a child of oboInOwl:SubsetProperty. This can cause problems with conversions to OBO format.
+
+**Solution:** Make the subset a child of oboInOwl:SubsetProperty.
+
+```sparql
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+
+SELECT DISTINCT ?entity ?property ?value WHERE {
+   VALUES ?property { oboInOwl:inSubset }
+   ?value ?property ?entity .
+   FILTER NOT EXISTS { ?entity rdfs:subPropertyOf oboInOwl:SubsetProperty }
+}
+ORDER BY ?entity
+```
--- a/docs/report_queries/missing_subset_declaration.md
+++ b/docs/report_queries/missing_subset_declaration.md
@@ -9,9 +9,9 @@ PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
 PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
 
 SELECT DISTINCT ?entity ?property ?value WHERE {
-   VALUES ?property { oboInOwl:inSubset }
-   ?value ?property ?entity .
-   FILTER NOT EXISTS { ?entity rdfs:subPropertyOf oboInOwl:SubsetProperty }
+    VALUES ?property { rdfs:subPropertyOf }
+    ?x oboInOwl:inSubset ?entity .
+    FILTER NOT EXISTS { ?entity ?property oboInOwl:SubsetProperty }
 }
 ORDER BY ?entity
 ```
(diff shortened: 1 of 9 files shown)


docs/report_queries/missing_subset_declaration.md
```diff
@@ -0,0 +1,17 @@
+# Missing Subset Declaration
+
+**Problem:** A subset is used in an annotation, but is not properly declared as a child of oboInOwl:SubsetProperty. This can cause problems with conversions to OBO format.
+
+**Solution:** Make the subset a child of oboInOwl:SubsetProperty.
+
+```sparql
```
beckyjackson, line 7: This query currently returns a line in the report each time the subset is used as an object to `oboInOwl:inSubset`. That greatly increases the number of errors when really, this should just be one error.

If we rewrite the query slightly, we can get just one error per subset:

```
    VALUES ?property { rdfs:subPropertyOf }
    ?x oboInOwl:inSubset ?entity .
    FILTER NOT EXISTS { ?entity ?property oboInOwl:SubsetProperty }
}

File: docs/report_queries/missing_synonymtype_declaration.md [hunk 2]
```diff
@@ -0,0 +1,17 @@
+# Missing Synonym Type Declaration
+
+**Problem:** A synonym type is used in an annotation, but is not properly declared as a child of oboInOwl:SynonymTypeProperty. This can cause problems with conversions to OBO format.
+
+**Solution:** Make the synonym type a child of oboInOwl:SynonymTypeProperty.
+
+```sparql
```
beckyjackson, line 7: Same comment as the `missing_subset_declaration` rule. This query could be rewritten using the same pattern to return one error per synonym type, instead of errors every time it's used.

robot-core/src/main/resources/report_profile.txt
```diff
@@ -23,3 +23,5 @@ ERROR	misused_obsolete_label
 ERROR	multiple_definitions
 ERROR	multiple_equivalent_classes
 ERROR	multiple_labels
+WARN	missing_subset_declaration
```
beckyjackson, line 26: Please put these in alphabetical order

beckyjackson (changes requested):
I think these are unique enough to the OBO Foundry that including them won't cause issues for non-OBO-Foundry users. It doesn't hurt to add extra checks because they're so small, but is there a point where we could end up with *too many* report checks?

matentzn:
Awesome thanks @beckyjackson for that very good suggestion to change the query pattern. I think its all changed!

test.sh
```diff
@@ -0,0 +1 @@
+#./bin/robot report -i $ONT --output results/report_uberon.tsv
```
beckyjackson, line 1: I think this file was mistakenly added. Can you remove it? Thanks!

beckyjackson (changes requested):
Please remove the `test.sh` file, but other than that, this looks good.

test.sh
matentzn: Yep oops.

beckyjackson (approved):

PR #874, opened 2021-05-18
Fix entity rendering in report when value is an entity

### LINKED ISSUE #855: HTML Report should render entities, not only literals
Right now, report only renders values that are literals. I think it would be nice if entities were rendered the same way as in the other two columns.

beckyjackson:
Resolves #855

Oddly enough, the Manchester render doesn't like `OWLObjects` so I switched the Violations to use `OWLEntities` and now the values render correctly when they're not just literals.

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
@@ -944,7 +944,7 @@ public class ReportOperation {
         OWLEntity e = dataFactory.getOWLClass(ioHelper.createIRI(property));
         if (value != null) {
           IRI valIRI = ioHelper.createIRI(value);
-          if (valIRI != null && !valIRI.toString().contains(" ")) {
+          if (valIRI != null) {
             OWLEntity v = dataFactory.getOWLClass(valIRI);
             violation.addStatement(e, v);
           } else {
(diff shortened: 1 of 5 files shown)


robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
```diff
@@ -944,13 +944,14 @@ private static String getQueryResultOrNull(QuerySolution qs, String var) {
         OWLEntity e = dataFactory.getOWLClass(ioHelper.createIRI(property));
         if (value != null) {
           IRI valIRI = ioHelper.createIRI(value);
-          if (valIRI != null) {
-            violation.addStatement(e, valIRI);
+          if (valIRI != null && !valIRI.toString().contains(" ")) {
```
matentzn, line 947: Seems like `!valIRI.toString().contains(" ")` should belong into ioHelper.createIRI()? Just a suggestion to push potential problems upstream.

matentzn (approved):
Nice!

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
beckyjackson: I think you're right about this. I'll change it so that `IOHelper.createIRI` returns `null` if the IRI contains a space.

matentzn: cool

jamesaoverton (approved):

PR #875, opened 2021-05-18
Add links for builtin rules to HTML reports

### LINKED ISSUE #871: HTML report should render link to implemented check
in column 2 (Rule), it would be great if a click on a check could bring you straight to the ROBOT docs.

image

@jamesaoverton, @beckyjackson could you take this on perhaps in the same go as https://github.com/ontodev/robot/issues/855?

beckyjackson:
Resolves #871

I'm not sure I'm super happy with this implementation, but aside from maintaining a list of builtin queries in the `Report` class, this is the best way I could think to do it. I don't want another list to add to every time we add a new default report query. The queries now have an extra layer of info - a boolean if they are builtin or not.

When we change things in Report, I need to deprecate old methods to maintain backwards compatibility. These deprecated methods are starting to build up. Are we planning on removing them at any point?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Links for default rules in HTML outputs for [`report`] [#875]
+
 ### Fixed
 - Fix printing violations in [`report`] [#823]
 - Fix handling of `rdf:type` in [`export`] [#834]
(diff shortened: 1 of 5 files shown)


jamesaoverton:
Sorry, but I don't understand why this change touches so many lines of code. Can't the HTML renderer tell the difference between a built-in rule and a custom rule, and add links to the built-in rules at that point?

beckyjackson:
We strip the rules down to their names, so unless we keep a list of
built-in rules hard coded somewhere, ROBOT can’t tell the difference. Hard
coding the list would remove a lot of the changes, but it’s another list to
maintain that we need to remember to update every time we add a new rule.

We could also keep the custom rules with their “file” prefix, but it would
look messier in the report output. I’m open to any other ideas - I just
wanted to do this without changing how the report looks.

jamesaoverton:
Ok, thanks. I'll take another look and think about it some more.

jamesaoverton:
There's something that's bothering me about this PR. I think we should try storing the profile information (and more profile information) in the report, and see if that looks better.

beckyjackson:
Replaced by #879

PR #879, opened 2021-06-03
Add ReportQuery objects to Report

### LINKED ISSUE #871: HTML report should render link to implemented check

beckyjackson:
This is an alternative to #875 that allows us more flexibility in the future. Resolves #871.

This adds a new `ReportQuery` object that carries the details of the "rules" including rule name, query contents, violation level, a list of the found violations, and an optional URL for documentation. If you are generating an HTML report, the default report queries will have links to their documentation.

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
@@ -288,8 +288,7 @@ public class ReportOperation {
       }
       queryString = String.join("\n", lines);
       // Use the query to get violations
-      List<Violation> violations =
-          getViolations(ioHelper, dataset, queryName, queryString, options);
+      Set<Violation> violations = getViolations(ioHelper, dataset, queryName, queryString, options);
       // If violations is not returned properly, the query did not have the correct format
       if (violations == null) {
         throw new Exception(String.format(missingEntityBinding, queryName));
@@ -440,8 +439,7 @@ public class ReportOperation {
       }
       queryString = String.join("\n", lines);
       // Use the query to get violations
-      List<Violation> violations =
-          getViolations(ioHelper, dataset, queryName, queryString, options);
+      Set<Violation> violations = getViolations(ioHelper, dataset, queryName, queryString, options);
       // If violations is not returned properly, the query did not have the correct format
       if (violations == null) {
         throw new Exception(String.format(missingEntityBinding, queryName));
@@ -846,7 +844,7 @@ public class ReportOperation {
    * @return List of Violations
    * @throws IOException on issue parsing query
    */
-  private static List<Violation> getViolations(
+  private static Set<Violation> getViolations(
       IOHelper ioHelper,
       Dataset dataset,
       String queryName,
@@ -910,9 +908,9 @@ public class ReportOperation {
    * @return list of Violation objects
    * @throws Exception on malformed query
    */
-  private static List<Violation> getViolationsFromResults(
+  private static Set<Violation> getViolationsFromResults(
       IOHelper ioHelper, String queryName, ResultSet violationSet, Integer limit) throws Exception {
-    List<Violation> violations = new ArrayList<>();
+    Set<Violation> violations = new HashSet<>();
 
     // Counter for stopping at limit
     int c = 0;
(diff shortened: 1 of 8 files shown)


beckyjackson:
This fails with:
```
Integration test file '../docs/examples/results/report.tsv' is different from example file '../docs/examples/report.tsv'
```

... but it passes locally.

I'm wondering if there's some sort of ordering difference in each report query? I sorted the report queries by name, but maybe the actual violations within each report query are getting sorted different. Any ideas?

We could sort these, but the Violation objects are kind of complex and we couldn't sort until rendering, which might get messy...

beckyjackson:
Without using `Collections.sort` and a custom comparator to make sure the report ends up the same across systems, the tests kept failing. I tried using sets instead hoping that that would maintain order, but it still failed.

To resolve this, I changed the diffing of TSV files in `CommandLineIT` so that we compare the set of lines, in no particular order. I can't think of a case here where we always expect TSVs to be in the same order...

jamesaoverton:
I think changing CommandLineIT is a good solution. I'll take one more look at the code tomorrow. Thanks!

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
```diff
@@ -893,9 +908,9 @@ private static String getQueryResultOrNull(QuerySolution qs, String var) {
    * @return list of Violation objects
    * @throws Exception on malformed query
    */
-  private static List<Violation> getViolationsFromResults(
+  private static Set<Violation> getViolationsFromResults(
```
jamesaoverton, line 911: You said you tried Set here to fix the diffs, but that didn't work. Do you really want Set now or is List better?

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
```diff
@@ -610,29 +612,42 @@ public static boolean processReport(Report report, String outputPath, Map<String
...
-   * @return map of rule name and query string
-   * @throws IOException on any issue reading the query file
-   * @throws URISyntaxException on issue converting URL to URI
+   * @param profile Map of rule name to violation level
+   * @return List of Report Query objects
+   * @throws IOException on problem getting query strings
+   * @throws URISyntaxException on problem getting query strings
    */
-  private static Map<String, String> getQueryStrings(Set<String> rules)
+  private static Set<ReportQuery> getReportQueries(Map<String, String> profile)
```
jamesaoverton, line 622: Do you think it would be better to preserve the order of the queries that the user provides, or not?

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
beckyjackson: I don't care either way, I can switch it back to lists if you prefer?

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
beckyjackson: We didn't in the past, but we can switch this to be a `LinkedHashMap` which should preserve the order

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
jamesaoverton: Yes please.

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
jamesaoverton: I was thinking a List.

Sorry to be picky. I'm quite happy with this PR now.

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
beckyjackson: Done!

jamesaoverton (approved):
All right, let's just do this.

PR #882, opened 2021-06-18
Add check for malformed IRI in IOHelper.createIRI

### LINKED ISSUE #880: ROBOT should fail hard if CURIE prefix is unknown
- [ ] Seems right now that robot templates does not fail hard when a prefix is unknown - instead an IRI is created in the form of a CURIE.

- [ ] I was wishing to mention that as well in general - I did have some cases what playing with SPARQL for example that I accidentally introduced curies as entity ids, and it would be kind of great if I could know about this: Maybe I should make a ROBOT report check for that?

beckyjackson:
Resolves #880

This is a bit more sane than trying to do a regex match for a valid URL. If the output IRI is just a CURIE, this will fail (e.g., undefined prefix when running `template`). If the output IRI has a space it it, this will also fail (which we were just checking for a space before).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix behaviour of `--preserve-structure` when using internal or external axiom selectors for [`remove`] or [`filter`] [#816]
 - Fix value rendering for entities in [`report`] [#874]
 
+### Changed
+- Do not allow malformed IRIs to be returned by `IOHelper` [#882]
+
 ## [1.8.1] - 2021-01-27
 
 ### Fixed
@@ -253,6 +256,7 @@ First official release of ROBOT!
 [`template`]: http://robot.obolibrary.org/template
 [`validate`]: http://robot.obolibrary.org/validate
 
+[#882]: https://github.com/ontodev/robot/pull/882
 [#874]: https://github.com/ontodev/robot/pull/874
 [#858]: https://github.com/ontodev/robot/pull/858
 [#850]: https://github.com/ontodev/robot/pull/850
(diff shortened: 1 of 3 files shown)


matentzn:
This is what I suggest as well - just consider the rare cases where this will fail:

1) ontology IRIs which are by default when saved with Protege `urns`
2) Its strictly speaking allowed to use URNs instead of URLs for IRIs as IRIs are extensions of URIs, but I think this is sane for 99.95% of all cases.

beckyjackson:
This won’t impact ROBOT loading ontologies that have the Protege URN, since
it’s only for creating new IRIs. I don’t think that would ever be used as a
new IRI in a template, but perhaps this is too strict?

beckyjackson:
We could also add to the method - if the IRI starts with `urn:` and doesn't have a space in it, it should be assumed valid. We could implement a check for URNs but I haven't seen them used in the wild, so I don't know if it's worth it (https://stackoverflow.com/a/5492927).

matentzn:
I would kinda tend towards adding URN checking support, but defer to your better judgement. A question that I never quite answered to myself: what is the advantage of using `try: new URL(...)` compared to using a regular expression for matching URLs?

beckyjackson:
Here's another option: we could use the [URLValidator](http://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/UrlValidator.html) from `commons-validator`, but that requires adding an additional dependency. This validator is more strict than `new URL(...)`, though, as just passing `http://` to that would pass - which would not be a valid IRI. `new URL(...)` will catch *most* issues, though. Maybe this is important enough to add another dependency, though.

I have concerns about just using regex to match a URL because there are so many variations on what a "good" regex pattern for a URL is. I just prefer to use existing and tested tools.

jamesaoverton (approved):

PR #888, opened 2021-07-17
Fix deprecated class query

### LINKED ISSUE #884: ROBOT report: broken deprecated_class_reference check
I just noticed something odd :

When I download the deprecated_class_reference check from http://robot.obolibrary.org/report_queries/deprecated_class_reference I get a lot of (correct) results:

```
robot --version
ROBOT version 1.8.1
robot query -i hp-edit.owl --query deprecated_class_reference.sparql dep.txt
```

When I run robot report:
```
robot report -i hp-edit.owl -o report.txt
```

I get none. On closer inspection, the docs were simply not updated, but the check (sparql query) was.. And it now has a few issues:

This seems to make no sense:

```
   FILTER EXISTS { ?property a owl:ObjectProperty }
   FILTER EXISTS { ?property a owl:DatatypeProperty }
```

There are many other problems with this query as well, I think its high priority to fix..

beckyjackson:
Resolves #884

The query was actually missing the part that checks if the deprecated class is a subclass of something else, it was only checking if *something else* was a subclass of the deprecated class. I also changed the `FILTER EXISTS` to use `VALUES` instead to match the other unions.

I ran `make all` in `docs/` and it looks like the queries in the docs are quite out of date. We should make sure to run this every time we update a report query.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -44,7 +44,7 @@ report_queries/index.md: ../robot-core/src/main/resources/report_profile.txt | r
 	echo 'with their default severity levels.' >> $@
 	echo 'See [`report`](../report) for more details.' >> $@
 	echo '' >> $@
-	echo '| Name  | Default Severity  ' >> $@
+	echo '| Name  | Default Severity' >> $@
 	echo '|-------|-------------------' >> $@
 	awk -F'	' '{x=$$2; gsub(/_/," ",$$2); print "| [" $$2 "](" x ")  | " $$1}' $< \
 	>> $@
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add links to query documentation for default rules in [`report`] [#879]
 - Ability to restrict [`report`] to base ontology [#872]
+- Add check for equivalent class with no genus to [`report`] [#865]
+
+### Changed
+- Split equivalent class check [#856]
 
 ### Fixed
 - Fix printing violations in [`report`] [#823]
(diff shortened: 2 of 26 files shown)


beckyjackson:
@matentzn when you get a chance, can you see if this works for you?

matentzn:
The query does not terminate on large ontologies.. Looking into it, but maybe you can spot something

beckyjackson:
@jamesaoverton this PR is ready for review, but it binds the blank node as "blank node" (which we agreed to switch to). Do you want me to change all the blank nodes to "blank node" in this same PR? Or should that be done separately?

matentzn:
I would say lets do this all at once - I can also do it, your time is maybe a bit too precious for such menial tasks :D Up to you. But since we define the new default behavior here (binding to `blank node` if blank) we should probably apply this consistently.

jamesaoverton:
Ok. @matentzn please update all the queries to use "blank node". Thanks!

matentzn:
@beckyjackson before I get into this: the branch here is so massively add odds with master - does it have to be redone? I am not sure exactly why what has changed where - let me know if you prefer me to redo, or whether you can fix conflicts?

beckyjackson:
I tried to remake the docs, but it looks like something has gone wrong with the new changes to the Makefile. Let's abandon this branch and make the changes to the deprecated class query in a new branch. Do you want to do that or should I @matentzn?

docs/Makefile
```diff
@@ -31,7 +31,8 @@ report_queries:
 # Convert each SPARQL file to a Markdown file.
 report_queries/%.md: ../robot-core/src/main/resources/report_queries/%.rq | report_queries
 	awk '/^PREFIX/ && !x {print "```sparql"; x=1} 1' $< \
-	| sed 's/^# \? \?//' \
+	| sed 's/^#//' \
+	| sed 's/^ //' \
```
jamesaoverton, line 35: FYI I changed this bit of the `docs/Makefile` on another branch to make the indentation better, which changes a all the `docs/report_queries/*.md` files.

matentzn:
My worry is that this branch has a bit more than just the changes to the deprecated query.. Look for example at this commit (and some [more](https://github.com/ontodev/robot/pull/888/commits)):
https://github.com/ontodev/robot/pull/888/commits/70d1a42832cad96f7fbe9c621d40182ad84913bf

I think I would lose orientation if I had to replicate the branch myself.. If you could do it, and believe it wouldn't take you too long, I would feel safer if you did it...

beckyjackson:
Closed in favor of #902

Issue #873, opened 2021-05-17
Report: Bind anonymous nodes to a generic label to avoid cluttering diff


matentzn:
I would like to suggest to generally bind blank nodes from sparql queries to some generic string. Its annoying to have huge diffs everytime due to the random number assigned to these.. There are probably more sophisticated solutions, but I think this is the best low effort, medium reward one.

```
PREFIX owl: <http://www.w3.org/2002/07/owl#>

SELECT DISTINCT ?entity ?property ?value WHERE {
 VALUES ?property { owl:equivalentClass }
 ?entity ?property ?value1 .
 ?entity ?property ?value2 .
 FILTER (?value1 != ?value2)
 FILTER (!isBlank(?entity))
 BIND(IF(isBlank(?value1),"anonymous expression",?value1) as ?value)
}
ORDER BY ?entity
```

matentzn:
I will check this again.

PR #889, opened 2021-07-18
884 fix patch

matentzn:
Two suggestions @beckyjackson :

1) We should ignore `some-obsolete subclassOf Thing.` this error is so frequent, it will totally pollute the test.
2) In an attempt to reduce spurious diffs over ROBOT report (see https://github.com/ontodev/robot/issues/873) I would suggest to avoid blank nodes in robot report.

### DIFF
--- a/robot-core/src/main/resources/report_queries/deprecated_class_reference.rq
+++ b/robot-core/src/main/resources/report_queries/deprecated_class_reference.rq
@@ -20,6 +20,7 @@ SELECT DISTINCT ?entity ?property ?value WHERE {
            ?property ?value .
    FILTER ( ?value NOT IN (oboInOwl:ObsoleteClass, owl:Thing) )
   }
+  UNION
   {
    VALUES ?property {
      owl:equivalentClass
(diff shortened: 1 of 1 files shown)


matentzn:
Something is wrong wait a moment..

matentzn:
Ok now

beckyjackson:
I think "Use in anonymous expression" should be rephrased as just "anonymous expression", since that's the entity we're referring to.

beckyjackson (changes requested):
Thanks! I just see two things:
* Remove `FILTER ( ?value != oboInOwl:ObsoleteClass )` at line 32; a deprecated class should never be equivalent or disjoint with `oboInOwl:ObsoleteClass`, only a subclass of it
* The anonymous expression check is duplicated. The first one should be removed, and just use the one where you're binding the string as `?entity`.

matentzn:
> The anonymous expression check is duplicated. The first one should be removed, and just use the one where you're binding the string as ?entity.

I wanted to increase the readability of cases where the axiom in question is a simple existential restriction; if you check the FILTER() expressions in both cases, you will find that one of these "duplicates" looks for cases were the entity is named, and then can be printed as part of the robot command.. Does this make sense?

beckyjackson:
OK - I don't think there should be duplicates within the report because that will conflate the number of issues. Both of these return anonymous expressions, and the first one just returns a (very unhelpful) blank node ID, which won't help the user locate the usage anyway, so I think it should be removed:
```
{
 VALUES ?property {
   owl:someValuesFrom
   owl:allValuesFrom
 }
 ?value a owl:Class ;
        owl:deprecated true .
 ?entity ?x ?rest .
 ?rest a owl:Restriction ;
       ?property ?value .
 FILTER(!isBlank(?entity))
}
UNION
{
 VALUES ?property {
   owl:someValuesFrom   
   owl:allValuesFrom
 }
 ?value a owl:Class ;
        owl:deprecated true .
 ?e ?x ?rest .
 ?rest a owl:Restriction ;
       ?property ?value .
 FILTER(isBlank(?e))
 BIND("anonymous expression" as ?entity)
}
```

PR #890, opened 2021-07-20
Update deprecated_class_reference.rq

matentzn:
Takes care of performance issue discussed on slack (the other suggestions were too slow on large ontologies).

### DIFF
--- a/docs/report_queries/deprecated_class_reference.md
+++ b/docs/report_queries/deprecated_class_reference.md
@@ -12,73 +12,58 @@ PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
 PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
 
 SELECT DISTINCT ?entity ?property ?value WHERE {
- {
-  VALUES ?property {
-    rdfs:subClassOf
+  {
+   VALUES ?property {
+     rdfs:subClassOf
+   }
+   ?entity a owl:Class;
+           owl:deprecated true ;
+           ?property ?value .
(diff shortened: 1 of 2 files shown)


beckyjackson:
Are we OK with "entity" being empty? That last `UNION` doesn't have an `?entity` binding, so it will only show the property & value. The entity is a blank node, anyway, for which we were referencing blank node IDs before. Blank node IDs are pretty unhelpful, so I think either leaving the column blank or binding "blank node" as `?entity` are both good options.

beckyjackson:
I can also confirm that this is MUCH faster 🙂

matentzn:
I dont mind either way! You can decide what to bind, not to bind.. I think empty column is fine.

beckyjackson:
@jamesaoverton - should we leave the "entity" column blank on a blank node, or bind "blank node" for this column?

jamesaoverton:
I prefer "blank node". But if we're doing something else in other checks then it's more important to be consistent with that.

beckyjackson:
Currently in other checks, we use a blank node ID, which isn't helpful. I believe @matentzn mentioned in another issue replacing these blank node IDs.

jamesaoverton:
@matentzn is off for a few days. Let's wait for him to chime in.

matentzn:
I prefer `blank node`, see https://github.com/ontodev/robot/issues/873

And yes, we should do this everywhere across the board. :) I volunteer if no one else wants.

beckyjackson:
@matentzn - I can take care of it, but it won't be for a little bit. Can you just update this PR to use "blank node" and then I'll merge it?

PR #902, opened 2021-08-18
Fix deprecated class query

### LINKED ISSUE #884: ROBOT report: broken deprecated_class_reference check

beckyjackson:
Resolves #884

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix behaviour of `--preserve-structure` when using internal or external axiom selectors for [`remove`] or [`filter`] [#816]
 - Fix duplicate_label_synonym [#864]
 - Fix value rendering for entities in [`report`] [#874]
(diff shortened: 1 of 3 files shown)


jamesaoverton:
I think this is good, but I'd appreciate a review from @matentzn.

jamesaoverton (approved):

matentzn (approved):
I have tested this with existential restrictions as well as subclass of axioms, not problems. Fast and doing what I think it should!

PR #903, opened 2021-08-23
Update IOHelper to correctly serialise OBO with check=false

### LINKED ISSUE #896: Branching logic for saving OBO files causes confusingly different behavior
followed on from here: https://github.com/obophenotype/uberon/pull/2019

Problem: in this section of code, robot makes a decision about one of two methods to save an obo file:

https://github.com/ontodev/robot/blob/5689f6f4db0efc3e85a5180f2e1270e370e8ba9b/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java#L1702-L1755

The decision is made based on presence of checkOBO. However, the behavior differs in ways that have nothing to do with checkOBO. In particular, if checkOBO is true, the native owlapi interface is used and tag comments are lost. These are crucial to most ontologies that use obo (for diffs etc).

The quick fix here is to NEVER use the native method, just create the OWLAPIOwl2Obo object directly

However, we should also investigate more. Protege never loses the tag comments. Perhaps there is similar ugly branching code in Protege. Or perhaps there is a semi-secret way to use the owlapi to preserve tag comments.

matentzn:
Resolves #896

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
@@ -1704,6 +1704,8 @@ public class IOHelper {
       throws IOException {
     // first handle any non-official output formats.
     // currently this is just OboGraphs JSON format
+    format.setParameter(OBODocumentFormat.VALIDATION, checkOBO);
+
     if (format instanceof OboGraphJsonDocumentFormat) {
       FromOwl fromOwl = new FromOwl();
       GraphDocument gd = fromOwl.generateGraphDocument(ontology);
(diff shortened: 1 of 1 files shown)


matentzn:
Ok, I have throughly tested the following 
- automatically (robot 1.8.1 vs robot 1.9 snapshot, then `diff`): pato ddpheno pato/pato-base ddpheno/ddpheno-base obi uberon uberon/uberon-base
- Manually (using git diff): uberon-edit.obo, fbbt-edit.obo and pato-edit.obo. 

I am convinced that:
- The change introduced does not do anything apart from adding `! label` comments to the OBO format.

image

- The change is not a breaking change, as many axioms already have that "comment". The reason why some do not is that the current rendering mode does not look at imports to determine the labels of entities.

matentzn:
As soon as this is approved, I will build a development version of ODK and run a few more pipelines.

PR #907, opened 2021-08-31
Handle error when saving as JSON

### LINKED ISSUE #906: Handle errors properly when saving to OboGraphJsonDocumentFormat
I think we're missing error handling code for this block:

https://github.com/ontodev/robot/blob/master/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java#L1709

beckyjackson:
Resolves #906

### OBO Graph Error

This error occurs when ROBOT is unable to convert the ontology into an [OBO Graphs](https://github.com/geneontology/obographs) object while saving an ontology in JSON format. This may be due to problematic annotations, so you can create a subset using [filter](/filter) or [remove](/remove) containing only the necessary annotations and save that as JSON, for example (keeping only labels and definitions):
```
robot remove --input ont.owl \
  --select annotation-properties \
  --exclude-term rdfs:label \
  --exclude-term IAO:0000115 \
  --output ont.json
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix duplicate_label_synonym [#864]
 - Fix value rendering for entities in [`report`] [#874]
 - Fix OBO serialisation issues when using `--check false` [#896]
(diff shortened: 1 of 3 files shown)


jamesaoverton (approved):
Thanks!

PR #916, opened 2021-10-15
Fixing queries to accomodate for blank nodes

### LINKED ISSUE #873: Report: Bind anonymous nodes to a generic label to avoid cluttering diff

matentzn:
Resolves #873

This PR addresses the issue that blank nodes are not printed deterministically, leading to spurious diffs of ROBOT reports in version control.

Note
- I have reviewed all queries in ROBOT to search for potential blank nodes (ignoring blank nodes in the `?property` position).
- In some cases rather than adding blank node handling I added a restriction to named entities instead `(FILTER (!isBlank())`. In all such cases I believe this was the intention, and ommiting the FILTER was an oversight.
- I simplified the `equivalent_class_axiom_no_genus.rq` query since I was on it, and the change was not entirely unrelated to this PR
- I reviewed the entire obo-dashboard log, and the main (I think the only!) real issue is the `multiple_equivalent_classes.rq` check. So this is by far the most important check.
- I tested all the changes in an OBO triple store, as I am currently without laptop and cant build robot. Sorry.

@jamesaoverton if you are happy with the changes, I can write the changelog and update the docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Split equivalent class check [#856]
 - Allow Dublin Core "terms" namespace (`http://purl.org/dc/terms/`) for `description` and `title` properties on ontology. [#741]
 - Allow numbers in lowercase_definition check [#866]
+- Blank nodes in [`report`] are now referred to as `blank node` rather than a random identifier [#873]
 
 ### Fixed
 - Fix printing violations in [`report`] [#823]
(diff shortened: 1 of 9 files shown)


jamesaoverton:
The tests pass (GitHub Actions and locally). The code looks good to me. If @beckyjackson has time, it would be great if she can look at the code -- if not, let me know and we'll just go ahead.

matentzn:
Thank you both, I added docs and comment to changelog.

Issue #829, opened 2021-02-23
Should unparsed triples be a WARN instead of ERROR?


beckyjackson:
Every time we load OBI now, we get the following error message:
```
2021-02-22 18:43:34,590 ERROR org.obolibrary.robot.IOHelper - Input ontology contains 42 triple(s) that could not be parsed:
 - <http://purl.obolibrary.org/obo/OBI_0000958> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494670.
 - <http://purl.obolibrary.org/obo/OBI_0000901> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494669.
 - <http://purl.obolibrary.org/obo/OBI_0000759> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494662.
 - <http://purl.obolibrary.org/obo/OBI_0001355> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494682.
 - <http://purl.obolibrary.org/obo/OBI_0000756> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494659.
 - <http://purl.obolibrary.org/obo/OBI_0000829> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494668.
 - <http://purl.obolibrary.org/obo/OBI_0002903> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494695.
 - <http://purl.obolibrary.org/obo/OBI_0001375> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494684.
 - <http://purl.obolibrary.org/obo/OBI_0001856> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494692.
 - <http://purl.obolibrary.org/obo/OBI_0000769> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494664.
 - <http://purl.obolibrary.org/obo/OBI_0001287> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494675.
 - <http://purl.obolibrary.org/obo/OBI_0000779> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> _:genid2147494667.
[...]

PR #925, opened 2021-10-29
Change println to logging in tests

### LINKED ISSUE #838: Consider logging outputs of unit tests instead of `System.out.println()`
After the incredible clean up in #837, the noisiest part of the output is from the unit tests. I used `System.out.println()` for the tests because it was easy, but now I think we should consider logging the messages properly.

beckyjackson:
Resolves #838

I used the debug level on the logger so it doesn't automatically display them unless you tweak the logging properties, but I think that's OK. The tests still show some error messages from unparsed triples, but that might change to warning with #829.

### DIFF
--- a/robot-core/src/test/java/org/obolibrary/robot/CoreTest.java
+++ b/robot-core/src/test/java/org/obolibrary/robot/CoreTest.java
@@ -11,15 +11,17 @@ import java.net.URISyntaxException;
 import java.util.Set;
 import org.semanticweb.owlapi.apibinding.OWLManager;
 import org.semanticweb.owlapi.model.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Helper methods for core tests. */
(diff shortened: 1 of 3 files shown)


jamesaoverton:
Thanks. I still see a bunch of output in the ` T E S T S` section. There's still a bunch of `println`s, like this one: https://github.com/ontodev/robot/blob/master/robot-core/src/test/java/org/obolibrary/robot/CoreTest.java#L94

Is there any more stuff that we can switch?

beckyjackson:
That link is for master - I changed it to the logger in this branch here: https://github.com/ontodev/robot/blob/e2844ce586b9fdb0b2158a6c9ff9ee3568223221/robot-core/src/test/java/org/obolibrary/robot/CoreTest.java#L96

Sorry if I'm missing something else...

jamesaoverton:
You're right, my mistake.

jamesaoverton (approved):

Issue #924, opened 2021-10-22
SPLIT fails with owl:NamedIndividual


hkir-dev:
In ROBOT version 1.8.1, I have a very simple template:

ID,Entity Type,Synonyms
ID,TYPE,A oboInOwl:hasExactSynonym SPLIT=|
ex:indv_2,owl:NamedIndividual,Sncg_4|Alpha7 cell

robot template --template ./template_fail2.csv --output ./output.owl

output is:
```xml
<NamedIndividual rdf:about="ex:indv_2">
        <oboInOwl:hasExactSynonym rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Sncg_4|Alpha7 cell</oboInOwl:hasExactSynonym>
</NamedIndividual>
```

if I change owl:NamedIndividual to owl:Class, owl:DataProperty etc. or completely delete the column, SPLIT works as expected:
```xml
<Class rdf:about="ex:indv_2">
        <oboInOwl:hasExactSynonym rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Alpha7 cell</oboInOwl:hasExactSynonym>
        <oboInOwl:hasExactSynonym rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Sncg_4</oboInOwl:hasExactSynonym>
</Class>
```

jamesaoverton:
Should be fixed by #924

PR #927, opened 2021-11-08
Fix Issue: SPLIT fails with owl:NamedIndividual #924

hkir-dev:
Resolves [#924]

- [No] `docs/` have been added/updated
- [OK] tests have been added/updated
- [OK] `mvn verify` says all tests pass
- [OK] `mvn site` says all JavaDocs correct
- [OK] `CHANGELOG.md` has been updated

Split command was not working properly with owl:NamedIndividual, since SPLIT definitions were deleted from the template while creating individual axioms. 'If check' implemented to add SPLIT definition back, while adding annotation axioms to the individual.


jamesaoverton (approved):
Much appreciated!

Issue #931, opened 2021-11-16
Add ROBOT domain and range inferred axiom generators


matentzn:
Fixes https://github.com/oborel/obo-relations/issues/140

jamesaoverton:
Closed by #932

PR #932, opened 2021-11-16
Adding inferred axiom gens for domains and ranges

matentzn:
Resolves [#931]

This solves an [RO use case](https://github.com/oborel/obo-relations/issues/140), fixing #931. Seems to be a weird omission from OWL API proper.

Note i am using `direct=false` settings because I can think of any use case for `direct=true` top of my head. Correct me if I am wrong.


robot-core/src/main/java/org/obolibrary/robot/reason/InferredObjectPropertyRangeAxiomGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+      @Nonnull Set<OWLObjectPropertyRangeAxiom> result,
+      @Nonnull Set<OWLObjectPropertyExpression> nonSimpleProperties) {
+    for (OWLClass range : reasoner.getObjectPropertyRanges(entity, false).getFlattened()) {
+      result.add(dataFactory.getOWLObjectPropertyRangeAxiom(entity, range));
+    }
+  }
+
+  @Override
+  public String getLabel() {
+    return "Inferred Object Property Range Axioms";
```
balhoff, line 30: Capitalization differs from InferredObjectPropertyDomainAxiomGenerator :-)

balhoff (approved):

balhoff:
> Note i am using direct=false settings because I can think of any use case for direct=true top of my head. Correct me if I am wrong.

@matentzn what does indirect mean in this situation? Given class hierarchy A SubClassOf B SubClassOf C SubClassOf D, if the domain is A, I don't think we want to assert also B, C, and D. Is that what would happen?

matentzn:
I was assuming that indirect means, if Domain(R, A) and B sub A, then we also get Domain(R, B) - you think this is not the case?

Hmm here are the docs:
https://owlcs.github.io/owlapi/apidocs_4/org/semanticweb/owlapi/reasoner/OWLReasoner.html#getObjectPropertyRanges-org.semanticweb.owlapi.model.OWLObjectPropertyExpression-boolean-

Its a bit.. strangely written.

```
Let N = getEquivalentClasses(ObjectSomeValuesFrom(ObjectInverseOf(pe) owl:Thing)) .
If direct is true: then if N is not empty then the return value is N, else the return value is the result of getSuperClasses(ObjectSomeValuesFrom(ObjectInverseOf(pe) owl:Thing), true) .
If direct is false: then the result of getSuperClasses(ObjectSomeValuesFrom(ObjectInverseOf(pe) owl:Thing), false) together with N if N is non-empty.
```

Weirdly enough, nothing is generated when trying to apply this to RO though... Something must be going wrong.

balhoff:
Which reasoner are you using?

matentzn:
Thank you @balhoff 

You were right about indirect vs direct, changing it:

image


Do you understand why this branch is failing checks?

balhoff:
This is the binary compatibility check, I think. I don't know exactly what it looks for. It seems like adding a class should be backwards compatible. @jamesaoverton can you comment?

robot-core/src/main/java/org/obolibrary/robot/reason/InferredObjectPropertyDomainAxiomGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+    extends InferredObjectPropertyAxiomGenerator<OWLObjectPropertyDomainAxiom> {
+
+  @Override
+  protected void addAxioms(
+      @Nonnull OWLObjectProperty entity,
+      @Nonnull OWLReasoner reasoner,
+      @Nonnull OWLDataFactory dataFactory,
+      @Nonnull Set<OWLObjectPropertyDomainAxiom> result,
+      @Nonnull Set<OWLObjectPropertyExpression> nonSimpleProperties) {
+    for (OWLClass domain : reasoner.getObjectPropertyRanges(entity, true).getFlattened()) {
```
balhoff, line 23: Getting ranges here instead of domains.

balhoff (changes requested):

matentzn:
There is a bit too much of:


```
<rdf:Description rdf:about="http://purl.obolibrary.org/obo/RO_0001001">
        <rdfs:domain rdf:resource="http://www.w3.org/2002/07/owl#Thing"/>
        <rdfs:range rdf:resource="http://www.w3.org/2002/07/owl#Thing"/>
        <oboInOwl:inSubset rdf:resource="http://purl.obolibrary.org/obo/ro/subsets#ro-eco"/>
    </rdf:Description>
```

being inferred. Does everyone agree that I can ignore owl:Thing as range altogether?

balhoff:
> Does everyone agree that I can ignore owl:Thing as range altogether?

Absolutely. Or else make it interact with the tautology options.

matentzn:
I think this just means - adding them to the results, filtering them out, testing them with the tautology checker.. I doubt all this extra code and processing is worth it.. I suggest I just ignore it to begin with.

robot-core/src/main/java/org/obolibrary/robot/reason/InferredObjectPropertyDomainAxiomGenerator.java
matentzn: oops. thanks

robot-core/src/main/java/org/obolibrary/robot/reason/InferredObjectPropertyDomainAxiomGenerator.java
```diff
@@ -20,13 +20,15 @@ protected void addAxioms(
       @Nonnull OWLDataFactory dataFactory,
       @Nonnull Set<OWLObjectPropertyDomainAxiom> result,
       @Nonnull Set<OWLObjectPropertyExpression> nonSimpleProperties) {
-    for (OWLClass domain : reasoner.getObjectPropertyRanges(entity, true).getFlattened()) {
-      result.add(dataFactory.getOWLObjectPropertyDomainAxiom(entity, domain));
+    for (OWLClass domain : reasoner.getObjectPropertyDomains(entity, true).getFlattened()) {
+      if (!domain.equals(dataFactory.getOWLThing())) {
```
balhoff, line 24: Nicer: `!domain.isOWLThing()`

robot-core/src/main/java/org/obolibrary/robot/reason/InferredObjectPropertyRangeAxiomGenerator.java
```diff
@@ -21,12 +21,14 @@ protected void addAxioms(
       @Nonnull Set<OWLObjectPropertyRangeAxiom> result,
       @Nonnull Set<OWLObjectPropertyExpression> nonSimpleProperties) {
     for (OWLClass range : reasoner.getObjectPropertyRanges(entity, true).getFlattened()) {
-      result.add(dataFactory.getOWLObjectPropertyRangeAxiom(entity, range));
+      if (!range.equals(dataFactory.getOWLThing())) {
```
balhoff, line 24: Nicer: `!range.isOWLThing()`

balhoff (changes requested):

robot-core/src/main/java/org/obolibrary/robot/reason/InferredObjectPropertyRangeAxiomGenerator.java
matentzn: I agree, done.

robot-core/src/main/java/org/obolibrary/robot/reason/InferredObjectPropertyDomainAxiomGenerator.java
matentzn: I agree, done :) my owlapi has become rusty.

balhoff (approved):

matentzn:
I can confirm it works:

image


Now we need someone to actually fix this error.. I googled for the docs, but cant find an actual explanation what this error means.. In any case, my work I think is done here :) Thank you @balhoff 

Maybe @jamesaoverton we should slowly start thinking about a new robot release as well.

jamesaoverton:
Yes, we can make a new release once we land this. I'll take a look tomorrow. @beckyjackson can you please review?

jamesaoverton:
> This is the binary compatibility check, I think. I don't know exactly what it looks for. It seems like adding a class should be backwards compatible. @jamesaoverton can you comment?

Yeah, I don't see why this would fail the japicmp check. It seems completely backwards compatible to me.

jamesaoverton:
@matentzn Could we have some sort of test for this feature?

jamesaoverton (approved):

robot-core/pom.xml
```diff
@@ -31,44 +31,6 @@
           </execution>
         </executions>
       </plugin>
-
-      <!-- Ensure backward compatibility -->
-      <plugin>
```
beckyjackson, line 36: Why were these lines removed?

docs/reason.md
```diff
@@ -116,6 +116,15 @@ robot reason --input unreasoned.owl
   --output reasoned.owl
 ```
 
+Note that for some of the axiom generators you may need the Hermit reasoner
+to get meaningful results:
+
+```
```
beckyjackson, line 122: In order for this to be included in the integration test, it should use the `    ` (four spaces) code format, not the backticks, otherwise it gets skipped.

beckyjackson (changes requested):
This looks good to me, other than some minor comments that I left in-line. Please add the `ObjectPropertyDomain` and `ObjectPropertyRange` generators to the list of axiom generators in the `reason.md` docs (https://github.com/ontodev/robot/blob/794c54f55b19efd0f99c23f0d4169cfef5a8fa04/docs/reason.md#generated-axioms).

robot-core/pom.xml
jamesaoverton: The PR was failing the japicmp check, for no good reason. I removed the check temporarily to ensure that all other tests are passing.

docs/reason.md
matentzn: Ooops I guess I was thrown because some of the code samples in reason.md actually use the triple-tick format, probably in order to not be run as part of the test suite.

beckyjackson (approved):

PR #948, opened 2021-12-10
Replace slf4j-log4j12 with logback

jamesaoverton:

ROBOT uses `slf4j-log4j12 1.7.30` for logging, which pulls in `log4j 1.2.17`. Log4j 1.2.17 is no longer supported, and has known vulnerabilities: <https://logging.apache.org/log4j/1.2/index.html>. I would like to try replacing it with `logback-classic 1.2.7`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,9 +196,9 @@
       <version>4.13.1</version>
     </dependency>
     <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-log4j12</artifactId>
-      <version>1.7.30</version>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <version>1.2.7</version>
     </dependency>
     <dependency>
       <groupId>org.semanticweb.elk</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -201,10 +201,27 @@
       <version>1.2.7</version>
     </dependency>
     <dependency>
+      <groupId>org.slf4j</groupId>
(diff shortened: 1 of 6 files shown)


jamesaoverton:
@beckyjackson This passes tests but since there's no logback configuration it's very noisy. We need:

1. logback configuration eqivalent to <https://github.com/ontodev/robot/blob/master/robot-core/src/main/resources/log4j.properties> and <https://github.com/ontodev/robot/blob/master/robot-command/src/main/resources/log4j.properties>
2. update the `-vvv` handling here <https://github.com/ontodev/robot/blob/master/robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java#L948>

Can you please try to make this work?

beckyjackson:
Those two items are working now, but we can't remove those old `log4j.properties` files. If we do, we get this message from ELK:
```
log4j:WARN No appenders could be found for logger (org.semanticweb.elk.config.ConfigurationFactory).
log4j:WARN Please initialize the log4j system properly.
```

jamesaoverton:
@beckyjackson Can you please try excluding the `log4j` from the `elk` import, and importing `org.slf4j:log4j-over-slf4j:1.7.32` instead? I'm hoping that will replace `log4j` in `elk` with some `slf4j` classes, which will then use `logback`.

jamesaoverton:
@ontodev/robot-team Hi everyone. Would dropping `log4j` in favour of `logback` like this cause problems for anyone? I would really like to merge this, so please don't hesitate to let us know!

PR #953, opened 2021-12-16
Exclude log4j from elk import, drop redundant logback.xml

jamesaoverton:
#948 did not actually exclude `log4j` from `elk` when I checked with `mvn dependency:tree`.

The maven-shade plugin also warned about duplicate `logback.xml` files in `robot-command` and `robot-core`.


beckyjackson (approved):
Works fine for me. Sorry, I should have checked the dependency tree!

PR #960, opened 2022-01-13
Add --rename option to rename

beckyjackson:

Right now, `rename` only accepts a file of mappings. If you're only renaming one or two IRIs, this can be a little inconvenient. I added a `--rename` option which accepts two arguments, the original IRI and the new IRI.


docs/rename.md
```diff
@@ -51,6 +58,13 @@ Old IRI,New IRI
 obo:BFO_0000051,fb:BFO_1234567
 ```
 
+If you also want to update the `rdfs:label` of the term you replaced, you can add a third column with the new label value:
```
matentzn, line 61: Perhaps add a sentence what happens to the original label (I assume _all_ existing rdfs:label annotations on the term get deleted) and potential annotations on it (language tags, annotation assertions). Also add a note about _multiple labels_, in case someone curates rdfs:label in multiple languages.

robot-command/src/main/java/org/obolibrary/robot/RenameCommand.java
```diff
@@ -151,17 +166,39 @@ public CommandState execute(CommandState state, String[] args) throws Exception
...
+          String newIRI = map.getValue().get(0);
+          String newLabel = map.getValue().get(1);
+          mappings.put(map.getKey(), newIRI);
+          if (newLabel != null) {
+            labels.put(ioHelper.createIRI(newIRI), newLabel);
+          }
+        }
+      }
+      // mappings from individual args
+      if (singleMappings != null) {
```
matentzn, line 188: arent the singleMappings and fullFile parameters mutually exclusive? Maybe better have them in separate conditionals to avoid some unforeseen errors (I am thinking of filenames with spaces and other menaces).

robot-command/src/main/java/org/obolibrary/robot/RenameCommand.java
```diff
@@ -216,12 +254,19 @@ private static char getSeparator(String fileName) throws IOException {
       reader.readNext();
       for (String[] nextLine : reader) {
         lineNum++;
-        if (nextLine.length != 2) {
+        if (nextLine.length < 2 || nextLine.length > 3) {
```
matentzn, line 257: Maybe nit-picky, but maybe wouldnt it be safer to determine the number of columns in row 1 and then ensuring any subsequent row to be of the same length? Not sure.

matentzn (commented):
Very happy about this feature. Thanks!

robot-command/src/main/java/org/obolibrary/robot/RenameCommand.java
beckyjackson: I didn't think they needed to be, if you had a file and wanted to add an extra mapping or two to it. @jamesaoverton what do you think?

beckyjackson:
Thanks for your review @matentzn ! I updated the RDFS label docs and added some code to validate the header line first.

robot-command/src/main/java/org/obolibrary/robot/RenameCommand.java
matentzn: Now that I think of it, I actually have a use case for this case:

```
--mappings a b --mappings some.file.tsv --mappings c d
```

i.e. allowing the `--mappings` parameter to be properly n-ary with a mix of mapping files and hard coded mappings. This happens when you have a process that dynamically generates a mapping file for some specific set of terms, but you have a few hard coded mappings you want to change every time. Right now, I basically have multiple mapping tables, and then `cat` them together, which is fine. I just wanted to mention this in case you think others may be faced with the same kind of issue. (sorry this is the opposite of my first comment)

beckyjackson: The single mapping is `--mapping` while the file input is `--mappings` (plural). Is this too similar and confusing?

matentzn: ah no, no problem. Sorry didnt see that. No, if --mapping is n-ary (multiple or whatever that is called in CLI speech) its all good!

PR #964, opened 2022-01-14
Implementation of expand command.

### LINKED ISSUE #963: new command: `expand`
I would like to add an `expand` command to ROBOT. This would be used to apply "macro" expansions to generate axioms, primarily from annotation property assertions, although this can be pretty open-ended. Initially `expand` will support applying SPARQL CONSTRUCT queries to generate content. The command could be later extended with options for other systems. I will request an annotation property in OMO, which would link other annotation properties to SPARQL CONSTRUCT literal strings. Example:

```turtle
RO:never_in_taxon OMO:defined_by_construct """
   CONSTRUCT { 
      ?x owl:disjointWith [ 
         a owl:Restriction; 
         owl:onProperty RO:in_taxon; 
         owl:someValuesFrom ?y 
      ]
   } WHERE { 
      ?x RO:never_in_taxon ?y 
   }""" .
```

Running expand would apply all such CONSTRUCT values found in the imports chain to the current ontology and imports.

balhoff:
Resolves #963.

Implement `expand` for #963.


balhoff:
This is complete; however it depends on https://github.com/information-artifact-ontology/ontology-metadata/issues/85.

balhoff:
Done.

docs/expand.md
```diff
@@ -0,0 +1,82 @@
...
+
+By default, all SPARQL CONSTRUCT queries found as values for the [defined by construct](http://purl.obolibrary.org/obo/OMO_defined_by_construct) property
+will be executed against the input ontology, and the results merged in.
+
+Alternatively, specific macro terms can be included or excluded from the expansion process:
+
+    robot expand \
+        --input unexpanded.ttl \
+        --expand-term 'http://example.org/never_part_of' \
+        --no-expand-term 'http://example.org/part_disjoint_with' \
```
matentzn, line 69: Not sure I understand this example: omitting `expand-term` will default to expanding all terms, right. So adding a specific macro with `expand-term` should IMO restrict the set of expanded terms to just that. So what would be the point to having a `no-expand-term` follow an `expand-term`?

docs/expand.md
```diff
@@ -0,0 +1,82 @@
...
+## Specifying expansions
+
+By default, all SPARQL CONSTRUCT queries found as values for the [defined by construct](http://purl.obolibrary.org/obo/OMO_defined_by_construct) property
+will be executed against the input ontology, and the results merged in.
+
+Alternatively, specific macro terms can be included or excluded from the expansion process:
+
+    robot expand \
+        --input unexpanded.ttl \
+        --expand-term 'http://example.org/never_part_of' \
```
matentzn, line 68: Can I expect --expand-term options to be executed in order of how we specify them in the command? What is the default order if no expand-term are given?

balhoff: Order is irrelevant. The results are not allowed to interact.

matentzn: Ok, so if I have expansions that interact, I can always do 

`expand --expand-term A expand --expand-term B`

Thanks! Mayne add a comment to that end to the docs (that order is irrelevant).

balhoff: Okay, I did that.

robot-core/src/main/java/org/obolibrary/robot/ExpandOperation.java
```diff
@@ -0,0 +1,150 @@
...
+                ax ->
+                    executeConstruct(
+                        dataset,
+                        ax.getValue().asLiteral().orNull().getLiteral(),
+                        (IRI) (ax.getSubject()),
+                        config.annotateExpansionAxioms()))
+            .reduce(ExpandOperation::combine)
+            .orElse(Collections.emptySet());
+    if (config.createNew) {
+      ontology.getOWLOntologyManager().removeAxioms(ontology, ontology.getAxioms());
```
matentzn, line 113: Lol, I didnt know this is how we do this.. I would have thought an entirely new ontology, but, why not. I can see this helps with piping. Sorry ignore comment just noise

balhoff: I followed the `reason` implementation for that.

matentzn (commented):
Very nice, especially the "expand()" method itself is very elegantly written. I approve, but I don't officially approve because of the blocking issue.

docs/expand.md
jamesaoverton: I haven't had a chance to review yet, sorry. Here's a thought (not a blocker or a requirement):

Because of my short-sighted design decision years ago, ROBOT only passes an `OWLOntology` between chained commands. So for a chain of expand commands like this, we are repeatedly calling `loadOntologyAsDataset()`. That's inefficient. If that's an important use case to support, or if we're worried about handling very large ontologies, maybe we could borrow from <http://robot.obolibrary.org/query#executing-on-disk>. We could add support for this later as long as there's room in the design.

So right now I'm just asking a question: Is there room to add TDB options later?

balhoff:
I've updated this PR to use the new OMO annotation property added for this command: https://github.com/information-artifact-ontology/ontology-metadata/pull/91

jamesaoverton:
@balhoff Glad to see this moving forward. @beckyjackson please review this PR.

matentzn (approved):
Aside from the question of prov:wasDerivedFrom is the correct property looks nice.

docs/expand.md
beckyjackson: @jamesaoverton it would be really easy to add a TDB option - you would just need to use this (maybe this should be moved to `QueryOperation` if we're going to reuse it.): https://github.com/ontodev/robot/blob/00ad6b8bf901757c3548ae18aa82ddcf8623a815/robot-command/src/main/java/org/obolibrary/robot/QueryCommand.java#L198

... instead this: 
https://github.com/ontodev/robot/blob/2efe4614357eb90c41b45abd92b2802359b1d020/robot-core/src/main/java/org/obolibrary/robot/ExpandOperation.java#L93

... and then you can execute everything on the `Dataset` the same way.

robot-command/src/main/java/org/obolibrary/robot/ExpandCommand.java
```diff
@@ -0,0 +1,128 @@
...
+      return null;
+    }
+    if (state == null) {
+      state = new CommandState();
+    }
+    IOHelper ioHelper = CommandLineHelper.getIOHelper(line);
+    state = CommandLineHelper.updateInputOntology(ioHelper, state, line);
+    OWLOntology ontology = state.getOntology();
+    // Override default expand options with command-line options
+    ExpandOperation.ExpandConfig config = new ExpandOperation.ExpandConfig();
```
beckyjackson, line 112: For other commands, we've done a `getDefaultOptions()`:
https://github.com/ontodev/robot/blob/2efe4614357eb90c41b45abd92b2802359b1d020/robot-core/src/main/java/org/obolibrary/robot/ExtractOperation.java#L49-L58

Then override them:
https://github.com/ontodev/robot/blob/2efe4614357eb90c41b45abd92b2802359b1d020/robot-command/src/main/java/org/obolibrary/robot/ExtractCommand.java#L163-L168

I don't see a problem with this method, but I'm wondering if - for consistency's sake - we should standardize how we handle default options? We've never really made this a "rule" in the past. @jamesaoverton

beckyjackson (commented):
Looks good! I just had one question about the default options.

robot-command/src/main/java/org/obolibrary/robot/ExpandCommand.java
balhoff: One of my goals there was to be a bit more type-safe vs. map of strings. But the object could be returned from a method `getDefaultOptions` if desired.

balhoff:
@jamesaoverton @beckyjackson anything else before merging this?

jamesaoverton:
@balhoff I don't think so. I'm working through the PR list now.

jamesaoverton:
@balhoff `expand` is using `prov:wasDerivedFrom` to point to an AnnotationProperty with an `OMO:0002000` 'defined by construct' annotation. That's fine, but how do you want this to interact with #977 provenance on axioms in `extract` and `merge`? The current version of #977 will not "overload" the `prov:wasDerivedFrom` axiom annotation. I think you would want both axiom annotations, but then one points to an AnnotationProperty and the other to a versionIRI.

jamesaoverton (approved):

balhoff:
> @balhoff `expand` is using `prov:wasDerivedFrom` to point to an AnnotationProperty with an `OMO:0002000` 'defined by construct' annotation. That's fine, but how do you want this to interact with #977 provenance on axioms in `extract` and `merge`? The current version of #977 will not "overload" the `prov:wasDerivedFrom` axiom annotation. I think you would want both axiom annotations, but then one points to an AnnotationProperty and the other to a versionIRI.

@jamesaoverton that's a good point. I'm tempted to change the default annotation property here to just `dct:source`. What do you think?

jamesaoverton:
Using `dct:source` is fine by me. Do you want to consult more widely on that decision before I merge this?

matentzn:
Gimme one day I am out for beer. Wanna think one sec about the best property maybe it's that one .

matentzn:
Ok, ideally we would use `prov:wasGeneratedBy o prov:used`, but since there is not really a shortcut property for that (and we don't want to start layer expressive prov provenance here on top of the ontology) I agree dc:source is as ok as it gets. Maybe one day we really do layer a proper provenance model on top of our axioms :)

BTW in Mondo we use dc:conformsTo to link an axiom with a pattern, which is sort of true here as well, but because we use the template to actually generate the triple, I think it may be confusing. dc:source is ok.

PR #973, opened 2022-02-24
Fix creation of equivalent class axioms from template

beckyjackson:

Currently, template creates an intersection for all equivalent class assertions, even if there is only one equivalent expression. Intersections should only be used for 2+ expressions. See: https://github.com/obi-ontology/obi/issues/1500

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/Template.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/Template.java
@@ -977,15 +977,20 @@ public class Template {
       throws Exception {
     Set<OWLAnnotation> axiomAnnotations = new HashSet<>();
     Set<OWLClassExpression> expressions = new HashSet<>();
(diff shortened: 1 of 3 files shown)


beckyjackson:
Note the difference between using `EC` and using `C` with an "equivalent" class type, from the docs:
- `EC %`: the class will be asserted to be an equivalent class of the intersection of *all* `EC` class expressions in a row
- `equivalent`: values of all `C` columns will be taken as an intersection and asserted to be an equivalent class

This is maintained in this PR (hence the new `legacy-template.owl` which keeps the intersection), but a single equivalent expression will still never be an intersection. This is to maintain backwards compatibility so users don't see any unexpected changes in their axioms. Essentially, using `EC` will create a single equivalent axiom whereas using `C` will create equivalencies between multiple expressions.

This shows up the same in Protege, but the axioms look a bit different in OWL.

Issue #978, opened 2022-03-15
QueryOperation.convertModel causes heap exhaustion for large ontologies.


psiotwo:
`QueryOperation.convertModel` uses byte array transfer to create an updated OWL ontology. If the input model is large, the method requires approx. twice as much heap (for input model and output ontology). 

To decrease the memory footprint, another type of stream (or conversion through a temporary file) would be beneficial.

psiotwo:
Resolved via https://github.com/ontodev/robot/pull/980.

PR #980, opened 2022-03-22
[#978] lowering memory consumption of convertModel.

psiotwo:
Resolves [#978]

Lowers memory consumption as per #978 for convertModel by saving it do disk. Without this, both the input and output model is kept in memory.

### DIFF
--- a/robot-command/src/main/java/org/obolibrary/robot/QueryCommand.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/QueryCommand.java
@@ -55,7 +55,7 @@ public class QueryCommand implements Command {
     o.addOption("O", "output-dir", true, "Directory for output");
     o.addOption("g", "use-graphs", true, "if true, load imports as named graphs");
     o.addOption("u", "update", true, "run a SPARQL UPDATE");
-    o.addOption("y","temporary-file", true, "(together with --update only) true reduces the heap memory used by storing intermediate results into a temporary file, but makes the execution slower.");
+    o.addOption("y","temporary-file", true, "if true, stores intermediate --update results in a temporary file to reduce heap memory usage.");
     o.addOption("t", "tdb", true, "if true, load RDF/XML or TTL onto disk");
     o.addOption("C", "create-tdb", true, "if true, create a TDB directory without querying");
     o.addOption("k", "keep-tdb-mappings", true, "if true, do not remove the TDB directory");
--- a/robot-command/src/main/java/org/obolibrary/robot/QueryCommand.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/QueryCommand.java
@@ -328,9 +328,7 @@ public class QueryCommand implements Command {
         catalogPath = null;
       }
     }
-    return useTemporaryFile
-        ? QueryOperation.convertModelThroughTemporaryFile(model, ioHelper, catalogPath)
-        : QueryOperation.convertModel(model, ioHelper, catalogPath);
+    return QueryOperation.convertModel(model, ioHelper, catalogPath, useTemporaryFile);
   }
 
   /**
--- a/robot-command/src/main/java/org/obolibrary/robot/QueryCommand.java
+++ b/robot-command/src/main/java/org/obolibrary/robot/QueryCommand.java
@@ -55,6 +55,7 @@ public class QueryCommand implements Command {
     o.addOption("O", "output-dir", true, "Directory for output");
     o.addOption("g", "use-graphs", true, "if true, load imports as named graphs");
     o.addOption("u", "update", true, "run a SPARQL UPDATE");
+    o.addOption("y","temporary-file", true, "(together with --update only) true reduces the heap memory used by storing intermediate results into a temporary file, but makes the execution slower.");
     o.addOption("t", "tdb", true, "if true, load RDF/XML or TTL onto disk");
     o.addOption("C", "create-tdb", true, "if true, create a TDB directory without querying");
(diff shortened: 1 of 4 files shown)


matentzn:
This looks good, but it introduces a new paradigm to ROBOT that has never been done before: the temporary file. Given the distinct risk of another file of that name being present, if a temporary file is created, we should follow the conventions of UNIX tools like `sed` and create something like a hashcode. Also, it will not be necessary to add a file extension, as OWLAPI will try all parsers anyways. 

However, independent of the question if @jamesaoverton likes the idea of temporary files (I am fine with it, and indeed, this PR could solve some key issues for me as well), I am wondering a bit about the benchmarking here. Yes, you drastically reduce the heap, which is great, but how much overhead to you introduce by adding a read/write in the middle of a chain? So your PR is going to be very interesting for resource constrained environments like GitHub actions, but if you have a massive amount of memory at your disposal, and you want to run an update query, then you just increased the duration of this process by an unknown factor (IO).

Can you provide some benchmarking details with a huge ontology that contrasts memory consumption with increased runtime for large ontologies?

matentzn:
It is probably more in the spirit of robot to

- add a new command line parameter (`--temporary-file`)
- Add a switch that allows the user to chose between the two options

jamesaoverton:
Thanks for the PR @psiotwo, and for the feedback @matentzn. I'd would also like to know what the performance differences are.

We're pretty strict about backwards compatibility in ROBOT, so I think we would want a new option like @matentzn suggested: `--temporary-file foo.owl` or something, or `--temporary-file true` and then use `createTempFile()` ([example](https://github.com/ontodev/robot/blob/5689f6f4db0efc3e85a5180f2e1270e370e8ba9b/robot-core/src/test/java/org/obolibrary/robot/IOHelperTest.java#L268)).

`robot query` does already include the `--tdb` and `--tdb-directory` options: <http://robot.obolibrary.org/query#executing-on-disk>. It might make sense to use a temporary file when also using `--tdb`, but I don't think that these necessarily go together in all cases.

psiotwo:
> This looks good, but it introduces a new paradigm to ROBOT that has never been done before: the temporary file. Given the distinct risk of another file of that name being present, if a temporary file is created, we should follow the conventions of UNIX tools like `sed` and create something like a hashcode as filename (not `robot.owl`). Also, it will not be necessary to add a file extension, as OWLAPI will try all parsers anyways.

The risk of another file of that name being present should not be an issue, as the arguments to Files.createTempFile are just prefix/postfix of the file - [a unique filename will be generated anyway](https://docs.oracle.com/javase/8/docs/api/java/io/File.html). As the file is temporary and is deleted right after the operation, I am not sure whether the name/extension would do any harm. If for any reason the app crashes and the file remains in temp dir, it might be worth to keep some informative prefix / suffix.

robot-core/src/main/java/org/obolibrary/robot/QueryOperation.java
```diff
@@ -254,6 +254,25 @@ public static OWLOntology convertModel(Model model, IOHelper ioHelper, String ca
...
+   * Given a Model, an IOHelper, and a path to an XML catalog, convert the model to an OWLOntology
+   * object.
+   *
+   * @param model Model to convert to OWLOntology
+   * @param ioHelper IOHelper to load ontology
+   * @param catalogPath String path to XML catalog
+   * @return OWLOntology object version of model
+   * @throws IOException on issue loading ontology
+   */
+  public static OWLOntology convertModelThroughTemporaryFile(Model model, IOHelper ioHelper, String catalogPath) throws IOException {
```
beckyjackson, line 267: Instead of `convertModelThroughTemporaryFile`, could we overload like so:
```java
public static OWLOntology convertModel(Model model, IOHelper ioHelper, String catalogPath) {
    return convertModel(model, ioHelper, catalogPath, false);
}

public static OWLOntology convertModel(Model model, IOHelper ioHelper, String catalogPath, boolean useTemporaryFile) {
    if (useTemporaryFile) {
        ...
    } else {
        ...
    }
}
```
I think it better matches previous patterns, and we can just call it like `convertModel(model, ioHelper, catalogPath, useTemporaryFile)` from `QueryCommand.java`.

robot-command/src/main/java/org/obolibrary/robot/QueryCommand.java
```diff
@@ -55,6 +55,7 @@ public QueryCommand() {
     o.addOption("O", "output-dir", true, "Directory for output");
     o.addOption("g", "use-graphs", true, "if true, load imports as named graphs");
     o.addOption("u", "update", true, "run a SPARQL UPDATE");
+    o.addOption("y","temporary-file", true, "(together with --update only) true reduces the heap memory used by storing intermediate results into a temporary file, but makes the execution slower.");
```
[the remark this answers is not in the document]
psiotwo: OK, moved the longer explanation to the QueryOperation.

robot-core/src/main/java/org/obolibrary/robot/QueryOperation.java
psiotwo: OK, updated.

psiotwo: @beckyjackson 🤦 sorry ...

beckyjackson: No problem! Thanks for making this change. 😄

beckyjackson (approved):

PR #982, opened 2022-03-31
update mapping examples in rename docs

reality:
replace commas in example mapping files with the required tabs

### DIFF
--- a/docs/rename.md
+++ b/docs/rename.md
@@ -46,16 +46,16 @@ The mappings for renaming should be specified with the `--mappings` (for full re
 
 For a full rename (you can use prefixes as long as they are defined by the defaults, `--prefix`, or `--add-prefix`):
 
-```
-Old IRI,New IRI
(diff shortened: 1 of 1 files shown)


reality:
When I tried it with a comma I got a message about expecting two columns but getting only one. The problem here, I think, is that I copied the command examples from this document, so my filename had a ".tsv." suffix, but was using commas as delimiters (per the example). I assume robot is using the filename suffix to determine the format? For consistency we could either accept this change, or instead change the command examples to use .csv filenames?

reality:
(the latter may be better since I think copying tabs from github can be problematic sometimes)

jamesaoverton:
Now I understand, and I agree: the format of the example table should match the file extension in the command-line example. In this case, I think making both CSV would be better, for the reason you point to about copy-pasting tabs.

jamesaoverton:
I changed my mind: Let's consistenly use TSV here.

Issue #997, opened 2022-05-04
In some environments, robot-core no longer builds using (openjdk) Java 11


kltm:
`robot` seems to no longer compile in some environments.

Clean and check environment:
```
sjcarbon@moiraine:~/local/src/git/robot[master]$:) java -version && javac -version
openjdk version "11.0.15" 2022-04-19
OpenJDK Runtime Environment (build 11.0.15+10-Ubuntu-0ubuntu0.18.04.1)
OpenJDK 64-Bit Server VM (build 11.0.15+10-Ubuntu-0ubuntu0.18.04.1, mixed mode, sharing)
javac 11.0.15
sjcarbon@moiraine:~/local/src/git/robot[master]$:) git pull
Already up to date.
sjcarbon@moiraine:~/local/src/git/robot[master]$:) mv ~/.m2 ~/.m2.20220504-1
sjcarbon@moiraine:~/local/src/git/robot[master]$:( lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.6 LTS
Release:	18.04
Codename:	bionic
```
Build errors out with:
```
[INFO] Reactor Summary for robot 1.9.0-SNAPSHOT:
[INFO] 
[INFO] robot .............................................. SUCCESS [  3.985 s]
[INFO] robot-maven-plugin ................................. SUCCESS [ 51.406 s]
[INFO] robot-core ......................................... FAILURE [ 37.748 s]
[INFO] robot-command ...................................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:33 min
[INFO] Finished at: 2022-05-04T10:37:39-07:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project robot-core: Compilation failure
[ERROR] /home/sjcarbon/local/src/git/robot/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java:[32,42] cannot access org.geneontology.obographs.core.io.OgJsonGenerator
[ERROR]   bad class file: /home/sjcarbon/.m2/repository/org/geneontology/obographs/obographs-core/0.3.0/obographs-core-0.3.0.jar(org/geneontology/obographs/core/io/OgJsonGenerator.class)
[...]

PR #934, opened 2021-11-18
Update OBOgraphs JSON

matentzn:
Resolves [#933]

- [ ] `docs/` have been added/updated
- [ ] tests have been added/updated
- [ ] `mvn verify` says all tests pass
- [ ] `mvn site` says all JavaDocs correct
- [ ] `CHANGELOG.md` has been updated

a new version of obographs added to ROBOT..

PR #998, opened 2022-05-04
Update Dockerfile to Java 11

jamesaoverton:
Update Dockerfile from Java 8 to Java 11, to support #934. See also #997.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
-FROM maven:3-jdk-8-alpine
+FROM maven:3-openjdk-11-slim
 
-RUN adduser -D robot
+RUN useradd -m robot
 RUN mkdir -p /usr/src/app
 RUN chown robot /usr/src/app
 


jamesaoverton:
This works for me, but I'd appreciate it if somebody could test this: build with `docker build --tag robot .` then run something like `docker run --rm robot --help`.

matentzn:
Oh cool I didnt know robot is maintaining its very own image. Here are the build instructions:

```
docker build --tag ontodev/robot .
```

You can then test like this (snapshot?):
```
docker run --rm -ti ontodev/robot --version
ROBOT version 1.9.0-SNAPSHOT
```

If you want this for local processing you need to bind PWD as a volume:

```
docker run -v $PWD/:/usr/src/app/ --rm -ti ontodev/robot convert -I http://purl.obolibrary.org/obo/bfo.owl  -o bfo.ofn
```

It all works.

jamesaoverton:
The Dockerfile was a user contribution. I'm not sure how much it is used, but I made the smallest changes I could to update to Java 11. Thanks for checking it @matentzn.

Issue #660, opened 2020-03-26
Consider adding `--reasoner elk5`


jamesaoverton:
There are some compatibility problems with ELK 0.5 that I don't fully understand. What about packaging both ELK 0.4.x and ELK 0.5.x in ROBOT? We would keep ELK 0.4.x as `--reasoner elk` and add ELK 0.5.x as `--reasoner elk5`.

The JARs might just conflict too much, but I think we should try.

matentzn:
What kind of compatibility issues are you facing? I would love elk5 as a reasoner alongside elk, but I would be a bit surprised if you could make it work. All the man classes like ElkReasonerFactory and so on I believe still have the same path..

cmungall:
We could of course have two ODK docker containers, one with a branch of robot that is built using Elk5. Of course this is not optimal for many people, but the groups that need to be running tests (uberon, go, ...) are already using odk.

I realize I am the blocker here, I added many high level constraints in various ontologies a while ago, these introduced cryptic incoherencies that are only resolved using Elk5, rolling out Elk5 universally now would cause a lot of problems...

jamesaoverton:
I'm not sure if it's possible, but I think it's worth spending an hour to try this out.

balhoff:
I think we probably need to publish a shaded version that renames the ELK packages to keep them separate.

matentzn:
As a last resort, Chris suggestion is also ok. It would allow us to switch between elk and elk5 without changing our often quite complex makefiles, so it has some benefits as well. I would still like to see James little experiment first. Elk5 Parameter would be nice.

matentzn:
I am wondering if its time now.. The "cryptic incoherencies" for GO and Uberon have been resolved.. Unfortunately we still do not know if there is soon going to be an ELK 0.6 with even more advanced features. 

What are the feeling all around - maybe create for now a branch with ELK 0.5, add it as a `robot-dev` command to ODK, play with it for 6 months and then discontinuing our support for ELK 0.43 altogether?

matentzn:
You can check out the branch and run:

```
mvn clean package
bin/robot reason -I http://purl.obolibrary.org/obo/go/extensions/go-plus.owl --reasoner elk -o test.owl
```

To see whats up. It looks safe, but will require serious testing obviously.

PR #999, opened 2022-05-07
Updating ELK 0.43 to 0.5

### LINKED ISSUE #380: Make a test version of ROBOT with Elk 0.5
Elk 0.5 is not yet released: https://github.com/liveontologies/elk-reasoner/wiki/GettingElk

but it's important that people get to test before it is released, as it is revealing lots of cryptic issues

 * https://github.com/geneontology/go-ontology/issues/16524
 * https://github.com/obophenotype/uberon/issues/1455

matentzn:
Resolves #380 #660 

- [ ] `docs/` have been added/updated (NOT NECESSARY)

This pull request migrates ELK to the 0.5 version. I have tested it in various ODK pipelines, and bundled it with the latest ODK:dev image.


matentzn:
@jamesaoverton and I discussed this and we will move this to a future version of ROBOT, maybe ROBOT 2.0 :P

balhoff:
@matentzn @jamesaoverton I think it's a high priority for us to get ELK 0.5 into ROBOT. I started to look at publishing a version in one of our own orgs, so we wouldn't need to use the snapshot repo (which I think would prevent ROBOT from being published to Maven central). But I found this one already on Maven central: https://search.maven.org/artifact/au.csiro/elk-owlapi4/0.5.0/jar

Can we just use that for the time being?

matentzn:
I would suggest we do this asap:

1. Start version Dev ODK
2. Create an experimental ROBOT Elk 0.5 version of ROBOT and load it into ODK-dev
3. We migrate all our universe to that ELK (esp. Github actions)
4. Maybe even make a proper parallel robot release for people to test
5. Use this for a while. Once all issues in the main OBO ontologies are identified and ironed out, we can schedule a proper ROBOT release.

Unless someone figures out how to have elk 0.4 and 0.5 co-exist, I am a bit worried who will deal with the 1 billion support issues coming from unsats revealed by 0.5.

jamesaoverton:
I never tried to use two versions of the same dependency in a Java project before. Some quick research makes it look like black magic to me, but if someone else has experience then let us know.

I'm OK with creating a ROBOT branch and beta release `robot-1.9.1-elk-0.5.0.jar` (or something) using ELK 0.5. (Most of my projects use HermiT, so I haven't personally seen the problems with this ELK update.)

matentzn:
Maybe I am overthinking this. @jamesaoverton and I have decided to make a quick experiment with OBO Dashboard and see what will happen if we simply swap in ELK 0.5. Basically, we will compare unsats with ELK 0.4.3 and make a decision based on that.

matentzn:
Sorry I forgot it again @jamesaoverton. Was it somehow possible to run an action and obtain a ROBOT jar from a branch using a URL (this makes it easy to inject a DEV version of ROBOT into ODK)?

matentzn:
Ok, assigned this to @anitacaron, but it will have to wait until mid February until after the OntoSummit Dashboard session is finished: https://github.com/OBOFoundry/OBO-Dashboard/issues/82

jamesaoverton:
> Was it somehow possible to run an action and obtain a ROBOT jar from a branch using a URL (this makes it easy to inject a DEV version of ROBOT into ODK)?

Kinda. Our GitHub Actions do create a JAR file that you can download with a browser, BUT the link uses some JavaScript so you don't get a URL that you can just `curl`, and they expire after 90 days (or something). So this GitHub Actions logs and artifacts for this PR have expired.

I'll try to make a pre-release here in the repo.

jamesaoverton:
Here is a pre-release with an attached JAR: https://github.com/ontodev/robot/releases/tag/v1.9.1-elk-0.5.0

I didn't know how to test that it actually includes ELK 0.5.0, so please confirm that before running hours of tests.

bvarner-ebi:
@jamesaoverton, checking in on this PR. Editors are switching over to Protégé 5.6.1 which has ELK 0.5.0. Based on the thread above, it is not clear what the next step is to update ROBOT to 0.5.0 and who needs to do it. Is there something specific someone needs to do to move this forward so Protégé and ROBOT can be aligned? Thanks for your attention.

matentzn:
@bvarner-ebi thanks for pushing, this is on me. I will soon create a ODK dev release with ELK 0.5 for testing

matentzn:
Trouble in Paradise:

ELK 0.5 cant recognise this ontology as incoherent:


```
Prefix: : <http://x.org/>

Ontology: <http://purl.obolibrary.org/obo/test>

Class: C
Class: O
   DisjointWith: C

ObjectProperty: P
  Domain: C
ObjectProperty: Q
  Domain: O
    SubPropertyOf: P
```

Seems that ELK 0.43 was able to for some reason. Replicated this in Protege as well.

robot-core/src/test/java/org/obolibrary/robot/ReasonerHelperTest.java
```diff
@@ -49,7 +49,9 @@ public void testIncoherentRBox()
       } catch (IncoherentRBoxException e) {
         isCaughtException = true;
       }
-      assertTrue(isCaughtException);
+      // To our great disappointment, ELK does not recognise this incoherent R Box
+      // https://github.com/ontodev/robot/pull/999#issuecomment-1666567017
+      // assertTrue(isCaughtException);
```
matentzn, line 54: @jclerman @balhoff to my great suprise, ELK 0.5 does not work with this recently introduced probing trick that @jclerman added. I checked it in Protege. It truly does not recognise it. Waiting for an Elk 0.5 fix is IMO out of the question: https://github.com/liveontologies/elk-reasoner/issues/66

robot-core/src/main/java/org/obolibrary/robot/ReasonerHelper.java
```diff
@@ -46,9 +46,8 @@ public class ReasonerHelper {
   public static Set<OWLObjectProperty> getUnsatisfiableObjectProperties(OWLReasoner reasoner) {
     Set<OWLObjectProperty> unsatObjectProps = new HashSet<>();
 
-    if (reasoner
-        .getPrecomputableInferenceTypes()
-        .contains(InferenceType.OBJECT_PROPERTY_HIERARCHY)) {
+    if (reasoner.getPrecomputableInferenceTypes().contains(InferenceType.OBJECT_PROPERTY_HIERARCHY)
+        && !reasoner.getClass().getName().equals("org.semanticweb.elk.owlapi.ElkReasoner")) {
```
matentzn, line 50: This is necessary because of https://github.com/liveontologies/elk-reasoner/issues/66#issuecomment-1667697830

robot-core/src/test/java/org/obolibrary/robot/ReasonerHelperTest.java
matentzn: Solved, see below.

robot-core/src/main/java/org/obolibrary/robot/ReasonerHelper.java
jclerman: Thanks @matentzn for tracking down the cause and for this solution!  I'm glad that it didn't require reverting the faster inference when using Hermit - that's been a huge boon in my workflow.

balhoff (approved):

matentzn:
Given my new [awesome diff system](https://github.com/matentzn/robot_ultimate_obo_tester), I determined these cases with difference in unsats:

### Differences in the number of unsatisfiable classes comparing Elk 0.4.3 and Elk 0.5 

ontology | 0.4.3 | 0.5
--- | --- | ---
ontoneo | 21 | 23
aism | 0 | 1238
upheno | 60521 | 83201
psdo | 21 | 46
rbo | 0 | 1
nomen | 0 | 1
kisao | 21 | 28
ecocore | 0 | 9

cc @FernandaFarinelli @JCGiron @zachll @DanBerrios @mjy @jonrkarr @pbuttigieg 


I hereby officially make my recommendation that this should be merged - its thoroughly tested, and ready for action.

jamesaoverton:
I don't have a good way to test this, so I'm taking @matentzn at his word.

PR #1000, opened 2022-05-11
new extract command method: subset

### LINKED ISSUE #497: new command: subset (filter option with ability to traverse all edges when preserve-structure is set)
Carried on from #263 

Given:

```
ontology: test
subsetdef: foo "foo"

[Term]
id: X:1
subset: foo

[Term]
id: X:2
relationship: part_of X:1

[Term]
id: X:3
relationship: part_of X:2

[Term]
id: X:4
relationship: part_of X:3

[Term]
id: X:5
is_a: X:4
subset: foo

[Typedef]
id: part_of
is_transitive: true
xref: BFO:0000050
```

Currently we have:

```
$ robot filter -i robot-core/src/test/resources/subset_test.obo --select annotations --prefix "subset: http://purl.obolibrary.org/obo/test#" --select "oboInOwl:inSubset=subset:foo" -o z.obo && cat z.obo
format-version: 1.2
ontology: test

[Term]
id: X:1
subset: foo

[Term]
id: X:5
subset: foo
```

We would like to see a part_of between X:5 and X:1.

This is pretty fundamental for all subset use cases involving GO, any anatomy ontology, CL, ENVO, any ontology that has part-of.

There are two ways this could be implemented. One is an ad-hoc graph walking procedure with rules for edge label composition (this is how the equivalent command in owltools worked).

The other is to use `materialize`. A special purpose case of this code could be written than will be more efficient for this use case, where for every selected class `C` we test for all inferred subclasses `R some C` for all structure preserving relations R.

To preserve backwards compatibility we should have the user specify the list of OPs they want to include. However, this is such a common use case it would be nice to have a command `subset` as originally specified in #263 that is a shorthand for `filter --preserve-structure true --use-all-relations true --select annotations -select "oboInOwl:inSubset=subset:$SUBSET"`

hkir-dev:
Resolves [#497, resolves #497]

Adds new `subset` method to the extract operation: `robot extract --method subset`
Logic of the method is as follows:
1. Materialized Axioms= Relation Graph(O)
2. OM = ADD(O, Materialized Axioms)
3. OF = RobotFilter(T, OM)
4. OR = RobotReduce(OF)

To achieve this [relation-graph](https://github.com/balhoff/relation-graph) dependency added to the ROBOT. Owl-api version increased to `4.5.9`. Following files updated to fix the errors raised after the owl-api version upgrade:
- docs/examples/example.obo
- docs/examples/metrics_all.json
- robot-core/src/main/java/org/obolibrary/robot/metrics/OntologyMetrics.java
- robot-core/src/test/java/org/obolibrary/robot/MeasureOperationTest.java


dosumis:
I think this now has the behaviour we expect, but I'm a bit worried about the efficiency of the approach used and what that might mean for scaling.  

* Huseyin has confirmed to me that the RG step currently materialises *all* inferred existential restrictions, rather than just those on object properties in the seed.  This is potentially a massive number of triples.  Presumably this has some effect on speed, but I'm rather more worried about the memory requirements.  As I understand it, RG can take a list of objectProperties to materialise => a much small number of triples than the current step.  
* I also wonder whether pushing filtering to ROBOT is also less efficient than a simple filter on triples following the RG step.  Might be less of a concern though.

@balhoff - can you comment on this?

robot-core/src/main/java/org/obolibrary/robot/ExtractOperation.java
```diff
@@ -214,6 +237,143 @@ public static OWLOntology extract(
...
+   * @param ontology ontology to materialize.
+   * @param outputIRI IRI of the output ontology.
+   * @return materialized ontology.
+   * @throws OWLOntologyCreationException
+   */
+  private static OWLOntology materialize(OWLOntology ontology, IRI outputIRI)
+      throws OWLOntologyCreationException {
+    Config config = new Config(null, false, true, true, true, true, false);
+    Set<OWLClassAxiom> materializedAxioms =
+        RelationGraphUtil.computeRelationGraph(ontology, new HashSet<IRI>(), config);
```
balhoff, line 280: @hkir-dev you can pass the desired relations to compute here, instead of the empty set.

balhoff:
@dosumis I commented about selecting object properties here: https://github.com/ontodev/robot/pull/1000/files#r871427940

robot-core/src/main/java/org/obolibrary/robot/ExtractOperation.java
hkir-dev: Fixed the code accordingly. Filtered the object properties from the seed terms and passed to the relation-graph.

balhoff:
@hkir-dev I would suggest making a separate PR for the OWL API upgrade.

matentzn:
Here is the PR: https://github.com/ontodev/robot/pull/891/files

Needs to be tested! High priority - I want to try to push this through by the end of the month..

dosumis:
@hkir-dev Where are we with this?  Are there still blockers?

robot-core/src/main/java/org/obolibrary/robot/ExtractOperation.java
matentzn: @balhoff can you resolve this comment if addressed to your satisfaction?

matentzn: Thx

cmungall:
so it looks like a chain of dependencies - this requires a new owlapi version which depends on a new protege (if we want to continue syncing) - is there any way to progress independent of these

balhoff:
Apparently in relation-graph I am using a method added in a newer OWL API compared to the current ROBOT release. I can work around that and make a new RG release. I think we got hung up because RG needs `4.5.9` (which is what I think Protege 5.5 is on) and that was the original change in this PR, but the open PR #1005 is for `4.5.22`, which is the target for the next Protege release.

hkir-dev:
Protege is using Owl-API `4.5.19` in its latest version. I can prepare a new PR to upgrade Robot to this version.

There are some significant improvements in Owl-API `4.5.22`, so we are trying to migrate both Protege and Robot to this version all together. However, we are still working to make a new stable Protege release and seems it will take some more time.

matentzn:
Everything needs to go to 4.5.22 or later no matter what - no need to bother with 4.5.19. Protege should be upgraded.

hkir-dev:
OWL API upgrade merged to this branch.

jamesaoverton:
Thanks @hkir-dev, this looks great. I made some small changes. I'll merge it.

In the future, you are welcome to make branches on this repository rather than your fork. It make life a little easier for me.

PR #1010, opened 2022-06-15
Fix Guava dependency for obographs.

### LINKED ISSUE #1009: Something is not working with the JSON parser
ROBOT 1.8.4: 

```
robot convert -I http://purl.obolibrary.org/obo/pato.owl -f json -o pato.json
```

causes:

```
Exception in thread "main" java.lang.NoSuchMethodError: 'java.util.stream.Collector com.google.common.collect.ImmutableSet.toImmutableSet()'
	at org.geneontology.obographs.owlapi.FromOwl.generateGraph(FromOwl.java:236)
	at org.geneontology.obographs.owlapi.FromOwl.generateGraphDocument(FromOwl.java:63)
	at org.obolibrary.robot.IOHelper.saveOntologyFile(IOHelper.java:1720)
	at org.obolibrary.robot.IOHelper.saveOntology(IOHelper.java:846)
	at org.obolibrary.robot.CommandLineHelper.maybeSaveOutput(CommandLineHelper.java:667)
	at org.obolibrary.robot.ConvertCommand.execute(ConvertCommand.java:141)
	at org.obolibrary.robot.CommandManager.executeCommand(CommandManager.java:244)
	at org.obolibrary.robot.CommandManager.execute(CommandManager.java:188)
	at org.obolibrary.robot.CommandManager.main(CommandManager.java:135)
	at org.obolibrary.robot.CommandLineInterface.main(CommandLineInterface.java:60)
```

matentzn:
Resolves #1009 

The problem was that other dependencies (OWL API-related) determined that the correct Google Guava version is an ancient one (18.0). This caused the JAR packaging to prefer that (probably because it was earlier in the POM file). 

In this PR, we explicitly include obographs-core as per recommendation in #1009 but we also add a fixed import to Guava.

I don't know if this is the right way to do this, I am too far out of Java land now. But I didn't want to search through 20 depencies to add exclusions. If there is a better way, great!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,11 @@
 
   <dependencies>
     <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>28.2-jre</version>
+    </dependency>
+    <dependency>
       <groupId>net.sourceforge.owlapi</groupId>
       <artifactId>owlapi-api</artifactId>
       <version>4.5.6</version>
--- a/robot-core/pom.xml
+++ b/robot-core/pom.xml
@@ -218,11 +218,6 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>org.geneontology.obographs</groupId>
-      <artifactId>obographs-core</artifactId>
-      <version>0.3.0</version>
-    </dependency>
-    <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-core</artifactId>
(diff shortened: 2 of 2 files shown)


jamesaoverton:
Forcing `guava` 28.2-jre seems to work. Since OWLAPI is pulling in the older `guava` 18, I added the new one to the root `pom.xml`, although I'm not sure what difference it makes.

I don't think adding `obographs-core` makes a difference so I remove it.

I'm still not at all clear about how the `guava` dependency is getting figured out at all.

PR #1011, opened 2022-06-15
Update docs

matentzn:
Just adding some test to json branch.. if you want it. Else just kill.

### DIFF
--- a/docs/convert.md
+++ b/docs/convert.md
@@ -4,6 +4,11 @@ Ontologies are shared in different formats. The default format used by ROBOT is 
 
     robot convert --input annotated.owl --output results/annotated.obo
 
+As in all other ROBOT commands, ontologies can be converted straight from their web location. In this example we convert PATO (using its PURL) to OBOGraphs JSON:
+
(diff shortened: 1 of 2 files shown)


matentzn:
my other tests looked ok

jamesaoverton:
I'm worried that this will cause many false-negative test failures due to network problems. For example, the first CI test on this branch failed because of a network problem:

https://build.obolibrary.io/job/ontodev/job/robot/job/add-docs-json/1/

This output might not be public: https://build.obolibrary.io/job/ontodev/job/robot/job/add-docs-json/1/execution/node/3/ws/robot-command/target/integration-tests.txt

```
org.semanticweb.owlapi.model.UnloadableImportException: Could not load imported ontology: <http://purl.obolibrary.org/obo/ro/annotations.owl> Cause: purl.obolibrary.org
```

I think that all our tests use local files, although there may be some unintentional exceptions.

matentzn:
Ok, I used an already existing local file instead. This conversion would have previously failed, and now works. Documentation wise, it illustrates the use of the `--format` option, which I think is a good enough reason to sneak it in?

jamesaoverton:
This is a good example to have, and also a good test. Thanks!

Issue #1017, opened 2022-06-24
Report queries not handling empty labels on ERROR level


psiotwo:
Standard set of report queries for the ERROR level does not seem to handle cases when a label is empty "".

There are rules e.g. to detect
- [missing label](http://robot.obolibrary.org/report_queries/missing_label)
- [label whitespace](http://robot.obolibrary.org/report_queries/label_whitespace)

But there does not seem to be check that a label is an empty string "" or an empty lang literal, e.g. ""@en. Should not this be covered on the ERROR level?

Also, this causes confusion in combination with the [duplicate label](http://robot.obolibrary.org/report_queries/duplicate_label) query because it reports all concepts with empty labels as potential duplicates.

psiotwo:
Clear @matentzn - I think this check would make sense because not having a label and having an empty label is often just a matter of inconsistent serialisation of the particular tool and does not bear any semantics (although it could, of course).

So having a check for missing_label not including empty labels seems a bug to me w.r.t. this check. 

But I understand your reasoning and your proposal for missing_label was one option I was considering. I will prepare some PR for this and we can discuss it through there ...

jamesaoverton:
Thanks @psiotwo! I think this will be a good addition, but yes let's ensure that there isn't a big performance hit.

psiotwo:
Actually, is there any established way to test (both unit/performance tests) report queries? I must be overlooking st. in the codebase ...

jamesaoverton:
No, you're not missing anything, there's no code here for that.

@matentzn runs a million ontology projects and the OBO Dashboard, so maybe he can suggest a project that takes 5-10 minutes to run `robot report`, and we can test against that.

matentzn:
Actually, it would not hurt to add a few unit tests here as well. But yeah, we develop a query, make sure it works on a corpus, then push it to robot.

psiotwo:
Closing via #1018 .

PR #1018, opened 2022-06-25
[#1017] updating missing_label query to support empty labels

psiotwo:
Resolves [#1017]

- [ x ] `docs/` have been added/updated
- [ x ] tests have been added/updated
- [ x ] `mvn verify` says all tests pass
- [ x ] `mvn site` says all JavaDocs correct
- [ x ] `CHANGELOG.md` has been updated

Added a condition checking for empty labels (also across languages) "" into the `missing_label.rq`. 

The test code added is more general than needed to allow adding other query tests easily.

I did some ad-hoc performance tests just on my Mac on DOID, EFO and HP - all showing up to 10% performance degradation with the new query. 

@matentzn if you have a more representative benchmark to push it through, would be great!


docs/report_queries/missing_label.md
```diff
@@ -1,10 +1,10 @@
-# Missing Label
-
-**Problem:** An entity does not have a label. This may cause confusion or misuse. Excludes deprecated entities.
-
-**OBO Foundry Principle:** [12 - Naming Conventions](http://www.obofoundry.org/principles/fp-012-naming-conventions.html)
-
-**Solution:** Add a label.
+# # Missing Label
```
matentzn, line 1: Introduction of redundant `#` for the MD file!

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
```diff
@@ -695,20 +699,21 @@ private static Map<String, String> getDefaultQueryStrings(Set<String> rules)
...
+          Files.list(Paths.get(dirURL.toURI())).collect(Collectors.toSet());
+      if (queryFilePaths.size() == 0) {
         throw new IOException(
             "Cannot access report query files. There are no files in the directory.");
       }
       Set<String> defaultRuleNames = new HashSet<>();
-      for (String qPath : queryFilePaths) {
-        String ruleName = qPath.substring(qPath.lastIndexOf("/")).split(".")[0];
+      for (final Path qPath : queryFilePaths) {
+        final String ruleName = qPath.getFileName().toString().split("[.]")[0];
```
matentzn, line 710: These kinds of changes are very hard to review and we prefer in the interest of resources to only do them if we have an issue that cleary explains what is wrong, with agreement all around it absolutely needs changing and then changing it. Are they _absolutely essential_ to this PR?

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
```diff
@@ -841,7 +846,7 @@ private static String getQueryResultOrNull(QuerySolution qs, String var) {
    * @return List of Violations
    * @throws IOException on issue parsing query
    */
-  private static List<Violation> getViolations(
+  public static List<Violation> getViolations(
```
matentzn, line 849: Ideally better not make these changes unless a separate issue explains why they are needed.

robot-core/src/test/resources/1017-empty-label-input.owl
```diff
@@ -0,0 +1,27 @@
+<rdf:RDF
+     xmlns:obo="http://purl.obolibrary.org/obo/"
+     xmlns:owl="http://www.w3.org/2002/07/owl#"
+     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
+     xmlns:xml="http://www.w3.org/XML/1998/namespace">
+    <owl:Ontology rdf:about="http://test.org/test"/>
+
+    <owl:Class rdf:about="http://test.org/test#A"/>
```
matentzn, line 9: excellent test

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
matentzn: Ah I see, you need this for the test to work?

matentzn (commented):
@psiotwo looks great! One (apart from the individual comment):

Is every single change you make here _absolutely_ necessary for the PR? This is a very important criterion for us now, because of the cost of accidentally introducing non obvious bugs.

Thank you for your contribution!

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
jamesaoverton: Thanks @matentzn for asking to keep the PR focused, and sorry if I’m being inconsistent, but this path handling looks like a nice improvement.

docs/report_queries/missing_label.md
jamesaoverton: Yes. Better to use the https://github.com/ontodev/robot/blob/master/docs/Makefile to generate the doc from the query file.

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
psiotwo: yes

robot-core/src/main/java/org/obolibrary/robot/ReportOperation.java
psiotwo: @matentzn you are absolutely right. Actually, I originally wanted to "hang" the test on the existing API of ReportOperation - the changes were needed to make it actually working in tests.

But this would have mean going through profile.txt which would be an extra overhead so I reconsidered and used the getViolations method to keep it focused.

This bit of code belonged to the original story and I accidentally committed. I reverted it back. I also don't like these types of extra changes - doing unnecessary changes when there are almost no tests around is dangerous.

docs/report_queries/missing_label.md
psiotwo: Hm, actually deleting `docs/report_queries/missing_label.md` and (naively) running `make` inside docs does generate these # in the MD file ... if I didn't miss any config, the only thing that jumps into my mind now is that some sed regex might behave differently on various platforms (I am using M1 Mac) ... but not sure, will take a look later.

psiotwo: I have just manually removed the #s at present.

jamesaoverton (approved):
Sorry for the long delay. The code looks good, and the detailed test is much appreciated! The performance hit you're reporting is more than I would like, but I feel that this is an important case to cover. Thanks again!

Issue #1016, opened 2022-06-22
Report command fails for some textual content when the output format is JSON


psiotwo:
Having input.owl:

```<?xml version="1.0"?>
<rdf:RDF
     xmlns:obo="http://purl.obolibrary.org/obo/"
     xmlns:owl="http://www.w3.org/2002/07/owl#"
     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
     xmlns:xml="http://www.w3.org/XML/1998/namespace">
    <owl:Ontology rdf:about="http://test.org/test"/>

    <owl:ObjectProperty rdf:about="http://test.org/test#A">
        <obo:IAO_0000115>skos:Concept interferes with &quot;quoted text&quot;</obo:IAO_0000115>
    </owl:ObjectProperty>
</rdf:RDF>
```

and running 

`java -jar robot.jar report --input input.owl --output output.json`

ends up with

```Violations: 5
-----------------
ERROR:      4
WARN:       0
INFO:       1
while parsing a block collection
 in 'reader', line 19, column 9:
            - "skos:Concept interferes with  ... 
            ^
expected <block end>, but found '<scalar>'
 in 'reader', line 19, column 44:
     ... skos:Concept interferes with \\"quoted text\\""
                                         ^

 at [Source: (StringReader); line: 19, column: 44]
Use the -vvv option to show the stack trace.
Use the --help option to see usage information.
```

Looks like some correlation inside the text field - removing the skos: prefix or the escaped double quotes makes things working. Also, textual output format (not JSON) works too.

matentzn:
I have to say @psiotwo your tickets are great :) Thanks for that. Its a bit hard right now to find resources to fix such bugs, but, regardless, your issues are really well reported and helpful! Thanks!

psiotwo:
👍 Sure, clear ... will try to help resolving the stuff, as time permits ... .

psiotwo:
Checked, works as of 1.9.1.

PR #1022, opened 2022-06-28
[#1016] Fixes JSON report serialisation

psiotwo:
Resolves [#1016]

Actually the added tests are maybe more valuable then the resolution itself - since I do not know what was the rationale behind the way the double-quotes in YAML serialisation were escaped before I honestly do not know what I have broken by this change - just hope that not too much :-) ...


jamesaoverton:
Much appreciated. I'm off this week, but I will do my best to catch up when I get back next week.

jamesaoverton:
Again this code looks good, but I'm worried about backwards compatibility. @matentzn Do you use YAML output from ROBOT report in any of your projects?

matentzn:
> Do you use YAML output from ROBOT report in any of your projects?

@jamesaoverton no, I only ever export to TSV or HTML!

Issue #886, opened 2021-07-02
New parameter --axioms "axiom annotations" for remove and filter


matentzn:
This could give us a simple way to remove certain axiom annotations while preserving others.

matentzn:
A use case if, for example, if we have a axiom annotation such as:

```
A subClassOf B [editors_note: "We are not sure this is in the right place", rdfs:seeAlso https://github.com/ontodev/robot/issues/886]
```

When running a release, we may want to remove all the `editors_note` annotations from all our axioms, but not remove them from, say, classes like:

```
A editors_note "this class should be used only after a few shots of Tequila"
```

joeflack4:
+1!

matentzn:
@hkir-dev cannot assign this to you, because you are not on the list of contributors.. @jamesaoverton is it possible to give @hkir-dev enough access to ROBOT so I can tag him on issues?

jamesaoverton:
Ok, I invited @hkir-dev to @ontodev/robot-team

matentzn:
Awesome thank you! @hkir-dev you can assign yourself once you have accepted!

shawntanzk:
spec:

`--axioms axiom_annotations`  -> remove all the axiom annotations (starting point)
However, design might be better to have a new parameter like `--drop_axioms_annotations` - might be an idea
Not sure if we should go into specific annotation etc. -> needs rethinking

jamesaoverton:
(Sorry, I just got back from a week off.)

I have come to regret the design of `remove`/`filter`, because nobody seems to understand it -- including myself, sometimes. When I sit down to use it (infrequently), I need to skim the docs again, and try some experiments before I hit on the right combination of options. A better design would be more "obvious" or "intuitive". In my defence, it's a tricky design problem.

Despite my regrets, we aren't changing the design of `remove`/`filter`. If somebody comes up with a better design, we'll make a new command (or commands).

This `--drop-axiom-annotations` in #1023 breaks the design, because it has no connection to any of the other parts of `remove`/`filter`.

A new "axiom annotations" option for `--axioms`, as originally proposed, would fit with the design. Then the base file examples from #1023 would work in two steps:

```
robot remove --input template.owl \
  --base-iri http://example.com/ \
  --axioms external \
  remove \
  --term IAO:0000117 \
  --term IAO:0000119 \
  --axioms axiom-annotations \
  --output results/template-drop-axiom-remove.owl
```

jamesaoverton:
[...]

PR #1023, opened 2022-07-06
Remove and filter operations new "drop_axiom_annotation" option

hkir-dev:
Resolves [#886]

New parameter `drop_axiom_annotation` added to `remove` and `filter` operations. An annotation property is provided to this option to drop/remove all axiom annotations using the property.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Add new command: [`expand`] [#964]
+- Add `--drop_axiom_annotation` option to drop axiom annotations in [`remove`] and [`filter`] [#1023]
```
matentzn, line 14: Always kebap case for robot long options; https://www.theserverside.com/definition/Kebab-case

docs/filter.md
```diff
@@ -28,6 +28,8 @@ Then `filter --term A --term R --term C --select "self parents" --axioms all --s
     - the objects for `ax2` are `{A, R, C}` (with `R some C` excluded), and all of these are in the target set, so `ax2` is matched and copied
     - the objects for `ax3` are `{D, E}`, and none of these are in the target set, so `ax3` is not matched and is not copied
 
+Finally `--drop_axiom_annotation` option lets you to specify an annotation property to drop all axiom annotations using it.
```
matentzn, line 31: kebap

docs/filter.md
```diff
@@ -94,3 +96,13 @@ Create a "base" subset that only includes internal axioms (alternatively, use `r
       --include-term IAO:0000119 \
       --output results/template-base-filter.owl
 
+Create a "base" subset that axiom annotations related with the two properties are removed:
```
matentzn, line 99: Create a "base" subset in which axiom annotations involving IAO:0000117 and IAO:0000119 are removed:

docs/remove.md
```diff
@@ -49,6 +49,8 @@ Then `remove --term A --term R --term C --select "self parents" --axioms all --s
     - the objects for `ax1` are `{A, B}`, and at least one of these is in the target set, so `ax1` is matched and removed
     - the objects for `ax2` are `{A, R, C}` (with `R some C` excluded), and at least one of these is in the target set, so `ax2` is matched and removed
     - the objects for `ax3` are `{D, E}`, and none of these are in the target set, so `ax3` is not matched and is not removed
+    
+Finally `--drop_axiom_annotation` option lets you to specify an annotation property to drop all axiom annotations using it.
```
matentzn, line 53: kebap

docs/remove.md
```diff
@@ -227,3 +229,12 @@ Create a "base" subset by removing external axioms (alternatively, use `filter -
       --select annotation-properties \
       --signature true \
       --output results/filter_annotations.owl
+
+Create a "base" subset that axiom annotations related with the two properties are removed:
```
matentzn, line 233: Create a "base" subset in which axiom annotations involving IAO:0000117 and IAO:0000119 are removed:

robot-command/src/main/java/org/obolibrary/robot/FilterCommand.java
```diff
@@ -43,6 +43,8 @@ public FilterCommand() {
     o.addOption("r", "trim", true, "if true, keep axioms containing only selected objects");
     o.addOption(
         "S", "signature", true, "if true, keep axioms with any selected entity in their signature");
+    o.addOption(
+        "d", "drop_axiom_annotation", true, "annotation property to drop axiom annotations");
```
matentzn, line 47: kebap

matentzn: "annotation property to drop axiom annotations" -> "drop all axiom annotations involving a particular annotation property"

robot-command/src/main/java/org/obolibrary/robot/FilterCommand.java
```diff
@@ -212,12 +216,40 @@ public CommandState execute(CommandState state, String[] args) throws Exception
...
+   * uses annotation properties specified by the drop_axiom_annotation argument.
+   *
+   * @param line CommandLine to get options from
+   * @param ioHelper IOHelper to get IRIs
+   * @param ontology OWLOntology to drop axiom annotations from
+   */
+  private void dropAxiomAnnotations(CommandLine line, IOHelper ioHelper, OWLOntology ontology) {
+    OWLOntologyManager manager = ontology.getOWLOntologyManager();
+    List<String> annotationsToDrop =
+        CommandLineHelper.getOptionalValues(line, "drop_axiom_annotation");
```
matentzn, line 239: keb.

robot-command/src/main/java/org/obolibrary/robot/RemoveCommand.java
```diff
@@ -198,12 +202,40 @@ public CommandState execute(CommandState state, String[] args) throws Exception
...
 
+  /**
+   * Given a command line and an IOHelper, drops axiom annotations from the given ontology which
+   * uses annotation properties specified by the drop_axiom_annotation argument.
+   *
+   * @param line CommandLine to get options from
+   * @param ioHelper IOHelper to get IRIs
+   * @param ontology OWLOntology to drop axiom annotations from
+   */
+  private void dropAxiomAnnotations(CommandLine line, IOHelper ioHelper, OWLOntology ontology) {
```
matentzn, line 222: Why isnt this method in OntologyUtils or some such? This should not be duplicated across both commands?

robot-core/src/main/java/org/obolibrary/robot/RelatedObjectsHelper.java
```diff
@@ -2344,4 +2344,26 @@ private static void spanGapsHelper(
...
+
+  /**
+   * Removes all of the axiom annotations for the given annotation property.
+   *
+   * @param ontology OWLOntology to get axioms
+   * @param property Annotation property to remove related axiom annotations.
+   */
+  public static void removeAxiomAnnotations(OWLOntology ontology, OWLAnnotationProperty property) {
+    OWLOntologyManager manager = ontology.getOWLOntologyManager();
+    for (OWLAxiom axiom : ontology.getAxioms()) {
```
matentzn, line 2356: We dont have to specify IMPORTS.INCLUDED or IMPORTS.EXCLUDED here?

matentzn (changes requested):
Excellent design to allow specifying the APs to remove.

One thing that would be great: can we add a blanket option for --drop-axiom-annotations (should be camel case, and should have an `s` in the end, because there can be multiple), that allows removing all axioms annotations? Like --drop-axiom-annotations `all`?

CHANGELOG.md
hkir-dev: Fixed

docs/filter.md
hkir-dev: Fixed

docs/remove.md
hkir-dev: Fixed

docs/remove.md
hkir-dev: Fixed

robot-command/src/main/java/org/obolibrary/robot/FilterCommand.java
hkir-dev: Fixed

robot-command/src/main/java/org/obolibrary/robot/FilterCommand.java
hkir-dev: fixed and method moved to `RelatedObjectsHelper.java`

robot-command/src/main/java/org/obolibrary/robot/RemoveCommand.java
hkir-dev: generic methods moved to `OntologyHelper.java` and methods only used by the FilterCommand and RemoveCommand are moved to `RelatedObjectsHelper.java`

robot-core/src/main/java/org/obolibrary/robot/RelatedObjectsHelper.java
hkir-dev: method moved to `OntologyHelper.java` and imports excluded

matentzn (approved):
The code looks great, let me summarise its current behaviour.

1. We introduce a new option `--drop-axiom-annotations` which removes all axiom annotations (or those using one or more specific properties) from a target set of axioms, determined by either the `filter` or `remove` commands. 
2. It works identically in `remove` and `filter`. This is cognitively much easier than applying "filter" logic in "filter" (i.e. only **including** the selected axiom annotations in the target set), which is correctly reserved to the --select/--axioms pipeline.

The only slight question mark I have, without wanting to cause an issue, is whether 

1. This should be a **command in its own right** after all `robot remove-axiom-annotations --select all`. Basically, we could then pipe the output of filter/remove to `remove-axiom-annotations` to achieve the same effect, without overloading the `remove` command too much.
2. Or, even if we go with the current implementation, we may not need to extend `filter`, and instead rely on passing the output of filter to `remove --drop-axiom-annotations all`. 

I will let @jamesaoverton think about this one, but I think this is the last remaining question here.

matentzn:
The implementation for this PR is correct, and we can use it to do what we want. But I have some doubts about the cleanliness of the design, as described in my comment above. To get in the headspace again:

Remove/filter does its thing to the very end.... and then, the target set (aka whatever remains) gets post processed using the --drop-axiom-annotations option. This does overload the filter and axiom commands a little bit, as this is, indeed, a conceptually independent operation. On the other hand: practically, I cant think of a case where we would use this command independently of filter and remove (maybe with a tiny exception of "extract", but very corner case), and there is overhead in adding a new command - you are a better judge of the trade-offs here.

jamesaoverton:
I realized that the integration tests were not running, so I made #1181 to fix that.

Issue #1030, opened 2022-07-20
Catalog not used for --input-iri


psiotwo:
Catalog file does not seem to be used for `--input-iri` (tested on `query` command). Based on the documentation I am not completely sure if it is a bug, or intended behaviour. Yet, for my use-case I would like to use the provided catalog for any IRI around (i.e. not only in `owl:imports` statements, but also in `--input-iri` parameters).

----
Example:

test.owl:
```
@prefix owl: <http://www.w3.org/2002/07/owl#> .
<http://test.org/test.owl> a owl:Ontology ;
    owl:imports <http://test.org/test-imported.owl> .
```

test-imported.owl:
```
@prefix owl: <http://www.w3.org/2002/07/owl#> .
<http://test.org/test-imported.owl> a owl:Ontology .
```

test.rq:
`ASK {}`

catalog-custom-name.xml
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<catalog prefer="public" xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog">
    <uri name="http://test.org/test-imported.owl" uri="test-imported.owl"/>
</catalog>

```

The following commands succeeds
`robot query --catalog catalog-custom-name.xml --input test.owl --query test.rq output.csv`
`robot query --input test-imported.owl --query test.rq output.csv`

while the following fails 
`robot query --catalog catalog-custom-name.xml --input-iri http://test.org/test-imported.owl --query test.rq output.csv`

with `org.semanticweb.owlapi.io.OWLOntologyCreationIOException: test.org`

jamesaoverton:
Yes, this is intended behaviour. Checking for a local catalogue file is cheap and fast, and matches Protege's behaviour. So ROBOT makes a guess and checks for local catalogs without asking. Making a network request based on a guess could take seconds and cause other unexpected behaviour.

@psiotwo are you proposing that when using `--input-iri http://example.com/foo.owl` we should interpret `--catalog cat.xml` as relative, so <http://example.com/cat.xml>? That seems slightly too "magical" to me. I doubt that anybody is using a remote input with a local catalogue, but changing this behaviour would break that.

Maybe we could add an explicit `--catalog-iri http://example.com/cat.xml` option.

balhoff:
[...]

PR #1035, opened 2022-07-28
[#1030] --input-iri uses catalog file

psiotwo:
Resolves [#1030]

`--input-iri` parameter now takes catalog file into account.


jamesaoverton:
The PR looks good. I'm worried about backwards compatibility. @matentzn Across your many projects, do you use `--input-iri` with a catalog file?

matentzn:
Yeah many times, but not intentionally: the ROBOT var in all our makefiles includes the --catalog reference. But I cant think of any case where things could break other than Uberon perhaps.. for all Makefile users some odditities May emerge in this case:

A file resolved with --input-url has an owl imports statement to some purl

That purl is an old stray in the catalog and has hence been removed. It now does not not resolve to a local file. 

Else I think it's fine.

PR #1056, opened 2022-09-30
Add `--input-format` option

### LINKED ISSUE #1038: Allow specifying input type in `robot convert`
Right now, when I get an error from `robot convert`, I have to add `-vvv` to get logs to figure out what is broken in the ontology, but this includes all of the logs from all of the different parsers robot tried. It would be nice to have a `--input-format` flag that lets me specify exactly what format the input it is, so this iteration through every possible parser type can be avoided and I can more easily get to my relevant logs

beckyjackson:
Resolves #1038

Creates a new global `--input-format` option which accepts one of the valid formats (`owl`, `obo`, `owx`, `ofn`, `omn`, `ttl`, or `json`) to specify which parser(s) to use when loading the ontology.

I would prefer to add `manager` to IOHelper instead of as a parameter for `loadOntology`, but that would require a big refactoring of the IOHelper class and potentially introduce some breaking changes? On the other hand, I don't want to build up a bunch of tech debt and this introduces a bunch more 'loadOntology' overrides. @jamesaoverton what do you think?

Last note - there were two test commands in the docs that were overwriting the example files. I fixed these so that they write to the results folder instead (had to remove an axiom to get the reason test to work).


robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
```diff
@@ -273,6 +283,34 @@ protected static boolean cleanTDB(String tdbDir) {
...
+      case "omn":
+        return Sets.newHashSet(new ManchesterOWLSyntaxOntologyParserFactory());
+      case "obo":
+        return Sets.newHashSet(new OBOFormatOWLAPIParserFactory());
+      case "ofn":
+        return Sets.newHashSet(new OWLFunctionalSyntaxOWLParserFactory());
+      case "owx":
+        return Sets.newHashSet(new OWLXMLParserFactory());
+      case "owl":
+        return Sets.newHashSet(new RDFXMLParserFactory(), new RioRDFXMLParserFactory());
```
matentzn, line 304: Why do we need to add the RIO parser here? Shouldnt we be more strict?

docs/global.md
```diff
@@ -131,6 +132,10 @@ _:blank rdf:type rdf:Statement .
 _:blank rdf:type owl:Axiom .
 ```
 
+## Input Format
+
+When loading an ontology, OWLAPI will attempt to load with a variety of ontology format parsers. When all parsers fail, many error logs are printed (when using verbose mode). Finding the cause of the error based on parser can be difficult due to number of messages. For debugging purposes, you can specify the `--input-format` of the ontology as one of: `owl`, `obo`, `owx`, `ofn`, `omn`, `ttl`, or `json`.
```
cthoyt, line 137: would be good to update this part of the documentation to have a table with columns:
1. extension (e.g., `owl`, `obo`, `owx`, `ofn`, `omn`, `ttl`, or `json`)
2. human readable name for this format
3. link to documentation about the format

jamesaoverton:
I think this addresses the use case in the issue. Thanks!

I looked into the suggestion about storing the `manager` in the `IOHelper` instance, but our JavaDocs say `@return a new ontology object, with a new OWLManager`, so I think we can't do that.

cthoyt:
@beckyjackson I am very happy you have been able to work on this - from a user perspective this will solve many of my debugging problems. Thank you :)

docs/global.md
cthoyt: Maybe important to make it more clear that these formats aren't actually the extension (which can be anything, even potentially mismatched, e.g., how VariO uses `.owl` for the OFN format; http://www.variationontology.org/vario_download/VariO_1.09.owl)

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
```diff
@@ -273,6 +283,34 @@ protected static boolean cleanTDB(String tdbDir) {
...
+      case "ttl":
+        return Sets.newHashSet(new RioTurtleParserFactory(), new TurtleOntologyParserFactory());
+      case "json":
+        return Sets.newHashSet(new RioJsonParserFactory(), new RioJsonLDParserFactory());
```
matentzn, line 308: This should be the OBOGraphs parser (stand-alone) or none at all I think.

TBH, I think the --input-format options should reflect identically the `-f` option. Why would we care about RioJson and not about the other more outlandish parsers like KRSS? I would want my parsing to fail if it try parsing with the exact parsers I specify - not because for some reason the second json parser in the list was successful because of its crappy implementation..

matentzn (commented):
Quite some changes to the core loading process. I went through it line by line, but remind me to do some extensive testing before the next release.

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
jamesaoverton: I was thinking that we should just divide up the parsers that OWLAPI currently uses, but you're right that if we're adding a new option then we don't have to be constrained in that way.

matentzn: The way you do it for `remove` is IMO the best. We use the same extensions:parser mappings here as in `-f`, but, if we ever so want to (I don't), add a bunch more switches later for compare complex combinations `any-owl` for ofn, owx, rdfxml, KRSSParser for krss etc (analogous to axiomtype option in `--axioms`). I would not bother here - the `-f` should be by far sufficient.

jamesaoverton:
I created an issue to discuss auto-detection of formats. @beckyjackson Don't feel obligated to dive even deeper into this -- this is already a great start!

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
beckyjackson: I spent quite a bit of time digging around OWLAPI and could not find anything for OBOGraphs. Here's a full list of the parsers in a default manager (`OWLManager.createOWLOntologyManager()`):
* RDFXMLParserFactory
* OWLXMLParserFactory
* OWLFunctionalSyntaxOWLParserFactory
* RioTurtleParserFactory
* ManchesterOWLSyntaxOntologyParserFactory
* RioNQuadsParserFactory
* RioJsonParserFactory
* RioNTriplesParserFactory
* RioTrigParserFactory
* RioBinaryRdfParserFactory
* RioJsonLDParserFactory
* RioN3ParserFactory
* RioRDFXMLParserFactory
* RioTrixParserFactory
* TurtleOntologyParserFactory
* OBOFormatOWLAPIParserFactory
* KRSS2OWLParserFactory
* RioRDFaParserFactory

Unless I'm missing something, I only see the Rio JSON parsers.

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
beckyjackson: I don't have strong feelings one way or another. I figured it would be used if RDFXML failed without `--input-format` anyway? @jamesaoverton do you have a preference for excluding Rio parsers?

docs/global.md
beckyjackson: Good idea - thanks for the suggestion.

robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
matentzn: OBOgraphs parser is not in owlapi..  If it is anywhere, it is in the obographs package..

beckyjackson: I had checked there as well (https://github.com/geneontology/obographs), but didn't see anything about a parser. The manager currently doesn't use OBOGraphs anyway when loading any JSON files, so it seems we're using one of the Rio parsers when we're loading in any JSON. We just use the `OgJsonGenerator` to write OBO graphs JSON.

The OG reader ([OgJsonGenerator](https://github.com/geneontology/obographs/blob/master/obographs-core/src/main/java/org/geneontology/obographs/core/io/OgJsonReader.java)) doesn't produce an OWLOntology.

I was curious to see what's going on... Interestingly, if you load an OG JSON file into ROBOT and print out the format:
```java
// Load the ontology
OWLOntology loadedOntology = manager.loadOntologyFromOntologyDocument(source, config);

// Check for unparsed triples - get the document format and then the loader metadata
OWLDocumentFormat f = manager.getOntologyFormat(loadedOntology);
```

...it tells you it's RDF/XML. Looks like at some point while loading, it's converted:
```
DEBUG Converting from JSON to RDF
```

balhoff (changes requested):
I looked at this code because I wanted to use the same approach in a different project. But I ran into a problem there which I confirmed also is a problem here. If you limit the parsers to a particular set, all the *imported* ontologies must also use that format. :-(

Try making an ro.ttl and setting input format to ttl. When it tries to import `http://purl.obolibrary.org/obo/ro/annotations.owl` (RDF/XML), it will fail.

matentzn:
@balhoff There is nothing though we can do about it, right? The only thing we can do is document that behaviour on the docs pages and that's that...

balhoff:
@matentzn I think you're right, but I'm not sure we can add this feature if it is inherently broken.

matentzn:
I am not going to fight for this feature. I just want to point out it is inherently broken only in cases where owl:imports are involved.. But yeah, I am not sure if this downside completely invalidates this PR.

balhoff:
I think we might want to go ahead with this if we document the limitation well. I'm running into a problem in another project where I really need to specify the format. It's a huge ontology and running through all the parsers is hitting memory problems.

jamesaoverton:
I don't have any plans for this, so I'm going to convert to Draft. If someone really wants this, feel free to push it forward.

PR #1061, opened 2022-10-10
Add support for exporting domain and range

### LINKED ISSUE #1025: rdfs:domain and range don't get exported
I've tried every way I can think of to export rdfs:domain and rdfs:range properties using export function.  None of them ever produce any exported info for the properties.  using "rdfs:domain" or the full property IRI doesn't produce any error message either.  It would be good to have these available for export....

beckyjackson:
Resolves #1025

I generalized `classExpressionsToString` and `getClassCell` to handle all OWL Objects (including IRIs) to support this without adding extra code. The same code can be used for properties, so that was replaced.


jamesaoverton:
Thanks @beckyjackson!!

PR #1073, opened 2022-11-15
Update to latest HermiT.

balhoff:
@ignazio1977 pointed out there is a newer release of HermiT available.


jamesaoverton:
I tested this with OBI and got identical OWL files for the main artefact and all our views/subsets.

PR #1095, opened 2023-02-16
Adding temporary gradle files to .gitignore

matentzn:
No related issue, just saw this after testing the latest version of ROBOT locally.


jamesaoverton:
I don't know where this is coming from, but yes it should be ignored.

PR #1104, opened 2023-03-21
Fix how Template adds entities to the QuotedEntityChecker

### LINKED ISSUE #1133: Question: How to correctly create/preserve cardinality constraints on ObjectProperties using template?
Dear ROBOT team,

We need to create OWL files from UML class diagrams, which contain all taxonomy, partonomy and topological relationships with inverses and cardinalities. The UML tool generates Excel files, which can easily processed to be usable for ROBOT [template](https://robot.obolibrary.org/template). Hence, we use a staged process, which 

1. creates the classes via a simple "ID | SC %" template, followed by 
2. creating the ObjectProperties with their domains/ranges via a template "ID | TYPE | DOMAIN | RANGE".

Now, since the UML-to-Excel export contains also distinct columns for the specified min/max cardinalities, these shall be also properly set in a third step, so that restrictions like 
```
Class: <Process>
SubClassOf: 
   (<contains> min 1 <Task>) and (<contains> max 5 <Task>)
```
are correctly inserted. In this example, both "Process" and "contains" have been successfully created als Class and ObjectProperty, respectively, in steps 1 and 2. 

However, there are several issues with the third step:
- In the UML-to-Excel export there are cases where the same ObjectProperty comes in multiple rows with identical domain but different ranges. Hence, in such cases the range should become a union, however, ROBOT template always creates intersections. Is there a way to change this behavior?
- Even when using manually created subclass axioms in the template, e.g., "(contains min 1 Task)", attempting to merge this with the previously created valid outputs from steps 1 and 2 either with [template ](https://robot.obolibrary.org/template) or with [merge ](https://robot.obolibrary.org/merge) in any combination always results in "Task" being redeclared as a Datatype instead of the Class it has actually been created as before (the owl:Restriction then gives an owl:onDataRange instead of owl:onClass). Protégé also gives a punning warning here. How can this be avoided, so that the previous class definitions and object properties are correctly used?
- Is there a way to set up a template that permits directly using the numerical contents for the cardinalities in creating the correct Manchester syntax for the restrictions and involved entities, e.g. from Excel "From-Class | Restricted-Property | To-Class | Min-Card. | Max-Card."?

Thank you in advance!

### LINKED ISSUE #1156: template axiom annotations do not work for class assertions on individuals
\>A columns to the right of TI columns do not add anything to the output ontology (there is also no error message - just silent failure to add annotation)
using the same template but changing to classes (so SC column for axiom to be annotated) works fine

matentzn:
(Updated by @jamesaoverton)

Resolves #1133, resolves #1156.

Replaces private `Template.addLabels()` with `Template.addEntities()`: for each row in the template, make sure to add an entity to the QuotedEntityChecker (QEC). The previous version skipped rows without a LABEL. The new version adds both the label (if present) and the ID to the QEC. I believe this is backwards compatible.

I had problems with CI that seem unrelated, so I also dropped a test for subset extraction. See #1166.


anitacaron:
It's not really a cycle but a sequence path.

matentzn:
I think the problem is that:

If there is no labels column,

Template.addLabels() never adds anything to the `QuotedEntityChecker`, which is needed to interpret curies in the ManchesterOWLSyntax parser. but I may be wrong.

anitacaron:
Thank you, Nico! You said in Slack that adding column `'TYPE': 'TYPE'` to the template solves the issue. I tried adding the label, and it didn't work, but adding the type Class worked well!

matentzn:
Thanks for confirming. It's still important for us to get to the bottom of this!

docs/extract.md
```diff
@@ -79,12 +79,14 @@ For more details see the [MIREOT paper](http://dx.doi.org/10.3233/AO-2011-0087).
...
-        --term "BFO:0000050" \
-        --output results/subset_result.owl
+```
+robot extract --method subset \
+    --input subset.obo \
+    --term "obo:ONT_1" \
+    --term "obo:ONT_5" \
+    --term "BFO:0000050" \
+    --output results/subset_result.owl
+```
```
matentzn, line 89: Why this change?

robot-core/src/main/java/org/obolibrary/robot/Template.java
```diff
@@ -559,33 +559,23 @@ private void addTable(List<List<String>> rows) throws Exception {
 
   /** Add the labels from the rows of the template to the QuotedEntityChecker. */
   private void addLabels() {
-    // If there's no label column, we can't add labels
-    if (labelColumn == -1) {
-      return;
-    }
```
matentzn, line 565: Removing this will necessitate looping through potentially a lot of rows for no reason (the phrase "for no reason" applies to the "looping", not the "removing"); I am assuming you are removing this because you want to add the "entity --> id" combination to the EntityChecker, right?

robot-core/src/main/java/org/obolibrary/robot/Template.java
```diff
@@ -559,33 +559,23 @@ private void addTable(List<List<String>> rows) throws Exception {
...
         // ignore
       }
 
-      if (idColumn != -1 && id == null) {
+      if (id == null) {
         continue;
       }
 
-      if (id == null || label == null) {
-        continue;
```
matentzn, line 588: Removing this does change the behaviour of the loop considerably; but it makes sense, since you want to add "id in case entity label is missing".

robot-core/src/main/java/org/obolibrary/robot/Template.java
```diff
@@ -653,7 +643,13 @@ private void addLabels() {
           entity = dataFactory.getOWLEntity(EntityType.NAMED_INDIVIDUAL, iri);
           break;
       }
-      checker.add(entity, label);
+
+      if (id != null) {
+        checker.add(entity, id);
+      }
+      if (label != null) {
+        checker.add(entity, label);
```
matentzn, line 651: This is probably the only thing someone need to physically check: does adding both (id AND label) ensure that both syntaxes can interpreted by the entitychecker? I assume so!

matentzn (commented):
@jamesaoverton looked through your changes, made some small comments for sanity checking but most likely everything is correct.

docs/extract.md
jamesaoverton: This test is causing intermittent CI failures. I can't see any connection to the changes in this PR. I opened a new issue #1166.

robot-core/src/main/java/org/obolibrary/robot/Template.java
jamesaoverton: What your test case here, #1133, and #1156 all have in common is that none of you are using a LABEL column. I had assumed that everyone would want a LABEL and would use the label in Manchester strings, but in these cases an ID and a TYPE are enough.

So we will change the behaviour so that every row ends up in the QuotedEntityChecker, and that fixes these three problems. I *think* it's a safe change.

robot-core/src/main/java/org/obolibrary/robot/Template.java
jamesaoverton: Yes, that's the decision.

robot-core/src/main/java/org/obolibrary/robot/Template.java
jamesaoverton: Yes, this should be tested, but I feel pretty confident about it. The bigger problem I've noticed is that `QuotedEntityChecker.add(entity, name)` will happily, quietly replace existing names.

robot-core/src/main/java/org/obolibrary/robot/Template.java
matentzn: Yes, makes sense! I agree with you, just wanted to make sure we both thought this way!

PR #1106, opened 2023-03-28
Preserve prefixes from loaded ontology.

balhoff:
Resolves #1101.

- [ ] `docs/` have been added/updated
- [ ] tests have been added/updated
- [ ] `mvn verify` says all tests pass
- [ ] `mvn site` says all JavaDocs correct
- [ ] `CHANGELOG.md` has been updated

This change gets the document format used to load the ontology, and sets it as the prefix manager for the document format being used to save the ontology.

PR #1107, opened 2023-03-28
Add a test case for conservation of prefixes.

gouttegd:
Add a test to #1106 to ensure that prefixes initially present in an ontology are preserved when the ontology is saved through IOHelper.


matentzn (approved):
Will merge this into the feature branch, this looks great.

PR #1119, opened 2023-05-16
Adding custom plugins to ROBOT

gouttegd:
This PR adds the possibility of using _pluggable commands_, or commands that are not part of the built-in set of commands but are instead provided by third-party plugins.

# Rationale

Reasons to enable the use of plugins include:

* fast prototyping and testing of possible future built-in commands (a new command could be first distributed as a plugin, without requiring a ROBOT release, before being ultimately co-opted to become a built-in command);
* allowing the creation and use of “niche” commands that someone may need, but that have a scope too narrow to warrant being included a built-in commands;
* lowering the bar for using Java to create specialized OBO manipulation tools (by delegating command line parsing and I/O operations to ROBOT, and focusing instead solely on the task at hand);
* simplifying and improving complex pipelines that currently require alternating between ROBOT and other tools, if those other tools could be replaced by ROBOT (pluggable) commands.

# Proposed behaviour

This PR implements the following behaviour:

At startup, after the built-in commands have been registered to the `org.obolibrary.robot.CommandManager`, ROBOT looks for Jar files in a pre-defined set of directories:

* the directory specified by the `robot.pluginsdir` Java system property, if any;
* the directory specified by the `ROBOT_PLUGINS_DIRECTORY` environment variable, if any;
* the calling user’s `~/.robot/plugins` directory.

In each Jar file found in those directories, ROBOT looks for a [service provider](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) for the `org.obolibrary.robot.Command` interface – that is, it looks for a resource file named `META-INF/services/org.obolibrary.robot.Command`, containing a list of classes implementing that interface. If such a service is found in a Jar file, the corresponding classes (i.e. the commands) are added to the set of commands recognized by the `CommandManager`.

To avoid clashes between command names (e.g. a plugin providing a command with the same name as a built-in command, or two plugins providing two commands with the same name), the basename of the Jar file that provides a command is prepended to the actual command name (as returned by the `org.obolibrary.robot.Command#getName()` method) when the command is added to the `CommandManager`. For example, if the file `foo.jar` provides a command named `bar`, that command will be registered under the name `foo:bar` – that’s how it will be listed in ROBOT’s help message, and how it will need to be invoked on the command line.

# How to create a plugin

Developers wishing to create a plugin have to:

1) Write a class implementing the `org.obolibrary.robot.Command` interface, for each command they wish to create, the same way the built-in commands are implemented.

2) Package the compiled Java classes in a Jar file.

3) Add to that Jar file a `META-INF/services/org.obolibrary.robot.Command` file, containing a list of all the fully qualified class names representing the commands they wish to “expose” to ROBOT.

# Examples

My [robot-plugins](https://git.incenp.org/damien/robot-plugins) repository contains an example of a plugin providing two pluggable commands: `hello`, a pure “proof-of-concept” command (it does nothing useful except injecting a silly annotation into the current ontology), and `merge-species`, a re-implementation of owltools’s `--merge-species-ontology` command.

Another example is available in my [ocelliless](https://github.com/gouttegd/ocelliless) repository, where a ROBOT command version of the `ocelliless` tool exists in the [as-robot-plugin](https://github.com/gouttegd/ocelliless/tree/as-robot-plugin) branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Support for pluggable commands [#1119]
+
 ## [1.9.4] - 2023-05-23
 
 ### Changed
@@ -365,6 +368,7 @@ First official release of ROBOT!
 [`template`]: http://robot.obolibrary.org/template
 [`validate`]: http://robot.obolibrary.org/validate
 
+[#1119]: https://github.com/ontodev/robot/pull/1119
 [#1100]: https://github.com/ontodev/robot/pull/1100
 [#1091]: https://github.com/ontodev/robot/issues/1091
 [#1089]: https://github.com/ontodev/robot/issues/1089
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -27,6 +27,7 @@
           <a href="/chaining">chaining commands</a><br>
           <a href="/global">global options</a><br>
           <a href="/make">makefile</a><br>
+          <a href="/plugins">plugins</a><br>
           - - - - - - - - - -<br>
           <a href="/annotate">annotate</a><br>
           <a href="/collapse">collapse</a><br>
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -1,0 +1,36 @@
+# Plugins
+
+The set of ROBOT commands can be extended locally with plugins. A ROBOT plugin is a Java archive file (`.jar`) providing one or more supplementary commands (hereafter called "pluggable commands").
+
+## Using plugins
+
+ROBOT searches for plugins in the following locations:
+
+* the `.robot/plugins` directory in the current user's home directory;
+* the directory specified by the Java system property `robot.pluginsdir`, if such a property is set;
+* the directory specified by the environment variable `ROBOT_PLUGINS_DIRECTORY`, if such a variable is set in the environment.
+
+Installing a plugin is therefore simply a matter of either
+
+* placing the Jar file into your `~/.robot/plugins` directory, or
+* placing the Jar file into any directory and making sure ROBOT knows to search that directory, by setting the `robot.pluginsdir` system property or the `ROBOT_DIRECTORY_PLUGINS` environment variable accordingly.
+
+Importantly, the basename of the Jar file (without the `.jar` extension) within the directory will become part of the name of any pluggable command provided by the plugin. For example, if the file is named `myplugin.jar` and it provides a command called `mycommand`, that command will be available under the name `myplugin:mycommand`. Because of that:
+
+* the name of the Jar file **must** be in lowercase only;
+* the name **should** be kept short and simple.
+
+Once the plugin is installed, any pluggable command it provides is immediately available to ROBOT. You can check by calling `robot` without any argument to get it to print the full list of available commands, which will include the commands provided by installed plugins, if any.
+
+## Creating plugins
+
+A pluggable command, just like any other ROBOT command, is a Java class that implements the `org.obolibrary.robot.Command` interface. A plugin is Java archive file that contains at least:
+
+* the compiled Java code ("bytecode") for at least one class implementing the `org.obolibrary.robot.Command` interface, and
+* a `META-INF/services/org.obolibrary.robot.Command` file that list all implementations of that interface available in the archive (one per line).
+
+For example, if the command `mycommand` is implemented in a class named `MyCommand` in the package `org.example.myplugin`, the `META-INF/services/org.obolibrary.robot.Command` file must contain a single line `org.example.myplugin.MyCommand`.
+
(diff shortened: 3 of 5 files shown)


udp:
I think this will be an extremely valuable addition and help the community consolidate on robot for I/O instead of having ad hoc scripts everywhere using different methods of loading/saving ontologies.

The main concern for us at EBI would be the addition of plugins breaking existing scripts, but I think the prefixing `foo:bar` comprehensively solves this.

balhoff:
@gouttegd how do plugins deal with dependencies that aren't already in ROBOT?

gouttegd:
@balhoff Such dependencies would need to be provided in the same Jar file as the one containing the plugin itself.

The best way to deal with that, I believe, would be to use the Maven plugin `shade` to selectively include or exclude the dependencies from the Jar to be built (excluding ROBOT itself and everything that comes with it – OWL API, ELK, etc. —, but including everything else).

The interesting question is: what would happen if two different plugins (in two different Jar files) both depends on the same libraries that are not already in ROBOT? My understanding is that the Java virtual machine would use the class files from whatever Jar file is loaded first, but I would need to do more testing to be sure. Thanks for raising this point.

balhoff:
@gouttegd that sounds reassuring, although I guess it might be possible for a plugin to break if a later version of ROBOT drops or changes a dependency that the plugin was using. I suppose if a plugin bundles a dependency that's already used by ROBOT, there would be a clash?

gouttegd:
> it might be possible for a plugin to break if a later version of ROBOT drops or changes a dependency that the plugin was using

Yes. More generally, any change in ROBOT could break a plugin, if the change touches something that the plugin was dependent on. A change in the dependencies bundled with ROBOT (e.g. if ROBOT no longer depends on a given library and therefore no longer bundles it, or if it bundles a newer version, etc.) is just one possibility among others.

> I suppose if a plugin bundles a dependency that's already used by ROBOT, there would be a clash?

No “clash” per se, but the plugin would not see the version of the dependency that it is bundled with – the version bundled with ROBOT would always take precedence. Whether this would lead to problems for the plugin would depend on how much the version provided by ROBOT and the version expected by the plugin differs. Best case scenario, the two versions are compatible and nothing bad happens; worst case scenario, the plugin could crash.

In any case, because ROBOT-provided dependencies always take precedence, breakage can only happen in one direction: ROBOT can break plugins, but plugins cannot break ROBOT. This I believe is the most important – we don’t want ROBOT to start behaving differently depending on whether some plugins are present or not.

jamesaoverton:
This is a good idea @gouttegd, and well explained. I haven't had time to look at it in detail yet, but it's on my to do list.

matentzn:
@gouttegd I think we can finalise this PR. Do you want to do it? I will deal with getting it merged when James is back. I can also finalsie it (CHANGELOG etc), just let me know how you want to play it. (for now, the "docs" update should probably be a single section in the global robot main docs, with a link to your own docs, we can work on folding this in more detail later).

gouttegd:
I can take care of updating the CHANGELOG and the doc (can probably do that tomorrow).

Not sure about the tests, though. Ideally we’d want to test that invoking a plugin works, but obviously that requires having a plugin to test. Either we test by downloading one of my already available plugins (e.g. the KGCL or SSSOM one), but I don’t like the idea of the test suite being dependent on something external to the repository, or we add a “mock” plugin in this very repository (such a plugin could also act as an example of how to create a plugin).

gouttegd:
@matentzn Documentation and changelog updated.

The only thing that is missing is some tests – for now all we have are the existing tests to show that the addition of the plugin feature does not break the CommandManager, but we have no tests to show that the plugin feature is working as expected.

Adding such tests would be relatively “invasive”, since ideally I think we should at a minimum:

* add an entire new top-level Maven module (e.g. `robot-plugin`), containing a “mock” plugin;
* build that module along with `robot-core` and `robot-command`;
* add a new test fixture in `robot-command` that tries to launch ROBOT after setting either the `robot.pluginsdir` property or the `ROBOT_PLUGINS_DIRECTORY` env variable and check that it loads the plugin correctly.

matentzn:
Does not sound like a very enjoyable task :D I am a bit on the fence if we need this or not. I guess it would be better if there were proper tests in place for this. I think your suggestion sounds great!

gouttegd:
I’ll work on that on a separate branch. No promises as to when I’ll find the time to do it, though. Your call (and James’) on whether it is worth waiting for or if you’re happy with the tests being added _a posteriori_.

matentzn:
No promises at all needed for anything - the earliest I can even _talk_ with James about this is last week of August :D

gouttegd:
There is a rudimentary test on a new PR (#1142).

I am not entirely happy with it as it only tests the bare minimum (trying to load one plugin from the `ROBOT_PLUGINS_DIRECTORY` environment variable), but I am not sure testing more conditions would be worth the added complexity.

jamesaoverton:
Closing in favour of #1142

PR #1135, opened 2023-08-02
Update OWL API to 4.5.26

matentzn:
Resolves #1129

- [ ] `docs/` have been added/updated (not necessary)
- [x] tests have been added/updated
- [x] `mvn verify` says all tests pass
- [X] `mvn site` says all JavaDocs correct
- [X] `CHANGELOG.md` has been updated

The last OWL API version (4.5.25) had some major issues with the way complex turtle axioms are written, see #1129. This version solves this problem.

PR #1137, opened 2023-08-05
Rename edit-annotated because of clash with other QC check

matentzn:
- [ ] `docs/` have been added/updated (No need)
- [X] tests have been added/updated (the nature of this PR)

In my attempts to migrate to the latest OWLAPI (#1135), I encountered a problem I could not understand which @gouttegd solved for me. Basically, the problem is that the test suite is slightly overlapping:

```
robot merge --input edit.owl --input edit2.owl --output results/merged2.owl
robot merge --inputs "edit*.owl" --output results/merged2.owl
```

Applying this PR, the test now passes,  because the file then no longer matches the edit*.owl pattern in the second command that produces merged2.owl above, so it is no longer merged and therefore the resulting merged2.owl is identical to what the test expects.

Thank you @gouttegd for this uncovering this, it really made me puzzle!

What is, however, a bit unclear: why did this error only show up now?

Something for future generations to puzzle about.


matentzn:
EDIT: This works, and fixes, indeed #1135. @gouttegd in James absence, can you give me the thumbs up on this PR, so I can merge it?

gouttegd:
If you mean formal approval, no I can’t: I am merely a contributor here, I don’t have commit privileges on this repo.

If you want an informal thumbs up, sure: I tested myself, it works. :+1:

matentzn:
@gouttegd thank you, that is all I needed.

PR #1141, opened 2023-08-08
Update command to create robot.bat file

### LINKED ISSUE #1140: Typo in command to create robot.bat file
There seems to be a typo in the command to create the `robot.bat` batch file on the ROBOT [Getting Started](https://robot.obolibrary.org/) page:

```
"java %ROBOT_JAVA_ARGS% -jar %~dr0robot.jar %*" | out-file robot.bat -encoding utf8
```
`%~dr0` doesn't seem to be a valid argument expansion syntax. I think this should be `%~dp0%`, based on the [batch script file](https://raw.githubusercontent.com/ontodev/robot/master/bin/robot.bat) and the help for the `CALL` command:
```
        %~dp1       - expands %1 to a drive letter and path only
```
Furthermore, the command only seems to work in Windows PowerShell (not Command Prompt), which the guide doesn't specify.

Also, in older versions of PowerShell (version 6 and below), the `utf8` encoding still inserts a byte-order mark (BOM), which breaks the command even after the typo is corrected (see [here](https://stackoverflow.com/questions/5596982/using-powershell-to-write-a-file-in-utf-8-without-the-bom/)).
```
C:\Users\user\robot>´╗┐java  -jar C:\Users\user\robot\robot.jar
'´╗┐java' is not recognized as an internal or external command,
operable program or batch file.
```
The documentation should probably specify that a specific version of PowerShell is required to create the batch file using the command.

jseager7:
This PR changes the Windows installation instructions, fixing a typo in the command to create the `robot.bat` file and clarifying that the command should be run with a minimum PowerShell version.

I've tested this on Windows 10 with PowerShell version 7 and it seems to work fine.

Resolves #1140

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -40,7 +40,7 @@ The command-line tool is packaged a Java JAR file and can be run via the `robot`
 2. Save the [ROBOT batch script](https://github.com/ontodev/robot/raw/master/bin/robot.bat).
     - Make sure this is saved as `.bat` and not `.bat.txt`
     - OR in [PowerShell](https://learn.microsoft.com/powershell/), run `"java %ROBOT_JAVA_ARGS% -jar %~dp0robot.jar %*" | out-file robot.bat -encoding utf8` in the same directory as `robot.jar` to create the batch script.
-    - Note that the above command requires PowerShell version 6 or later: previous versions will write a Unicode byte order mark to the file, which breaks the command.
+    - Note that the above command requires PowerShell version 6 or later: previous versions will write a Unicode byte order mark (BOM) to the file, which breaks the command.
 3. Put both files on your [system PATH](https://en.wikipedia.org/wiki/PATH_(variable)) in the same directory.
     - this could be `C:\Windows\`
     - OR [update your PATH](https://docs.oracle.com/javase/tutorial/essential/environment/paths.html) to include the new directory.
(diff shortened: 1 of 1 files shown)


matentzn:
@jseager7 thanks a ton!

@ehartely would you be able to review this PR for me? No rush at all, just checking.

docs/index.md
```diff
@@ -39,7 +39,8 @@ The command-line tool is packaged a Java JAR file and can be run via the `robot`
 1. Download the `robot.jar` file from the [latest release](https://github.com/ontodev/robot/releases/latest).
 2. Save the [ROBOT batch script](https://github.com/ontodev/robot/raw/master/bin/robot.bat).
     - Make sure this is saved as `.bat` and not `.bat.txt`
-    - OR enter `"java %ROBOT_JAVA_ARGS% -jar %~dr0robot.jar %*" | out-file robot.bat -encoding utf8` in the same directory as `robot.jar` to create the batch script.
+    - OR in [PowerShell](https://learn.microsoft.com/powershell/), run `"java %ROBOT_JAVA_ARGS% -jar %~dp0robot.jar %*" | out-file robot.bat -encoding utf8` in the same directory as `robot.jar` to create the batch script.
+    - Note that the above command requires PowerShell version 6 or later: previous versions will write a Unicode byte order mark to the file, which breaks the command.
```
ehartley, line 43:
```suggestion
    - Note that the above command requires PowerShell version 6 or later: previous versions will write a Unicode byte order mark (BOM) to the file, which breaks the command.
```

ehartley:
This PR looks good to me. I just made 1 suggestion to add "(BOM)" since I think the acronym is more recognizable. 

I have PowerShell version 5.1, so I confirmed that running `"java %ROBOT_JAVA_ARGS% -jar %~dp0robot.jar %*" | out-file robot.bat -encoding utf8` in an older version creates the `robot.bat` file with a BOM. However, there aren't any other issues since the `robot` command works fine after I re-save the file without the BOM encoding.

jseager7:
@ehartley Thanks for the suggestion. I've added the BOM acronym.

PR #1147, opened 2023-09-09
Fixes robot-core Junit test failures due to line ending differences

### LINKED ISSUE #1146: Tests pass on Mac/Linux but fail on Windows
6 of the robot-core JUnit tests incorrectly fail on Windows but pass on Mac/Linux due to line ending differences in expected vs calculated values. 
The tests and causes are:
```
DiffOperationTest.testCompareModifiedWithLabels
DiffOperationTest.testCompareModified
   These read the contents of the simple1.diff file which uses CRLF and compare it to a generated string which uses LF.
```
```
IOHelperTest.testPrefixHandling
IOHelperTest.testAddPrefixes
   IOHelper.getContextString() uses JsonUtils.toPrettyString() which returns a platform-specific form (CRLF) and compares it to a hard-coded string using LF.
```
```
ReportOperationTest.testReportProducesValidJson
ReportOperationTest.testReportProducesValidYaml
  These read the "/1016-report-json-failure/output.[json,yaml]" files which is OS-specific and compares it to the generated string which is OS agnostic
```
Other than line endings all the above tests return identical values on both Windows and Mac/Linux

Although not elegant, the simplest fix to all of the above is just to regex the OS-specific strings to convert it to *nix form before comparing strings in the existing tests.
Note that the files are correct in the git repo. The conversion to use CRLF is done by git when the repo is cloned on a Windows machine.

dlutz2:
Resolves [#1146]

- [x] `docs/` have been added/updated - no doc changes needed
- [x] tests have been added/updated - fixes existing test
- [x] `mvn verify` says all tests pass - on Windows and *nix
- [x] `CHANGELOG.md` has been updated - minor code change, no change to log

[DESCRIPTION, fixes #1146 by replacing Windows line endings with \n in failing tests. ]

### DIFF
--- a/robot-core/src/test/java/org/obolibrary/robot/DiffOperationTest.java
+++ b/robot-core/src/test/java/org/obolibrary/robot/DiffOperationTest.java
@@ -55,7 +55,8 @@ public class DiffOperationTest extends CoreTest {
     assertFalse(actual);
     String expected =
         IOUtils.toString(
-            this.getClass().getResourceAsStream("/simple1.diff"), Charset.defaultCharset());
+                this.getClass().getResourceAsStream("/simple1.diff"), Charset.defaultCharset())
+            .replaceAll("\r\n", "\n");
     assertEquals(expected, writer.toString());
   }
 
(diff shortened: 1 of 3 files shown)


jamesaoverton:
Much appreciated!

PR #1148, opened 2023-09-10
Export cmd bugfixes

### LINKED ISSUE #1131: ROBOT export generates error in large excel exports
When trying to export an ontology with more than 64,000 classes into an Excel file, ROBOT (v1.9.4) export generated an error "_The maximum number of Cell Styles was exceeded. You can define up to 64000 style in a .xlsx Workbook_", which I believe is coming from the POI library. I think it is caused by the the `CellStyle style = wb.createCellStyle();` call in robot.export.Row.addToWorkbook() which creates a new CellStyle for each row. The export works fine when exporting to text and html.

### LINKED ISSUE #1144: Expected behavior of export "Disjoint With"
ROBOT 1.9.4
The export operation with a "Disjoint With" heading results in a set of class/property expressions which includes the entity being described e.g. in BFO:
"_continuant_" is described as being disjoint from "_continuant|occurrent_". 
This is the behavior of the underlying OWLAPI EntitySearcher which apparently returns the members of the mutually disjoint set including the entity of interest. 
Is including a self-disjoint value the desired result for export? Reconstructing the original ontology from the exported values (e.g. using template) requires a manual edit to remove that self-disjoint,

### LINKED ISSUE #1145: Export throws exception when using "Equivalent Property" heading
ROBOT 1.9.4
When exporting an ontology containing an annotation property (defined or used), an export command with an "Equivalent Property" header throws an `ClassCast` exception when attempting to cast an Annotation property to an Object property in `EntitySearcher.getEquivalentProperties()`. This is called from `ExportOperation.getRow()` in the `switch (iriStr)` statement with case "http://www.w3.org/2002/07/owl#equivalentProperty" with an Annotation Property as the entity. The `if (entity.isOWLAnnotationProperty())` portion of that case statement should be a no-op since there is no `Equivalent Annotation Property` axiom in OWL.
EntitySearcher naively assumes if the input property is not a data property its an Object Property, doesn't check and attempts to cast the input to an ObjectPropertyExpression
This can be reproduced with an ontology which contains a single annotation property declaration.

dlutz2:
Resolves [#1145, resolves #1144, resolves #1131]

- [x] `docs/` have been added/updated - no doc changes needed
- [x] `CHANGELOG.md` has been updated - minor changes
 
fixes #1145 - removed case statement looking for "equivalent annotation property", added test exporting all named headers, added test ontology containing one of all OWL axioms and class expressions, plus some puns
fixes #1144 - removed self-disjoint entry in returned entities
fixes #1131 - refactored to avoid creating unnecessary styles and fonts, added matching test

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/ExportOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/ExportOperation.java
@@ -11,13 +11,9 @@ import org.semanticweb.owlapi.apibinding.OWLManager;
 import org.semanticweb.owlapi.model.*;
 import org.semanticweb.owlapi.search.EntitySearcher;
 import org.semanticweb.owlapi.util.*;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /** @author <a href="mailto:rbca.jackson@gmail.com">Becky Jackson</a> */
 public class ExportOperation {
-  /** Logger. */
-  private static final Logger logger = LoggerFactory.getLogger(ExportOperation.class);
 
   /** Namespace for error messages. */
   private static final String NS = "export#";
@@ -728,11 +724,10 @@ public class ExportOperation {
             }
           }
           break;
+          // TODO
         case DATA_HAS_VALUE:
-          logger.warn(t + "is not yet implemented as target of data property" + ce);
+          break;
         default:
-          // This should never happen in valid OWL
-          logger.error("A " + t + " cannot be a target of a data property:" + ce);
           break;
       }
     }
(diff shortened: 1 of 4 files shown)


matentzn:
@dlutz2 thank you so much for your contribution! I will try to find a reviewer for this now, forgive use that it may take a bit of time. Your work is highly appreciated!

matentzn (commented):
This is only a partial review now, because we have a big backlog for reviews. I asked a community member for a cursory review to get started, and this is what they said:

- The code does not follow, in some places at least, the “Google Java style” that is used by ROBOT. 
- The fix for issue 1131 (export to Excel with > 64k rows) is only a partial one. The problem would still occur if there are more 64k rows with a non-null violationLevel, because the code still creates a new style for each cell in that case. That being said, the PR is a major improvement compared to the current situation where a new style is created for every cell, even those without a violation level. So I would still accept the fix while noting that it can be improved even further later on.
- For issue 1144, no comment except that (as already noted in the issue), maybe the proper fix belongs to the OWL API (the behaviour of EntitySearcher.getDisjointClasses() does seem wrong to me, or its documentation is misleading).

Let me know @dlutz2 what you think!

matentzn:
As an amendment to the above from our collaborator @gouttegd:

ROBOT uses the `com.coveo:fmt-maven-plugin` Maven plugin to enforce its code style as part of the build process, so all you have to do is to check out the branch and do a simple build (or even just mvn process-sources if you don’t want to do a full build); then use git diff to check what changes the formatter made to the source.

For example:

```
-    for(short colIx=minColIx; colIx<maxColIx; colIx++) {
-       Cell cell = header.getCell(colIx);
-      if(cell == null) {
+    for (short colIx = minColIx; colIx < maxColIx; colIx++) {
+      Cell cell = header.getCell(colIx);
+      if (cell == null) {
```

This is also how the authors can fix their PR: all they have to do is to run `mvn process-sources` and commit the result.

dlutz2:
Thanks for the review.
I'll look at the Excel issue with many non-null violation Levels and add that to the branch. (I assume that is the proper approach to amend a PR)
Sorry about the formatting. I'm using Eclipse on Windows and both the built in formatter and the google-java-format code seem to have issues. I had not looked at the Maven plugin and didn't realize it had formatting capability. Thanks for the pointer, I'll reformat and add that to the branch as well.

dlutz2:
Another bit: Is there anything I can do to make the review easier (e.g. smaller PRs?) or is it just a matter of time/attention of the reviewers?

gouttegd:
Regarding the formatting: if you’re using Eclipse, you may want to use the code style description file available [here](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml). This way Eclipse will automatically format the code using the expected style as you save it.

Regarding the Excel issue: as I noted, the PR is already a major improvement over the current situation, so I personally would be fine with accepting it as it is and leave any further improvement for another PR. That being said, if you do want to try to improve it right now, then of course feel free to do so!

jamesaoverton:
Looks good! I'm very happy that we can now support up to 64k rows with violations, up from 64k total rows. If someone needs more than that, they can make another PR :smile:. The additional tests are much appreciated!

Issue #1157, opened 2023-10-12
'--annotate-with-source true'  does not work with extract --method subset


allysonlister:
(Apologies if I have named this ticket incorrectly, please update as required).

I'm trying to add `--annotate-with-source true` to my ODK build procedure and I'm testing the run with this robot command:

`java -jar ~/Programs/robot.jar --catalog catalog-v001.xml query -i mirror/po.owl --update ../sparql/preprocess-module.ru extract -T imports/po_terms_combined.txt --copy-ontology-annotations true --force true --individuals exclude --annotate-with-source true --method subset --output test-po.owl`

The above is a modified version of the command that is generated by the ODK's Makefile, and I'll put it in my own makefile if I can get it working. However, this command does not add any rdfs:isDefinedBy annotations at all.
My terms to extract only has one class (http://purl.obolibrary.org/obo/PO_0025131) and it looks great EXCEPT for the lack of an rdfs:isDefinedBy property as outlined in the ROBOT command:

```
<owl:Class rdf:about="http://purl.obolibrary.org/obo/PO_0025131">
        <obo:IAO_0000115>An anatomical entity that is or was part of a plant.</obo:IAO_0000115>
        <oboInOwl:created_by>rwalls</oboInOwl:created_by>
        <oboInOwl:creation_date>2010-11-15T11:41:38Z</oboInOwl:creation_date>
        <oboInOwl:hasDbXref>PO_GIT:224</oboInOwl:hasDbXref>
        <oboInOwl:hasExactSynonym>entidad anat&amp;#243mica vegetal (Spanish, exact)</oboInOwl:hasExactSynonym>
        <oboInOwl:hasExactSynonym>植物 解剖学（形態）的実体 (Japanese, exact)</oboInOwl:hasExactSynonym>
        <oboInOwl:hasOBONamespace>plant_anatomy</oboInOwl:hasOBONamespace>
        <oboInOwl:id>PO:0025131</oboInOwl:id>
        <rdfs:comment>Includes both material entities such as plant structures and immaterial entities such as plant anatomical spaces. CARO:0000000 &apos;anatomical entity&apos; is defined as: A part of a cellular organism that is either an immaterial entity or a material entity with granularity aboove the level of a protein complex.  Or, a substance produced by a cellular organism with granularity above the level of a protein complex.  Refers to BFO:0000004 &apos;independent continuant&apos;.</rdfs:comment>
        <rdfs:label>plant anatomical entity</rdfs:label>
    </owl:Class>
[...]

PR #1160, opened 2023-10-20
Extract subset "annotate-with-source" support

hkir-dev:
Resolves [#1157, resolves #I1157]

`--annotate-with-source true` support added to the `extract --method subset`.

Code review revealed that  `annotate-with-source` is only supported by `SLME` (not supported by `MIREOT` and `subset`). This pull request adds `annotate-with-source` support to the `subset` method.

While `SLME` and `subset` commands take place in the ExtractOperation.java, `MIREOT` is implemented in the ExtractOperation.java. Adding `annotate-with-source` support to the MIREOT requires further refactoring if we want to do it without code redundancy.

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/ExtractOperation.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/ExtractOperation.java
@@ -189,20 +189,7 @@ public class ExtractOperation {
     }
     // Maybe annotate entities with rdfs:isDefinedBy
     if (OptionsHelper.optionIsTrue(options, "annotate-with-source")) {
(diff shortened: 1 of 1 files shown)


matentzn:
Awesome initiative @hkir-dev THANKS

jamesaoverton:
Thanks @hkir-dev! What do you think about sharing code with [`MireotOperation.maybeAnnotateSource()`](https://github.com/ontodev/robot/blob/master/robot-core/src/main/java/org/obolibrary/robot/MireotOperation.java#L782) ?

jamesaoverton:
Let's not worry about the duplication.

PR #1168, opened 2023-11-29
Handle IRIs that are not entities in export

### LINKED ISSUE #1165: Export command does not export a column for "rdfs:isDefinedBy"
An export command that includes `rdfs:isDefinedBy` in the header doesn't export any values for this column. 
I believe the issue is in ExportOperation.getPropertyValues() where it attempts to find an entity in the ontology to use as the value for this column. Since the target for rdfs:isDefinedBy is usually an IRI for an ontology (not a term in an ontology), the call `ontology.getEntitiesInSignature(iri)` never returns an entity (the IRI for the ontology and any imported ontologies are not in the signature for the ontology ) so no value is created.
This could be handled either by checking for the IRIs for the ontology and imported ontologies in addition to the term IRIs or just using the IRI as the value if no entities are found for that IRI.

jamesaoverton:
Resolves #1165

### DIFF
--- a/robot-core/src/test/java/org/obolibrary/robot/ExportOperationTest.java
+++ b/robot-core/src/test/java/org/obolibrary/robot/ExportOperationTest.java
@@ -2,6 +2,7 @@ package org.obolibrary.robot;
 
 import static org.junit.Assert.assertEquals;
 
+import java.io.FileOutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


robot-core/src/test/java/org/obolibrary/robot/ExportOperationTest.java
```diff
@@ -152,9 +155,18 @@ public void testExportNamedHeadingsSimple() throws Exception {
...
+    // Write test.xlsx for debugging
+    // try {
+    //   System.out.println("Writing result to robot-core/test.xlsx");
+    //   FileOutputStream out = new FileOutputStream("test.xlsx");
+    //   wb.write(out);
+    //   out.close();
+    // } catch (Exception ex) {
+    //   System.out.println("Error: " + ex);
+    // }
+
```
matentzn, line 167: Do you like keeping code like the above in your output?

matentzn (approved):
LGTM

Issue #1166, opened 2023-11-28
Integration test for extract subset failing in CI


jamesaoverton:
See #1104 and recent Action results. The integration test here [docs/extract.md#subset](https://github.com/ontodev/robot/blob/master/docs/extract.md#subset) always passes for me locally, but it's failing 90% of the time on my runs for #1104. I really don't get it. I saw a "zio-fiber" error (?) in relationgraph on one run, but I can't find that message now.

Any ideas?

jamesaoverton:
I saw that zio error again, this time in a unit test:

```
testExtractSubset(org.obolibrary.robot.ExtractOperationTest)  Time elapsed: 0.598 sec  <<< ERROR!
Error: Exception in thread "zio-fiber-19" java.lang.InterruptedException: Interrupted by thread "zio-fiber-19"
	at org.renci.relationgraph.RelationGraph.processRestrictionAndExtendQueue(RelationGraph.scala:131)
	at org.renci.relationgraph.RelationGraph.computeRelations.streamZ(RelationGraph.scala:75)
```

jamesaoverton:
@balhoff I fixed my link to point to the first attempt of that run, which failed. I just re-ran it and then it worked.

Yes this is hard to debug, but also really disruptive to development workflows.

jamesaoverton:
This issue is breaking CI for another (unrelated) PR: https://github.com/ontodev/robot/actions/runs/7107481969/job/19348979228#step:5:21700

I need a solution for this. The best I've got right now is to disable the test, which sucks.

jamesaoverton:
Hopefully closed by https://github.com/ontodev/robot/pull/1174

jamesaoverton:
I think this is still a problem. A large percentage of my CI runs have been failing, e.g. https://github.com/ontodev/robot/actions/runs/22157649265/job/64065844347#step:5:12256

I never see the problem locally.

```
Error:  Errors: 
Error:    ExtractOperationTest.testExtractSubset:175 » Interrupted
```

balhoff:
@jamesaoverton I'm trying something in relation-graph. Hopefully I'll come back soon with a solution.

PR #1174, opened 2023-12-06
Update to latest relation-graph.

balhoff:
This is just an update to the latest relation-graph library, which itself is mainly an update to its dependencies. I'm hoping that updating to a newer ZIO dependency will resolve the mysterious test issues reported in #1166.


matentzn:
> one of those ran for an hour and I restarted it.

What do you mean exactly?

jamesaoverton:
We aren't paying for GitHub Actions in this repo. Won't they eventually timeout?

balhoff:
This one was in the test running phase for about an hour, so it seemed hung to me.

Issue #1183, opened 2024-02-09
relax is currently incomplete


cmungall:
`A subClassOf B and R some C` is currently shielded from relax.

This has a long chain of unintended consequences...

Ideally relax would be provably complete, I believe can be done by first converting to NNF, more later...

cmungall:
For now I am relaxing in place [using this](https://chat.openai.com/share/854231ba-6243-4596-a789-b82595b7f77b)

(perl, functional syntax, and ChatGPT... what could go wrong?)

cmungall:
Additionally it would be useful to do relaxation of cardinality constraints too.

E.g.

```
id: UPa:UCR00783
name: H(2)O + ferricytochrome c + nitric oxide = H(+) + ferrocytochrome c + nitrite
namespace: reaction
xref: KEGG:R00783 "KEGG reaction"
xref: RHEA:15236 "Rhea reaction"
xref: METACYC:NITRITE-REDUCTASE-CYTOCHROME-RXN "MetaCyc REACTION" {source=Rhea,source=MetaCyc}
relationship: has_input_compound UPa:UPC00001 {cardinality="1"} ! UPa:UPC00001 ! H(2)O
relationship: has_input_compound UPa:UPC00125 {cardinality="1"} ! UPa:UPC00125 ! ferricytochrome c
relationship: has_input_compound UPa:UPC00533 {cardinality="1"} ! UPa:UPC00533 ! nitric oxide
relationship: has_output_compound UPa:UPC00080 {cardinality="1"} ! UPa:UPC00080 ! H(+)
relationship: has_output_compound UPa:UPC00088 {cardinality="1"} ! UPa:UPC00088 ! nitrite
relationship: has_output_compound UPa:UPC00126 {cardinality="1"} ! UPa:UPC00126 ! ferrocytochrome c
relationship: part_of UPa:UER00707 {cardinality="1", order="1", direction="RL"}
```

Should generate someValuesFrom on relaxation

gouttegd:
Is that really within the scope of `relax` though?

ROBOT [currently defines](http://robot.obolibrary.org/relax) the `relax` operation as being solely about removing _equivalence axioms_:

> Robot can be used to relax Equivalence Axioms to weaker SubClassOf axioms.
> […]
[...]

PR #1188, opened 2024-03-10
Relax command extension

### LINKED ISSUE #1060: Relax relax
Right now, relax indiscriminately relaxes all equivalent class axioms regardless of their shape. The original idea was to be able to turn logical definitions of the form `A = B and R some C and R some D and R some E` (simple conjunctions) into

```
A sub B
A sub R some C
A sub R some D
A sub R some E
```

Using relax on all axioms indiscriminately does not make much sense, see here:

https://ols.monarchinitiative.org/ontologies/upheno2/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FFYPO_0004107

There is little value in relaxing a complex axiom like that.

I would like to propose adding a new option to relax:

```
--simple-equivalent-classes-only
```

Which restricts the operation of relax to only axioms of the form:

`A = B and C and D`

where B, C and D are all class names or simple existential restrictions.

**Motivation:**

The main use case of `relax`, as far as I understand, is to project a complex logical structure using equivalent class axioms into a more graphy structures that can be leveraged by graph processing. The workflow usually goes something like: `relax->reduce-> delete equivalent class axioms` to generate a simple representation of the ontology ala OBO format.

For uPheno, I need to be able to run `relax`, and then delete all equivalent class axioms in the ontology and rewrite them to something else. Unfortunately, because ODK mandates the `relax->reduce` chain in default releases, I do get these terrible subClassOf axioms everywhere now - which are much much harder to filter out after the fact with ROBOT.

matentzn:
Fixes #1060, #1183

- [X] `mvn site` says all JavaDocs correct

These PR extends the relax command by adding three new options:

- `--exclude-named-classes`: if true, A EquivalentTo B between named classes are excluded from `relax`
- `--include-subclass-of`: if true, `A Subclass A and R some B` is `relaxed` the same way as `A EquivalentTo A and R some B`, see @cmungall request in #1183 
- `--enforce-obo-format`: If true, only simple existentials and named subsumptions are asserted during the process

I like all the functionality, but the `--enforce-obo-format` is really clunky. I would be happy if someone made a better suggestion; I also dont like `--exclude-nested`. 

The only "breaking" change in this PR is that the CLI gets `--exclude-named-classes` by default as `true`. Previously, before this option existed, A=B was relaxed to A sub B and B sub A. No one I personally now has ever said that this is a good idea, so we changed the default, as discussed in #1060.


docs/relax.md
```diff
@@ -58,6 +67,75 @@ Running `reduce` will remove the redundant axiom (5), leaving the following axio
 
 This SubClassOf graph is complete and non-redundant, and can be used for intuitive visualization and browsing of the ontology
 
+## Note about normalization of Qualified Number restrictions
+
+For convenience of downstream processing, `relax` rewrites expressions of the kind `:R min 1 :A` to `:R some :A`. These two expressions are logically identical.
```
balhoff, line 72: What about `:R min 2 :A` (or any number greater than zero)?

robot-core/src/main/java/org/obolibrary/robot/RelaxOperation.java
```diff
@@ -89,6 +90,34 @@ public static Map<String, String> getDefaultOptions() {
    * @param options A map of options for the operation
    */
   public static void relax(OWLOntology ontology, Map<String, String> options) {
```
balhoff, line 92: This `options` map is completely ignored?

balhoff (approved):
Looks good, but I had two questions.

docs/relax.md
matentzn:
```
else if (x instanceof OWLObjectCardinalityRestriction) {
      OWLObjectCardinalityRestriction ocr = (OWLObjectCardinalityRestriction) x;
      OWLClassExpression filler = ocr.getFiller();
      OWLObjectPropertyExpression p = ocr.getProperty();
      if (ocr.getCardinality() > 0) {
        OWLObjectSomeValuesFrom svf = dataFactory.getOWLObjectSomeValuesFrom(p, filler);
        svfs.add(svf);
      }
```

All of them are rewritten actually. 

```suggestion
For convenience of downstream processing, `relax` rewrites expressions of the kind `:R min 1 :A` or `:R min 2 :A` to `:R some :A`. This is safe, because `:R some :A` is implied by any cardinality restriction > 0.
```

matentzn: Thanks for pointing that out, clarified the docs.

robot-command/src/main/java/org/obolibrary/robot/RelaxCommand.java
```diff
@@ -24,6 +23,21 @@ public RelaxCommand() {
     Options o = CommandLineHelper.getCommonOptions();
     o.addOption("i", "input", true, "relax ontology from a file");
     o.addOption("I", "input-iri", true, "relax ontology from an IRI");
+    o.addOption(
+        "e",
+        "enforce-obo-format",
+        true,
+        "if true, only axioms allowed in OBO format are asserted as a consequence of relax.");
+    o.addOption(
+        "x",
```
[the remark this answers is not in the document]
matentzn: Excellent, done

PR #1193, opened 2024-04-08
Add --drop-axiom-annotations feature with value constraints

### LINKED ISSUE #886: New parameter --axioms "axiom annotations" for remove and filter

matentzn:
Resolves #886

This PR implements the last part of #886, which is to allow "value" expressions as constraints on the --drop-annotation-assertion parameter.

This does not at all affect the current behaviour on the `master` branch, and is a bit incomplete compared to the more powerful selector in `remove` and `filter`: only exact values and regex constraints are supported, nothing else (in particular not constraints like `<IRI>`).

- [ ] @jamesaoverton note that i cannot run `mvn site` anymore: I get thousands of errors. Not sure why!


robot-core/src/main/java/org/obolibrary/robot/OntologyHelper.java
```diff
@@ -1660,11 +1662,15 @@ public static void trimOntology(OWLOntology ontology) {
    * @param ontology OWLOntology to remove axiom annotations
    * @param properties Annotation property IRIs to remove related axiom annotations.
    */
-  public static void removeAxiomAnnotations(OWLOntology ontology, List<IRI> properties) {
```
jamesaoverton, line 1663: This breaks a public interface by removing the method with this signature. We try hard to avoid that. Can the old method/signature (with `List<IRI>`) be maintained (maybe as a thin wrapper) while the new method (with `Map<IRI, String>`) is added?

jamesaoverton (changes requested):

robot-core/src/main/java/org/obolibrary/robot/OntologyHelper.java
matentzn: Done! Thanks you!

jamesaoverton (approved):

PR #1194, opened 2024-04-08
Make repair command more flexible

### LINKED ISSUE #1099: Feature request: Add a parameter that enables merging multiple instances of the same axiom with different sets of axiom annotations
Derived from #1098

When terms are defined in two different inputs, the act of merging may lead to duplicate annotations. It would be nice to have a quick way of removing such duplicates when is clear that the annotation is indeed the same.

EDIT BY @matentzn 

- [ ] Fold in docs from https://github.com/ontodev/robot/pull/1103 when fixing this

matentzn:
Resolves #1099, and makes repair command more flexible.

This PR makes it possible to selectively invoke repair functions rather than having all or nothing approach (which still works by default).

WARNING: This PR contains a breaking change: 

```
robot repair --input in.owl --merge-axiom-annotations true -o out.owl
```

would have previously ran the _entire_ repair pipeline (`RepairOperation.repair()`). Now, it will only run the merge axiom annotations pipeline.


docs/repair.md
```diff
@@ -17,7 +17,14 @@ To repair an ontology (with the name `need-of-repair.owl`) run the following com
...
 This will generate a new file `results/repaired.owl`. You can compare this with the original file (either using unix `diff` or [robot diff](diff)). If the changes that were made look good then you can simply replace the source file with the repaired file (`mv results/repaired.owl need-of-repair.owl`).
 
-By default, annotation axioms are not migrated to replacement classes. However, this can be enabled for a list of annotation properties passed either as arguments to `--annotation-property` or in a term file `--annotation-properties-file`:
+To use ROBOT solely for merging axiom annotations, you can use:
+
+    robot repair --input uberon_axiom_annotation_merging.owl \
+      --merge-axiom-annotations true \
+      --output results/uberon_axiom_annotation_merged.owl
+
+By default, annotation axioms are not migrated to replacement classes.
```
matentzn, line 26: @balhoff what does this mean? I dont get this part of the repair method at all. This sentence also makes no sense to me.

jamesaoverton: @balhoff ?

balhoff: If you use that option, you can automatically move xrefs from an obsolete class to its replacement.

matentzn: I have updated the documentation to make the command a bit clearer, but this particular part I still dont get it.

Lets say we have

```
A sub R some B
```

And B is deprecated.

When I "repair" my ontology I would assume this happens:

```
A sub R some X
```

where X is a replacement of B.

How does this relate to "fixing" anything about B? This seems like a totally different command to me (`robot deprecate` or similar).

balhoff: I guess it's "repairing" the ontology itself.

PR #1196, opened 2024-04-22
Fix docs on missing_subset_declaration.rq

cmungall:
The previous docs were copypastad without modification

### DIFF
--- a/robot-core/src/main/resources/report_queries/missing_subset_declaration.rq
+++ b/robot-core/src/main/resources/report_queries/missing_subset_declaration.rq
@@ -1,6 +1,6 @@
 # # Missing Subset Declaration
 #
-# **Problem:** A synonym type is used in an annotation, but is not properly declared as a child of oboInOwl:SynonymTypeProperty. This can cause problems with conversions to OBO format.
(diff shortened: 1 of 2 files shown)


robot-core/src/main/resources/report_queries/missing_subset_declaration.rq
```diff
@@ -1,6 +1,6 @@
 # # Missing Subset Declaration
 #
-# **Problem:** A synonym type is used in an annotation, but is not properly declared as a child of oboInOwl:SynonymTypeProperty. This can cause problems with conversions to OBO format.
+# **Problem:** A subset is used in an annotation (via inSubset), but is not properly declared as a child of oboInOwl:SubsetProperty. This can cause problems with conversions to OBO format, and should be avoided as all subsets should have metadata
```
matentzn, line 3:
```suggestion
# **Problem:** A subset is used in an annotation (via oboInOwl:inSubset), but is not properly declared as a child of oboInOwl:SubsetProperty. This can cause problems with conversions to OBO format, and should be avoided as all subsets should have metadata.
```

PR #1200, opened 2024-05-09
Update OWLAPI to version 4.5.28

matentzn:
Update OWLAPI to version 4.5.28

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -209,7 +209,7 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>log4j-over-slf4j</artifactId>
-      <version>1.7.32</version>
+      <version>2.0.11</version>
     </dependency>
     <dependency>
       <groupId>au.csiro</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -209,7 +209,7 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>log4j-over-slf4j</artifactId>
-      <version>2.0.11</version>
+      <version>1.7.32</version>
     </dependency>
     <dependency>
       <groupId>au.csiro</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -204,12 +204,12 @@
     <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
-      <version>1.3.0</version>
+      <version>1.5.6</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>log4j-over-slf4j</artifactId>
-      <version>2.0.11</version>
+      <version>2.0.13</version>
     </dependency>
     <dependency>
       <groupId>au.csiro</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -166,7 +166,7 @@
     <dependency>
       <groupId>net.sourceforge.owlapi</groupId>
       <artifactId>owlapi-api</artifactId>
-      <version>4.5.26</version>
+      <version>4.5.28</version>
       <exclusions>
         <exclusion>
           <groupId>com.google.code.findbugs</groupId>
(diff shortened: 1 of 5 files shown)


balhoff (approved):

balhoff:
Not specific to this PR, but it made me notice that we have tons of references to a nonexistent FOAF property ("depicted_by"). It should be `foaf:depiction`. I think this came from a mistake in Uberon a long time ago.

pom.xml
```diff
@@ -204,12 +204,12 @@
     <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
-      <version>1.2.13</version>
+      <version>1.3.0</version>
```
gouttegd, line 207: This is fine.

But just FYI, it would also be possible to make ROBOT compile against the new OWLAPI without having to drastically bump the versions of the logging libraries (Logback and Log4-over-SLF4J): simply add an explicit dependency to `org.slf4j:slf4j-api` version 1.7.x.

The reason the build fails if we do not update the logging libraries is because the OWLAPI declares a dependency to `org.slf4j:slf4j-api` 2.x. ROBOT “inherits” that dependency (since it depends on the OWLAPI), so when we build the `robot.jar` archive, that archive ends up containing a copy of the `slf4j-api` 2.x (inherited from the OWLAPI) and a copy of Logback-classic (explicitly declared here as a direct dependency) — and those two versions are not compatible.

But SLF4J-API is strictly backwards compatible. Just because the OWLAPI declares it depends on `slf4j-api` >= 2.x does not mean it cannot work with `slf4j-api` 1.x.

So we could explicitly declare `slf4j-api` 1.7.x as a _direct_ dependency of ROBOT here (alongside Logback-classic and Log4J-over-SLF4J). We would then end up with a `robot.jar` archive containing only versions of logging libraries that are compatible with each other (SLF4J-API < 2 and Logback-classic < 1.3).

Not saying that we _should_ do that, of course. It is perfectly fine to bump to Logback-classic 1.3 and Log4J-over-SLF4J 2.0. Just mentioning the other possibility.

gouttegd (approved):

pom.xml
matentzn: Oh thanks @gouttegd I didnt know that!

@jamesaoverton what is your preference? Bumping versions up, or adding this additional (transitive) dependency as a constraint?

jamesaoverton: Thanks @gouttegd for the explanation.

I despise Java logging, which may be clouding my judgement. But I think the simplest thing is just to bump these two dependencies to their latest versions: 

- pkg:maven/ch.qos.logback/logback-classic@1.5.6
- pkg:maven/org.slf4j/log4j-over-slf4j@2.0.13

This built and passed the test suite for me locally, so I don't see the benefit of adding yet another dependency (`slf4j-api`) that I don't understand. If I misunderstood something, please let me know.

gouttegd: > I despise Java logging

I just spent the last 24 hours debugging issues caused by Java logging libraries, so… I share your feeling! :)

> I don't see the benefit of adding yet another dependency (slf4j-api) that I don't understand. If I misunderstood something, please let me know.

Just that ROBOT is _already_ dependent on `slf4j-api` – it’s just that currently that dependency is an indirect one.

But I fully agree that adding the explicit dependency on `slf4j-api` just to avoid bumping `logback-classic` and `log4-over-slf4j` would have no particular benefit in ROBOT’s case. It would only be useful if for some reason we _wanted_ to keep using older versions of those libraries.

matentzn: The slf4j upgrade has been removed. Only the logback upgrade remains now.

jamesaoverton:
@matentzn Why not use the latest versions of these dependencies?

- pkg:maven/ch.qos.logback/logback-classic@1.5.6
- pkg:maven/org.slf4j/log4j-over-slf4j@2.0.13

matentzn:
I have no opinion on these matters, other than not constraining the dependencies more than absolutely necessary- I leave it to you and am very happy to support using the most up to date versions!

jamesaoverton:
Ok. Since we're updating, I'd like to update to the latest stable releases of these dependencies.

gouttegd (approved):

PR #1201, opened 2024-05-25
Update Elk to 0.6.0 in pom.xml

matentzn:
This PR adds and updated version of the Elk reasoner to ROBOT, see https://github.com/liveontologies/elk-reasoner/issues/48#issuecomment-2130090254.

- [ ] `docs/` have been added/updated (NA)
- [ ] tests have been added/updated (Note needed)


ykazakov:
Are you sure you need `elk-distribution-owlapi` and not just `elk-owlapi`? Also, log4j was removed from ELK, so there is no need to exclude this dependency.

matentzn:
> Are you sure you need elk-distribution-owlapi and not just elk-owlapi? Also, log4j was removed from ELK, so there is no need to exclude this dependency.

@ykazakov thanks for the comment. When I checked I think there was only "distribution", but most likely I was just blind. I changed it now, and it does work!

PR #1205, opened 2024-06-19
Multi annotation column

pk-mitre:


jamesaoverton:
The PR looks good. I'm a bit hesitant about this feature, because I don't see the benefit of using a single column over multiple columns. Can you describe your use case? Is it that your single AP column might contain many different annotation properties?

jamesaoverton:
I'm still having trouble seeing the use case for this, but I'm not opposed to it. Can you please complete the documentation for this feature in `template.md`?

dlutz2:
I'd like to see this capability added. We use dozens of annotation properties for curation and domain specific descriptions but an individual term is likely to use only a small number (but necessarily the same ones in a batch). Putting both the property and literal into a single pipe-separated field keeps us from maintaining very wide but sparse spreadsheets and we don't have to change the template when we add new annotation properties.

jamesaoverton:
Thanks @dlutz2. Have you used or tested this PR?

dlutz2:
We've been using the multi-annotation capability in production since it was proposed. Been working fine.

jamesaoverton:
I made some changes and generalizations to this in #1205. Please take a look.

PR #1207, opened 2024-06-21
Attempt 1: replacing DLExpressivity by DLExpressivityChecker

matentzn:
This is try to sort out the problem that the previous solution shows gibberish expressivity levels:

"expressivity": "RRESTRCUCINTUNIVRESTREROIF(D)". But it does not work..

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/metrics/OntologyMetrics.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/metrics/OntologyMetrics.java
@@ -8,7 +8,6 @@ import org.obolibrary.robot.IOHelper;
 import org.obolibrary.robot.providers.CURIEShortFormProvider;
 import org.semanticweb.owlapi.metrics.AbstractOWLMetric;
 import org.semanticweb.owlapi.metrics.AverageAssertedNamedSuperclassCount;
-import org.semanticweb.owlapi.metrics.DLExpressivity;
 import org.semanticweb.owlapi.metrics.GCICount;
 import org.semanticweb.owlapi.metrics.HiddenGCICount;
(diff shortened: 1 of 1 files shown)


ckindermann:
A workaround might be using `Collection<Languages> languages = checker.expressibleInLanguages();` If the set is not a singleton, we'd still need to determine the least expressive [language](https://owlcs.github.io/owlapi/apidocs_4/org/semanticweb/owlapi/util/Languages.html) using [isSubLanguageOf](https://owlcs.github.io/owlapi/apidocs_4/org/semanticweb/owlapi/util/Languages.html#isSubLanguageOf-org.semanticweb.owlapi.util.Languages-).

ckindermann:
Your commit message is not correct: `expressibleInLanguages()` does not return all languages. Each language in the collection 
- allows for all necessary constructors (meaning the ontology can be expressed in the language),
- is minimal regarding sublanguages (meaning the set does not include more expressive languages than necessary).

However, there may be multiple [languages](https://owlcs.github.io/owlapi/apidocs_4/org/semanticweb/owlapi/util/Languages.html) satisfying these criteria. In this case, sorting just makes the output deterministic.

matentzn:
Thank you @ckindermann - I will make sure to rectify the commit message when its time to create a squash merge.

ckindermann:
Thanks! In any case, `getDescriptionLogicName()` is the correct function to call. The OWL API currently implements this incorrectly by concatenating the set of constructors used in the ontology (see  [here](https://github.com/owlcs/owlapi/blob/version4/api/src/main/java/org/semanticweb/owlapi/util/DLExpressivityChecker.java#L174) in version 4 and [here](https://github.com/owlcs/owlapi/blob/version5/api/src/main/java/org/semanticweb/owlapi/util/DLExpressivityChecker.java#L176) in version 5).

PR #1211, opened 2024-07-31
Treat raw IRI as an OWLClass in logical template cells

jamesaoverton:
When ROBOT expects to see a class label or CURIE in a template cell, and it sees a raw IRI instead, interpret it as an OWLClass.

This is in response to a report on Slack.

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/TemplateHelper.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/TemplateHelper.java
@@ -992,11 +992,13 @@ public class TemplateHelper {
         // This allows class labels with single quotes
         expr = checker.getOWLClass(content, false);
         if (expr == null) {
-          // If not found by label, is it just an IRI?
-          IRI iri = IRI.create(content);
-          if (iri != null) {
-            // If so, create a new class for this IRI.
-            expr = checker.getOWLClass(content, true);
+          // If not found by label, is it just an HTTP IRI?
+          if (content.startsWith("http")) {
+            IRI iri = IRI.create(content);
+            if (iri != null) {
+              // If so, create a new class for this IRI.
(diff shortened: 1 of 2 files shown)


robot-core/src/main/java/org/obolibrary/robot/TemplateHelper.java
```diff
@@ -991,6 +991,14 @@ protected static OWLClassExpression tryParse(
         // If we have a checker, try to get the class by label
         // This allows class labels with single quotes
         expr = checker.getOWLClass(content, false);
+        if (expr == null) {
+          // If not found by label, is it just an IRI?
+          IRI iri = IRI.create(content);
```
balhoff, line 996: @jamesaoverton if someone enters a CURIE using an unknown prefix, is there a danger of turning it into an IRI here? In my experience OWL API will turn those into fake IRIs where the prefix becomes a scheme, e.g., "UNKNOWN:1234" -> `<UNKNOWN:1234>`. I find those mistakes really maddening (just speaking in general, not about robot template). Should you make some assumptions about IRIs here, such as only doing this if it starts with `http:`, `https:`, `urn:`, etc.?

balhoff (commented):
Seems good but I had one question.

robot-core/src/main/java/org/obolibrary/robot/TemplateHelper.java
jamesaoverton: Good point. I agree.

jamesaoverton: I added a unit test that 'UNKNOWN:1234' fails, and it did without any other changes. But I also added a check for HTTP IRIs.

PR #1215, opened 2024-08-15
create skeleton for adding indirect existentials in materialize

### LINKED ISSUE #1213: Add a --include-indirect option to materialize
From slack:

Hi all, I have a question about the use of the materialize command. Let's consider the following axioms:

```
car SubclassOf has part some wheel
convertible SubclassOf car
```

I would like to assert the inferred axiom:

`convertible SubclassOf has part some  wheel`

My comprehension is that I should use the materialize command like this: robot materialize -i input.owl --reasoner ELK --term IRI:convertible -o output.owl

However, I don't see any difference between input and output (Robot v1.9.6)

https://github.com/INCATools/ontology-access-kit/releases/tag/v0.6.4

souzadevinicius:
Resolves #1213 


docs/examples/hubcap.owl
```diff
@@ -0,0 +1,85 @@
...
+    //
+    // Object Properties
+    //
+    ///////////////////////////////////////////////////////////////////////////////////////
+     -->
+
+    
+
+
+    <!-- http://www.semanticweb.org/vinicius/ontologies/2024/7/untitled-ontology-2#has_part -->
```
matentzn, line 24: These are very ugly IRIs, maybe run a string replace on the file to replace

```
http://www.semanticweb.org/vinicius/ontologies/2024/7/untitled-ontology-2
```

with something like

```
http://example.org/ontology
```

robot-command/src/main/java/org/obolibrary/robot/MaterializeCommand.java
```diff
@@ -44,6 +44,8 @@ public MaterializeCommand() {
     o.addOption("o", "output", true, "save reasoned ontology to a file");
     o.addOption("t", "term", true, "a property to materialize");
     o.addOption("T", "term-file", true, "load properties from a file");
+    o.addOption(
```
matentzn, line 47: option should be "false" by default

robot-core/src/main/java/org/obolibrary/robot/MaterializeOperation.java
```diff
@@ -143,7 +169,7 @@ public static void materialize(
         logger.debug("Excluding classes not in main ontology: " + c);
         continue;
       }
-      Set<OWLClassExpression> sces = emr.getSuperClassExpressions(c, true);
+      Set<OWLClassExpression> sces = emr.getSuperClassExpressions(c, !includeIndirect);
```
matentzn, line 172: Ok this _seems_ to make sense! Needs someone from gene ontology reasoner time to check.. @balhoff @cmungall

PR #1218, opened 2024-10-08
dockerfile - fixed pom location error - reduced image size by close to 2/3

styk-tv:
Existing Dockerfile did not work. It couldn't find pom files of one of the dependencies. I have updated Dockerfile and verified everything works.

Reasoning: Personally I do not have Java installed on any of my systems. It is very handy to have access to `robot` but not have to rely on local java.  Once you build this robot container using command listed in the main README.md file, you can just run robot from anywhere on windows for example without having java installed. All prerequesites are in docker. I have verified and it works

Below you can see i run java to demonstrate java is not on my windows. But i run `robot` from container and everything works.

image

Also, I have updated how the container is built. As opposed to storing all the files required to build it, we instead now build in intermediate container. This intermediate container is downloading and building robot.jar.   After that we simply copy robot.jar to new container and discard all the build files. The result is drop from 1.2GB to 490M.  Still huge, but a lot less.

I hope this is useful to somebody.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,42 @@
-FROM maven:3-openjdk-11-slim
+# Stage 1: Build the application using Maven
+FROM maven:3-openjdk-11-slim AS build
 
+# Create a working directory and set permissions
 RUN useradd -m robot
 RUN mkdir -p /usr/src/app
 RUN chown robot /usr/src/app
 
+WORKDIR /usr/src/app
+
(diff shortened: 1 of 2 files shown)


gouttegd (approved):
Everything looks fine to me.

styk-tv:
@gouttegd - thank you
@jamesaoverton - done, thank you

PR #1223, opened 2024-11-11
Fix repair merge axiom annotations

matentzn:
For some reason, ROBOT repair --merge-axiom-annotations not only was restricted to annotation axioms (which is wrong), it also drops logical axioms with annotations as they are (1) first added to the heap of axioms that need to be processed, (2) dropped from the ontology because they are in that heap and finally (3) because of the wrong conditional we are fixing here, not added back.


matentzn:
@jamesaoverton to review this more easily, @gouttegd taught me a great trick: review the commits one by one. This way, you don't see the big normalisation diffs in the test cases, and you only have to review 2 lines of code and about 10 lines of test cases.

PR #1227, opened 2024-11-19
Update owl-diff dependency for stable ordering and to avoid large string creation

### LINKED ISSUE #1192: diff feature request: sort axioms in Added and Removed sections in Markdown output
Currently the Markdown output of `robot diff ... -f markdown` seems to randomly order the lines under "Added" or "Removed" for each changed entity (e.g., class).

I think the output would be much easier to read (and compare, when needed) if it had a canonical sort-order for those lines.  I'd prefer an order like:
* declaration axiom
* equivalency axiom(s)
* subclass axiom(s)
* annotation axioms, themselves sorted by
  * annotation property label
  * value

balhoff:
Resolves #1192. owl-diff now accepts a Writer or OutputStream so it can avoid creating a huge string. Also, the ordering of output should be consistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ## Fixed
+- Update owl-diff dependency for stable ordering and to avoid large string creation [#1227]
 - Improve disambiguation of properties in QuotedEntityChecker [#1226]
 - Skip "non-robot" columns in templates for the purposes of axiom annotations [#1216]
 - Add missing filter for deprecated in lowercase_definition check [#1220]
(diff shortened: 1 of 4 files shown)


jamesaoverton:
This seems straightforward to me, and it passes the tests. Ready for me to merge it @balhoff?

balhoff:
Thanks @jamesaoverton, I think it's ready.

jclerman:
With apologies for the late comment here and also for the pings: I am just testing this in my local, via my use of `robot` in `git diff` - and the improvement is dramatic.  Really nice to see, for example, the class-declaration as the top line when a class is created, etc.

Thanks @balhoff and @jamesaoverton!

Looking forward to having this in a `robot` release! Meantime I'm using the latest snapshot.

Issue #1222, opened 2024-11-08
Consider making the --output argument more universal (particularly robot export & query)


allenbaron:
Would it be possible to add `--output` / `-o` as an argument to robot export as an alias for the `--export` argument. Most of the other robot commands (report, extract, etc.) use `--output` so I invariably try it for robot export and forget that it is not there until I get the error (and then have to look it up).

A similar thing happens when I use robot query with the `--query` argument. I often specify the output path with the `-o` argument but it is only an accepted input in connection with the `--update` argument. It is unusual to me that the `--query` argument takes two strings, the path to the query and the path for the output.

It seems maybe originally the `-o` option was meant to apply only for ontology output files? The use of `-o` in robot report does break that pattern though. I imagine it would be possible to implement these changes without altering the current behavior of the functions. robot query would probably be the hardest to fix.

I would offer a fix myself but don't have the bandwidth at this time (and basically no experience with java). This does not need to be high priority since it is primarily an inconvenience.

Thanks!

balhoff:
@allenbaron I've also been tripped up sometimes by `--query` taking two arguments, but finally realized it's because you can specify any number of queries on the same command line, each with their own output. It's quite a bit more efficient this way vs. chaining multiple `query` commands. Maybe robot could be extra smart and pass along the internal triplestore it creates, in addition to the ontology, when `query` commands are chained.

jamesaoverton:
I reread the first part of the `query` doc, and I agree that it's not clear that you can use `--query SPARQL RESULT` multiple times. A PR to improve https://github.com/ontodev/robot/blob/master/docs/query.md would be appreciated.

tmprd:
This has confused me a few times due to the error message, and I always forget that the syntax is different. Even adding the -q doesn't work.
```shell
robot query --input "test.ttl" -q "test.sparql" -o "output.ttl"
Missing argument for option: q
Use the -vvv option to show the stack trace.
Use the --help option to see usage information.
usage: robot [command] [options] <arguments>
    --add-prefix <arg>      add prefix 'foo: http://bar' to the output
    --add-prefixes <arg>    add JSON-LD prefixes to the output
    --catalog <arg>         use catalog from provided file
 -h,--help                  print usage information
    --noprefixes            do not use default prefixes
 -p,--prefix <arg>          add a prefix 'foo: http://bar'
 -P,--prefixes <arg>        use prefixes from JSON-LD file
    --strict                use strict parsing when loading an ontology
[...]

PR #1233, opened 2024-12-12
Clarify that robot query accepts multiple --query options

allenbaron:
Requested in #1222.

This adds a brief comment and example, built from combining the previous two examples, to the docs for the `query` command to clarify that it accepts multiple `--query` options in a single invocation.

I also confirmed that the example works by executing it in the docs/examples dir.

How does the query command handle the `--format` option in this scenario. Does that need to be included in the docs as well?

### DIFF
--- a/docs/query.md
+++ b/docs/query.md
@@ -31,6 +31,12 @@ CONSTRUCT produces RDF data, if there are any results, defaulting to Turtle form
 
 This produces <a href="/examples/part_of.ttl" target="_blank">`part_of.ttl`</a>.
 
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Much appreciated!

`QueryCommand` iterates over the queries, handling each one separately with its own format: https://github.com/ontodev/robot/blob/master/robot-command/src/main/java/org/obolibrary/robot/QueryCommand.java#L399

The second paragraph says "The output format will be inferred from the output file extension, or you can use the --format option." so I think that's enough.

Issue #995, opened 2022-04-25
Discuss adding --clean-obo option to convert


beckyjackson:
Some ontologies contain axioms that will not properly convert to OBO, and the OBO writer sometimes fails with no good information on how to fix the issue - see https://github.com/OBOFoundry/COB/issues/168

We should consider adding a hack to `convert` that removes these bad axioms (e.g., `inverse` in subClass axioms). My main concern with this is how to handle is multiple annotations (label, definition, comment). From above discussion:
> Do we drop all of them, or do we just pick a random one to drop? That could result in some faulty outputs and spurious diffs.

cc @matentzn

jamesaoverton:
I agree that this is a good feature to add. It needs to be deterministic, and @matentzn's suggestion of first in lexical order, and INFO for dropped annotations, sounds good to me.

Are you volunteering to implement, or find someone to implement this @matentzn?

matentzn:
I will send my sensors out. :)

gouttegd:
Should that `--clean-obo` option also direct the removal of the “untranslated axioms” (which normally end up in the `owl-axioms` header tag)?

I would argue that it should, given the number of pipelines in the wild that contain variations of:

```sh
robot convert -i input.owl -f obo -o tmp.obo && grep -v ^owl-axioms tmp.obo > output.obo
```

Those untranslated axioms are clearly not wanted by a lot of people! And it’s a pretty low hanging fruit, since the OWLAPI already has an option to prevent them from being included in the output, all ROBOT has to do is to use that option.

matentzn:
Yes! The logger should probably Report which axioms are dropped by such an option, either with info or warning.

gouttegd:
OK, then should it also drop GCI axioms? This is [something that has been requested for CL](https://github.com/obophenotype/cell-ontology/issues/2856), on the rationale that many OBO parsers are too simplistic, do not expect the presence of GCIs, and therefore do not interpret them correctly.

I am inclined to say it should (or at least, that it should _allow_ to do it). If we add a “clean OBO“ output option, it should be suitable for existing use cases.

matentzn:
Formally ideally, there should be sepearte options for all subproblems

- "drop GCIs" fits best under `--axioms gci`
- `--drop-duplicate-labels`
- `--drop-duplicate-comments`
- `--axioms obo-format-incompatible`

And whatever else needs to go there, and then have a super option `--clean-obo` that sets the truth values of all of the above to `true`?

[...]

PR #1236, opened 2025-01-19
Add `--clean-obo` option to `convert`.

gouttegd:
Resolves [#995]

This PR adds a `--clean-obo` option to the `convert` command, as discussed in #995.

The option accepts a space-separated list of keywords that dictate how OBO output should be tweaked:

* `drop-extra-labels`: if an entity has more than one `rdfs:label` annotation, keep the first label only;
* `drop-extra-definitions`: likewise, but for `IAO:0000115` definitions;
* `drop-extra-comments`: likewise, but for `rdfs:comment` annotations;
* `merge-comments`: alternative to `drop-extra-comments`: merge `rdfs:comment` annotations into a single annotation, instead of keeping only the first one;
* `drop-untranslatable-axioms`: drop axioms that cannot be represented in the OBO format instead of putting them in a `owl-axioms` header tag;
* `drop-gci-axioms`: drop general concept inclusion axioms.

In addition, the option accepts the following shortcut keywords:

* `strict`: equivalent to `drop-extra-labels drop-extra-definitions drop-extra-comments`;
* `simple`: equivalent to `strict drop-untranslatable-axioms drop-gci-axioms`;
* `true`: equivalent to `simple`.

### DIFF
--- a/robot-core/src/main/java/org/obolibrary/robot/OntologyHelper.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/OntologyHelper.java
@@ -1829,7 +1829,19 @@ public class OntologyHelper {
   }
 
   /**
-   * Merges supernumerary annotations on all entities in the given ontology.
+   * Merges supernumerary annotations on all entities in the given ontology, using a single space
+   * character between each annotation value.
+   *
+   * @param ontology OWLOntology in which to merge supernumerary annotations
+   * @param properties set of annotation property IRIs to merge
+   */
+  public static void mergeExtraAnnotations(OWLOntology ontology, Set<IRI> properties) {
+    mergeExtraAnnotations(ontology, properties, " ");
+  }
+
+  /**
+   * Merges supernumerary annotations on all entities in the given ontology, using a custom
+   * separator between each annotation value.
    *
    * <p>This method iterates over all entities (excluding imported entities) of the given ontology;
    * if an entity is found to have more than one annotation for each of the indicated properties,
@@ -1837,8 +1849,10 @@ public class OntologyHelper {
    *
    * @param ontology OWLOntology in which to merge supernumerary annotations
    * @param properties set of annotation property IRIs to merge
+   * @param separator Separator string to insert between each annotation value (may be {@code null})
    */
-  public static void mergeExtraAnnotations(OWLOntology ontology, Set<IRI> properties) {
+  public static void mergeExtraAnnotations(
+      OWLOntology ontology, Set<IRI> properties, String separator) {
     Set<OWLAnnotationAssertionAxiom> axiomsToRemove = new HashSet<>();
     Set<OWLAnnotationAssertionAxiom> axiomsToAdd = new HashSet<>();
     OWLDataFactory factory = ontology.getOWLOntologyManager().getOWLDataFactory();
@@ -1865,8 +1879,8 @@ public class OntologyHelper {
           Set<OWLAnnotation> axiomAnnotations = new HashSet<>();
           boolean first = true;
           for (OWLAnnotationAssertionAxiom ax : sortAxioms(axioms)) {
-            if (!first) {
-              sb.append(' ');
+            if (!first && separator != null) {
+              sb.append(separator);
             } else {
               first = false;
             }
--- a/robot-core/src/main/java/org/obolibrary/robot/OntologyHelper.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/OntologyHelper.java
@@ -1780,4 +1780,146 @@ public class OntologyHelper {
       manager.addAxiom(ontology, cleanedAxiom);
     }
   }
+
+  /**
+   * Removes supernumerary annotations on all entities in the given ontology.
+   *
+   * <p>This method iterates over all entities (excluding imported entities) of the given ontology;
+   * if an entity is found to have more than one annotation for each of the indicated properties,
+   * all annotations beyond the first one are removed.
+   *
+   * @param ontology OWLOntology from which to remove the supernumary annotations
+   * @param properties set of annotation property IRIs to remove
+   */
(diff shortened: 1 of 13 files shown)


gouttegd:
Not sure what caused the Java 11 test to take more than 10 minutes (thereby triggering a failure)… Can’t reproduce here, on my machine the test suite runs fine with Java 11, Java 17, and Java 21 in about 2.5 minutes.

matentzn:
> Not sure what caused the Java 11 test to take more than 10 minutes (thereby triggering a failure)… Can’t reproduce here, on my machine the test suite runs fine with Java 11, Java 17, and Java 21 in about 2.5 minutes.



Whatever it was it was temporary

jamesaoverton:
> Not sure what caused the Java 11 test to take more than 10 minutes (thereby triggering a failure)… Can’t reproduce here, on my machine the test suite runs fine with Java 11, Java 17, and Java 21 in about 2.5 minutes.

I've also seen intermittent timeouts from one of the three Java versions, but haven't noticed a pattern, and it will be hell to debug. So I just re-run the failing one and it usually passes the second time. Not ideal.

gouttegd:
@jamesaoverton Tests added. :)

robot-core/src/main/java/org/obolibrary/robot/OntologyHelper.java
```diff
@@ -1780,4 +1780,146 @@ public static void removeAllAxiomAnnotations(OWLOntology ontology) {
...
+            logger.warn("Cannot merge non-literal <{}> annotations on <{}>", property, entity);
+            continue;
+          }
+          // Build the merged annotation axiom
+          StringBuilder sb = new StringBuilder();
+          Set<OWLAnnotation> axiomAnnotations = new HashSet<>();
+          boolean first = true;
+          for (OWLAnnotationAssertionAxiom ax : sortAxioms(axioms)) {
+            if (!first) {
+              sb.append(' ');
```
matentzn, line 1869: I would prefer a more distinct separator for multiple comments to avoid two merged comments concatenated to cause confusion.. I am not at all worried about verbosity here, so from `|` to "comment 1, comment 2" I would be fine with everything

matentzn: (perhaps make configurable?)

matentzn (commented):
Great code, nicely written docs. I went through most things top to bottom (not all the Java lines in extreme detail) and I find this very readable, nicely written code. I have a tiny opinionated suggestion that I have no concern abandoning.

robot-core/src/main/java/org/obolibrary/robot/OntologyHelper.java
gouttegd: No strong opinion about the separator character, though I do personally prefer a space over a pipe. (A pipe is fine when the reader is a machine, not when the reader is a human.)

Strongly opposed to making the separator character configurable. This would complicate both the command-line interface _and_ the underlying code for very little benefits.

People who would like this level of customisability can always craft their own scripts.

gouttegd: > Strongly opposed to making the separator character configurable. This would complicate both the command-line interface and the underlying code for very little benefits.

More correctly: I am not opposed to making that configurable at the level of the `mergeExtraAnnotations` method (it could just be one optional parameter, which defaults to whatever we decide). But strongly opposed to exposing that parameter at the level of the command-line, which would then always use the default separator (so, customisability would only be available to people using ROBOT as a library, not as a command-line tool).

matentzn: It is fine! I am thinking about cases where the comment does not end with a period, leading to a weird impression that there is one long sentence which does not make sense. If you think its not worth contemplating, I am fine as is

matentzn: I imagine, but might be wrong, that sometimes a comment is just a word, like "USELESS", "DEPRECATE" etc..

gouttegd: But, again: I simply have a mild preference for a space character, but I wouldn’t object to switch to something else. I strongly object only to making the separator configurable from the command line.

matentzn: Semicolon?

gouttegd: That would be even worse than a pipe in my opinion… What about all the comments that _correctly_ end with a period (the vast majority of the comments I have ever seen in ontologies – including in CL, which is currently the only ontology to have requested the “merge comments” feature)? You would make the vast majority of comments look weird (`First comment.; Second comment.`).

Seriously: if an ontology wants to merge comments upon exporting to OBO (which, incidentally, would not be the default behaviour, it has to be explicitly requested with `--clean-obo merge-comments`), then the onus is on that ontology to make sure its comments are suitably formatted for that purpose. And with a space character as the separator, all they have to do is to ensure that every comment ends with a period (which is already the case of most comments out there, from what I have seen).

matentzn: Ok; lets leave the space for now :) Not a big deal! Thanks for entertaining the discussion as always!

gouttegd: Separator string is now configurable _in code_ (third argument to `OntologyHelper.mergeExtraAnnotations()`). The default is still a single space character, and cannot be changed from the command-line of the `convert` command.

matentzn: Awesome, thank you!

matentzn (approved):
Super clean code, great docs, everything looks pretty much flawless!

Issue #1239, opened 2025-02-03
repair --merge-axiom-annotations is removing axioms it shouldn’t remove


gouttegd:
I believe there are two (maybe related) problems with the `repair --merge-axiom-annotations` feature.

They can both be illustrated with the following minimal example:

```
Ontology(<http://purl.obolibrary.org/obo/cl/imports/merged_import.owl>

# Declarations omitted for brevity...

# Object Property: <http://purl.obolibrary.org/obo/RO_0001025> (located in)
AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/RO_0001025> "located in")
ObjectPropertyDomain(Annotation(<http://purl.obolibrary.org/obo/IAO_0000116> "Some annotation.") <http://purl.obolibrary.org/obo/RO_0001025> <http://purl.obolibrary.org/obo/BFO_0000004>)
ObjectPropertyDomain(<http://purl.obolibrary.org/obo/RO_0001025> ObjectComplementOf(<http://purl.obolibrary.org/obo/BFO_0000006>))

# Class: <http://purl.obolibrary.org/obo/GO_0005319> (lipid transporter activity)
AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/GO_0005319> "lipid transporter activity")
SubClassOf(<http://purl.obolibrary.org/obo/GO_0005319> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/BFO_0000050> <http://purl.obolibrary.org/obo/GO_0006869>))
SubClassOf(Annotation(<http://www.geneontology.org/formats/oboInOwl#source> "GO_REF:0000090") <http://purl.obolibrary.org/obo/GO_0005319> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/BFO_0000050> <http://purl.obolibrary.org/obo/GO_0006869>))

)
```

Running `robot repair -i minimal.ofn --merge-axiom-annotations true -o repaired.ofn` yields

```
Ontology(<http://purl.obolibrary.org/obo/cl/imports/merged_import.owl>

# Declarations omitted for brevity...

# Object Property: <http://purl.obolibrary.org/obo/RO_0001025> (located in)
AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/RO_0001025> "located in")
[...]

PR #1240, opened 2025-02-06
Fix `repair --merge-axiom-annotations`

gouttegd:
Resolves [#1239]

The `RepairOperation::mergeAxiomAnnotations()` method only ever looks at annotated axioms. As a result, it misses the case where the ontology contains both an unannotated axiom and one (or more) annotated version(s) of that same axiom. For example:

```
SubClassOf(UBERON:1 UBERON:2)
SubClassOf(Annotation(rdfs:comment "A comment") UBERON:1 UBERON:2)
SubClassOf(Annotation(rdfs:comment "Another one") UBERON:1 UBERON:2)
```
    
The last two axioms would be merged into a single one carrying the two comments, but the first axiom would remain distinct:
    
```
SubClassOf(UBERON:1 UBERON:2)
SubClassOf(Annotation(rdfs:comment "A comment"
           Annotation(rdfs:comment "Another one")
           UBERON:1 UBERON:2)
```

This PR fixes the issue by rewriting the method to look at _all_ axioms, whether they are annotated or not.
    
Theoretically, this is slightly inefficient as axioms that are not annotated and that do not have any annotated duplicate (which probably represent the majority of axioms in most ontologies) would still be forcibly removed and then added back. However I do not think that a more elaborated approach would yield enough performance gain to be worth the increased complexity. For what it’s worth, on my machine and with Uberon, the rewritten method is about 20% slower than the original (incorrect) one (running in ~5 seconds instead of ~4 seconds).

Importantly, the test file for this feature had to be converted from RDF/XML to OFN in order to add an instance of an unannotated axiom alongside annotated duplicates of the same axiom -- something the RDF/XML format does not allow to represent).


matentzn (approved):
This implementation is not only more correct, but also 100 times more readable than my old implementation. Full approve!

Issue #1241, opened 2025-02-07
[annotate] Allow interpolation of version IRI within annotation values


gouttegd:
One of the standard operations performed by the ODK when preparing import modules is to insert a `dc:source` ontology annotation with the value of the ontology’s version IRI. This is done with `robot query --update` and the following SPARQL query:

```sparql
PREFIX dc:  <http://purl.org/dc/elements/1.1/>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>

INSERT { 
    ?ontology dc:source ?version_iri .
}

WHERE {
  ?ontology rdf:type owl:Ontology ;
        owl:versionIRI ?version_iri .
}
```

While the query itself is very simple, going through all the convoluted process to execute SPARQL queries with ROBOT (converting the ontology from the OWLAPI model to the Jena model, running the query on the Jena model, dumping the updated model in Turtle, then parsing the dumped Turtle back into a OWLAPI model) is a considerable waste of resources for a task that could be done with a handful of lines of Java code. In fact, that task is so resource-intensive that the ODK has to disable it for “large” import modules.

It would be very convenient if ROBOT allowed to perform that task without having to rely on SPARQL.

I propose that it should be possible to do something like this:

```sh
robot annotate --input my-ontology.owl \
               --enable-interpolation true \
               --link-annotation dc:source "%{version_iri}" \
               --output my-annotated-ontology.owl
```

That is, make it possible to use `%{version_iri}` or `%{ontology_iri}` as placeholders within the values of the various `*-annotation` options, to be replaced with the version IRI or the ontology IRI of the current ontology.

Thoughts?

matentzn:
I like the idea, but what is the thought behind  `--enable-interpolation true`? Isnt this implicit in the syntax of the interpolated value?
`

gouttegd:
The idea was that interpolation would not be enabled by default, so that, if people have been using `%{version_iri}` in `annotate` commands before, they could still do so without any unexpected replacement unless they explicitly enable that feature.

Admittedly this is a very unlikely possibility. If you think it’s so unlikely that it is not worth it to have a guard rail, and that we can always perform interpolation without hiding the feature behind an option, that would be fine with me.

PR #1242, opened 2025-02-07
Allow interpolation of placeholders in annotation values.

gouttegd:
Resolves [#1241]

This PR implements the proposition outlined in #1241. It adds the possibility of using the `annotate` command to inject ontology annotations with values derived from the ontology itself, by substituting placeholders (of the form `%{NAME}`) within the values provided on the command line.

For now, only two placeholders are supported (`%{ontology_iri}` and `%{version_iri}`). For that reason, the implementation is very basic, and is merely a couple of search-and-replace operations.


docs/annotate.md
```diff
@@ -31,6 +31,19 @@ Including at least the following annotations is recommended:
...
 
+If the `--interpolate` option is used, then some placeholders, of the form `%{NAME}`, can be used within the annotation values and will be automatically replaced by computed values. Currently supported placeholers are:
+
+  * `%{ontology_iri}`, replaced by the ontology's IRI;
+  * `%{version_iri}`, replaced by the ontology's version IRI.
+
+Example:
+
+    robot annotate --input fbcv-module.owl \
+      --interpolate \
```
matentzn, line 42: I dont think there is a precedence for using option flags in ROBOT, I am pretty sure the general strategy is to always use values (`--interpolate true`) however redundant.

robot-command/src/main/java/org/obolibrary/robot/AnnotateCommand.java
```diff
@@ -75,6 +75,7 @@ public AnnotateCommand() {
         "annotate-defined-by",
         true,
         "if true, annotate all entities in the ontology with the ontology IRI");
+    o.addOption("e", "interpolate", false, "enable interpolation within annotation values");
```
matentzn, line 78: I would personally prefer if the "interpolation" flag would be computed dynamically like so (pseudocode):

```
boolean interpolate = value.contains("{%");
```

This is a theoretically breaking change but I cannot imagine that anyone in the world of `robot annotate` has ever used the combination of `%{` (let alone `%{onotlogy_iri}` if you want to play it unnecessarily safe) in an annotation value.

A slightly less destructive option would be:

```
 boolean interpolate = CommandLineHelper.getBooleanValue(line, "interpolate", true, true);
```

So people could, in an emergency, switch of interpolation (but its used by default).

matentzn (commented):
Very nice code, good function (every ontology we have  using ODK would make use of this), two considerations for the design (using flags instead of normal options; the need for the "interpolate" flag), which @jamesaoverton will have to decide / look over.

docs/annotate.md
gouttegd: I’ve been wondering about that.

ROBOT is not always consistent on the issue, even within the context of a single command. For example, the `annotate` command has both the `--remove-annotation` flag, which does not accept a value, and the `--annotate-derived-from <true|false>` and `--annotate-defined-by <true|false>`.

Happy to follow the “general strategy”… assuming that there _is_ one, and that it is stated somewhere.

robot-command/src/main/java/org/obolibrary/robot/AnnotateCommand.java
gouttegd: > I would personally prefer if the "interpolation" flag would be computed dynamically > like so (pseudocode):
> 
> boolean interpolate = value.contains("%{");

That would mean that we search through the value twice: a first time just to decide whether we need to interpolate, and a second time to do the actual interpolation.

We might as well try to perform interpolation unconditionally then – and if the value happens _not_ to contain a placeholder, then nothing will be interpolated.

gouttegd: > I cannot imagine that anyone in the world of robot annotate has ever used the combination of %{

Well, I can imagine that.

I can imagine a scenario where someone hacked their own interpolation mechanism downstream of ROBOT. That is, they use ROBOT to inject `%{...}` placeholders into annotations, and then do a post-processing step using some custom scripts to replace those placeholders with the actual values they need. Such a custom pipeline would break if ROBOT suddenly does its own interpolation.

Far-fetched? Maybe, or maybe not.

[Mandatory XKCD](https://xkcd.com/1172)

docs/annotate.md
gouttegd: > ROBOT is not always consistent on the issue

I looked at that again, and I think I had exaggerated the inconsistency. It is clear that the vast majority of boolean-type options in ROBOT take a `true|false` argument – in fact only `annotate` and `template` also accept a few argument-less “flags”.

I’ll update the PR to make `--interpolate` take a `true|false` argument.

(Of course this point becomes moot if we decide instead that there is no need for an option and that interpolation should be unconditionally enabled.)

jamesaoverton: While there are some exceptions that we maintain for backwards compatibility (aka "mistakes"), I do want all new CLI options to take an argument. This has been a hard-won lesson: by including a boolean argument in the beginning, we leave room for additional options down the road. Thanks!

robot-command/src/main/java/org/obolibrary/robot/AnnotateCommand.java
jamesaoverton: Sorry @matentzn, I want to keep the explicit behaviour in the current version that requires `--interpolate true`. I hope that being explicit will (1) avoid surprises for existing workflows, and (2) encourage confused users to look up "interpolate" in the docs if they see strange glyphs in the command.

jamesaoverton (approved):

jamesaoverton:
Thanks for another nice improvement to ROBOT @gouttegd! I marked the two conversations as resolved. Ready for me to merge this?

PR #1246, opened 2025-03-05
Implemented #1245 (Support SPLIT=\\n)

### LINKED ISSUE #1245: Support SPLIT=\\n in template strings
SPLIT in template strings doesn't current support splitting on new lines.  I have a use case where I have multiple lines per cell in a CSV (as a list of synonyms for each term) and so I'd like to split on newlines.  Ideally I would like to allow SPLIT to support regexes but there isn't a clear, non-risky way of doing this without possibly breaking backwards compatibility.

After some discussion with @jamesaoverton on the OBO Slack channel we came to the conclusion that we would just explicitly support new line characters in SPLIT.

Note:  In the template string the slash character must be escaped so that it is preserved.  Thus `SPLIT=\\n` works but `SPLIT=\n` will not.

matthewhorridge:
Resolves #1245 

As specified in #1245 this PR add the ability to split values in cells using a new line character

- Refactored code to use a single method call for splitting values
- Support splitting on new line characters
- Added tests for splitting values


jamesaoverton:
Thanks for including tests and docs! I think this is a safe change.

matthewhorridge:
Brilliant!  Thanks for merging!

Issue #1212, opened 2024-08-13
New duplicate_exact_synonym test in robot report does not display values


allenbaron:
In robot report, the updated `duplicate_exact_synonym` test does not display the values for the `?entity ?property ?value` triples.

Examples:
Mondo OBO dashboard output - http://dashboard.obofoundry.org/dashboard/mondo/robot_report.html

DO using ROBOT 1.9.6
<img width="850" alt="Screenshot 2024-08-13 at 5 20 34 PM" src="https://github.com/user-attachments/assets/0f79cf0e-aa1f-468a-bfdb-bfc2857d9349">
_Note that this has a value for `duplicate_label_synonym` test._

Using the query at https://github.com/ontodev/robot/blob/master/robot-core/src/main/resources/report_queries/duplicate_exact_synonym.rq in robot query _does_ include the values. In the test I did with DO, it also returned an additional row -- one class had synonyms matching two other classes, so that class appeared twice in the output that contained values and only once in the robot report output (in each case only distinct outputs were retained).

matentzn:
@allenbaron thanks for detecting this! Any chance you might be willing to fix it? Maybe with the help of ChatGPT :D Sorry for asking but straaaaaapppped for ROBOT dev time!

allenbaron:
Okay, for 1.9.6... when I run the robot report with a profile limiting to just duplicate_exact_synonym I get 17 results and the values column is empty using doid-edit.owl (same as the full report), repeating this error.

**Command**
```
robot report -i doid-edit.owl --profile profile.txt -o DEL.tsv
```

Where profile.txt has this:
```
WARN	duplicate_exact_synonym
```

<details>

<summary>Output: 17 results, no values</summary>

```
Level	Rule Name	Subject	Property	Value
WARN	duplicate_exact_synonym	DOID:0050035	oboInOwl:hasExactSynonym	
WARN	duplicate_exact_synonym	DOID:14095	oboInOwl:hasExactSynonym	
WARN	duplicate_exact_synonym	DOID:0070475	oboInOwl:hasExactSynonym	
WARN	duplicate_exact_synonym	DOID:4464	oboInOwl:hasExactSynonym	
WARN	duplicate_exact_synonym	DOID:0110861	oboInOwl:hasExactSynonym	
WARN	duplicate_exact_synonym	DOID:898	oboInOwl:hasExactSynonym	
WARN	duplicate_exact_synonym	DOID:2477	oboInOwl:hasExactSynonym	
[...]

PR #1247, opened 2025-03-06
Include line separator when reading queries from Jar resources.

gouttegd:
Resolves [#1212]

When we read a default query string from a resource embedded within the Jar file, we need to make sure to explicitly include a line separator between each line (the line terminator is *not* included in the value returned by `BufferedReader#readLine()`). Otherwise the entire query is stored as a single line, which prevents its from being analyzed correctly by the SPARQL parser.


jamesaoverton:
This one is horrifying. I'm amazed it didn't cause a bigger problem. Thanks!

Issue #1208, opened 2024-06-27
reduce not removing axioms where one relationship is a subproperty of another between the same two classes


Clare72:
For example:
A part_of some B
A overlaps some B
part_of SubObjectPropertyOf overlaps

I would expect the `A overlaps some B` axiom to be removed by reduce, but it is not.

running `robot reduce -r ELK -i test.ofn -o result.ofn`
using `ROBOT version 1.9.6`
test file (`test.ofn`):

```
Prefix(:=<http://www.semanticweb.org/test/ontologies/2024/5/untitled-ontology-1023/>)
Prefix(owl:=<http://www.w3.org/2002/07/owl#>)
Prefix(rdf:=<http://www.w3.org/1999/02/22-rdf-syntax-ns#>)
Prefix(xml:=<http://www.w3.org/XML/1998/namespace>)
Prefix(xsd:=<http://www.w3.org/2001/XMLSchema#>)
Prefix(rdfs:=<http://www.w3.org/2000/01/rdf-schema#>)


Ontology(<http://www.semanticweb.org/test/ontologies/2024/5/untitled-ontology-1023>

Declaration(Class(<http://purl.obolibrary.org/obo/TEST_A>))
Declaration(Class(<http://purl.obolibrary.org/obo/TEST_B>))
Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/overlaps>))
Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/part_of>))

############################
#   Object Properties
############################

# Object Property: <http://purl.obolibrary.org/obo/overlaps> (overlaps)

AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/overlaps> "overlaps"@en)

# Object Property: <http://purl.obolibrary.org/obo/part_of> (part_of)

AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/part_of> "part_of"@en)
SubObjectPropertyOf(<http://purl.obolibrary.org/obo/part_of> <http://purl.obolibrary.org/obo/overlaps>)



############################
#   Classes
############################

# Class: <http://purl.obolibrary.org/obo/TEST_A> (A)

AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/TEST_A> "A"@en)
[...]

Issue #1014, opened 2022-06-20
Reduce operation with property chains


hkir-dev:
For the [given ontology](https://github.com/ontodev/robot/files/8942924/test_reduce.zip):

```owl
Prefix(:=<http://purl.obolibrary.org/obo/test.owl#>)
Prefix(owl:=<http://www.w3.org/2002/07/owl#>)
Prefix(rdf:=<http://www.w3.org/1999/02/22-rdf-syntax-ns#>)
Prefix(xml:=<http://www.w3.org/XML/1998/namespace>)
Prefix(xsd:=<http://www.w3.org/2001/XMLSchema#>)
Prefix(rdfs:=<http://www.w3.org/2000/01/rdf-schema#>)


Ontology(<http://purl.obolibrary.org/obo/test.owl>

Declaration(Class(<http://purl.obolibrary.org/obo/Kupffer_cell>))
Declaration(Class(<http://purl.obolibrary.org/obo/liver>))
Declaration(Class(<http://purl.obolibrary.org/obo/liver_lobule>))
Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/located_in>))
Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/part_of>))

############################
#   Classes
############################

# Class: <http://purl.obolibrary.org/obo/Kupffer_cell> (<http://purl.obolibrary.org/obo/Kupffer_cell>)

SubClassOf(<http://purl.obolibrary.org/obo/Kupffer_cell> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/located_in> <http://purl.obolibrary.org/obo/liver>))
SubClassOf(<http://purl.obolibrary.org/obo/Kupffer_cell> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/located_in> <http://purl.obolibrary.org/obo/liver_lobule>))

# Class: <http://purl.obolibrary.org/obo/liver_lobule> (<http://purl.obolibrary.org/obo/liver_lobule>)

SubClassOf(<http://purl.obolibrary.org/obo/liver_lobule> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/part_of> <http://purl.obolibrary.org/obo/liver>))


SubObjectPropertyOf(ObjectPropertyChain(<http://purl.obolibrary.org/obo/located_in> <http://purl.obolibrary.org/obo/part_of>) <http://purl.obolibrary.org/obo/located_in>)
)
```

[...]

Issue #16, opened 2015-04-30
Add op: remove-redundant-axioms


cmungall:
For discussion, cc @dosumis

## Motivation

For various reasons, the release version of an ontology can end up with redundant axioms; e.g.

 1. `A SubClassOf B`
 2. `B SubClassOf C`
 3. `A SubClassOf C`

Here 3 is redundant. Generally this should be pruned from the public release. There may be situations where this is not desired: for example, 3 may be adorned with a useful axiom annotation

## How it is used

As part of a release process, after reasoning and after #7

Note on GO-centric corner case: should never be applied to editors ontology as it may cause "ontology autophagy" when used as part of a cached link approach.

## Extensions

The notion of redundancy can be extended to the existential graph too, but this is more subtle, and my be best considered out of scope for this op, and a new op developed. The name `perform-transitive-reduction` or similar should be reserved for this task and @fbastian's code adopted.

## Spec

For simplicity we will only consider the graph formed by SubClassOf axioms between named classes. Equivalence axioms are ignored for the purposes of determining redundancy. Thus in the following 1 will be retained

```
1. A SubClassOf B
2. A EquivalentTo B and R some Z
```

The user should perform #7 `relax` step first to ensure a complete SubClassOf axiom graph is present.  Motivation: if we use reasoning to determine redundancy it complicates structural and logical requirements (TBD).

If the ontology contains an axiom `A SubClassOf Z`, and there exist a chain of two or more axioms `A SubClassOf B`, `B SubClassOf ...` ... `... SubClassOf Z`, then `A SubClassOf Z` is considered redundant.

Alternate reasoner based way of finding redundancy that does not require pre-reasoning: If the ontology contains an asserted axiom `A SubClassOf Z` and two inferences, `A SubClassOf B`, 'B SubClassOf Z` (as determined by the owlapi reasoner api, i.e. proper SubClassOf), then A-Z is redundant.

All redundant axioms are found, and then removed

TODO: spec equivalence between named classes corner case

fbastian:
Hi Chris, 

just wondering, why not using the OWLGraphManipulator class? OK, it might not be the most beautiful piece of code, and it uses a graph-walk approach, not a reasoner-based approach, but I think it respects pretty much all of your specs, or can be easily adapted to your needs.

It takes into account super object properties, indirect redundant relations, GCI axioms, and could easily targets SubClassOf axioms only. It is extensively unit tested, and has already been used a lot on Uberon.

cmungall:
[...]

PR #1248, opened 2025-03-06
Include SubObjectPropertyOf axioms when reducing.

gouttegd:
Resolves [#1014, #1208]

This PR amends the _Reduce_ code to include `SubObjectPropertyOf` and `SubPropertyChainOf` axioms in the graph used to evaluate redundant axioms.

By including `SubObjectPropertyOf` axioms, if we have something like

> SubClassOf(C, P some D)
> SubClassOf(C, Q some D)
> SubObjectPropertyOf(Q, P)

then the first `SubClassOf` axiom can be considered redundant and removed (#1208).

By including `SubPropertyChainOf` axioms, if we have

> SubClassOf(C, P some D)
> SubClassOf(C, P some E)
> SubClassOf(D, Q some E)
> SubObjectPropertyOf(ObjectPropertyChain(P Q) P)

then the second `SubClassOf` axiom (_SubClassOf(C, P some E)_) can be considered redundant and removed (#1014).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add missing filter for deprecated in lowercase_definition check [#1220]
 - Bug was fixed that caused logical axioms with axiom annotations not to be processed correctly when merging axiom annotations [#1223]
 - Correctly merge unannotated and annotated duplicated axioms [#1239]
+- Subproperties being ignored when evaluating redundancy in `reduce` [#1014], [#1208]
 
 ## [1.9.7] - 2024-10-30
 
--- a/docs/examples/reduced-redundant-over-subproperties.ofn
+++ b/docs/examples/reduced-redundant-over-subproperties.ofn
@@ -1,0 +1,40 @@
+Prefix(:=<http://purl.obolibrary.org/obo/test.owl#>)
+Prefix(owl:=<http://www.w3.org/2002/07/owl#>)
+Prefix(rdf:=<http://www.w3.org/1999/02/22-rdf-syntax-ns#>)
+Prefix(xml:=<http://www.w3.org/XML/1998/namespace>)
+Prefix(xsd:=<http://www.w3.org/2001/XMLSchema#>)
+Prefix(rdfs:=<http://www.w3.org/2000/01/rdf-schema#>)
+
+
+Ontology(<http://purl.obolibrary.org/obo/test.owl>
+
+Declaration(Class(<http://purl.obolibrary.org/obo/Kupffer_cell>))
+Declaration(Class(<http://purl.obolibrary.org/obo/liver>))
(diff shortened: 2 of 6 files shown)


gouttegd:
@cmungall Asking for your review since you write the original code for the Reduce operation (and especially commit cdbb8043c4a1f662c6c3f4cd7e657eef31ebfe51, which restricted the types of axioms considered for evaluating redundancy).

@jamesaoverton Should the behaviour be configurable by a new command line option to `reduce` (something like `--include-subproperty-axioms true` or similar)?

matentzn (commented):
This change is very consequential - and I am actually a bit surprised that this was not already happening. 

I would, however, suggest to add a test for this change, as it has the potential for huge consequences on a lot of OBO (esp. ODK-managed) ontologies..

gouttegd:
@matentzn I plan to add tests once I have confirmation the feature (which is more of a bug fix in my opinion, as I think `reduce` should always do that -- which is also why I didn't include an CLI switch) is wanted.

matentzn:
I personally agree that this is a bugfix, but best get @balhoff and @cmungall opinion on this as well.

gouttegd:
> I personally agree that this is a bugfix

For what it’s worth, the people who reported the #1208 issue (`reduce` ignoring _SubPropertyOf_ axioms) and the #1014 issue (`reduce` ignore property chains) both thought it was a bug. :)

cmungall:
Why not just add a `reduce` configuration, and preserve the current default behavior?

Redundancy is not straightforward, see:

- https://github.com/INCATools/relation-graph/issues/63

Apologies for any confusion caused. When I originally proposed this:

- #16         

I suggested separate commands from removing redundancy (sensu OWL entailment) and graph transitive reduction, we probably should have stuck to that (with graph operations happening in a framework independent of the owlapi / owl reasoning), as these are really different concerns and use cases. But I think I may have overloaded them.

balhoff (changes requested):
I think it sounds good to have this feature, but it should be added as a non-default CLI option. We have been relying on the previous behavior, at least in GO, to materialize more standard relations from highly specific ones (e.g. 'part of' from 'bounding layer of'). My understanding is that this would stop working after this change.

gouttegd:
@balhoff Option tentatively added. :) I defer to @jamesaoverton for the decision as to whether it should be enabled or disabled by default (for now it’s disabled, it’s one line to change in `ReduceOperation#getDefaultOptions()` if we want to make it enabled by default).

jamesaoverton:
I agree with @balhoff: we should stick to the old behaviour by default, to maintain backwards compatibility, even if we agree that the old behaviour was buggy. (Sorry, but backwards compatibility is very important to the ROBOT project.)

Thanks very much!

gouttegd:
@jamesaoverton Fine with me. :) But for the record, I then plan to enable the option by default in ODK workflows (users would still have the possibility to disable it in their ODK configuration).

cmungall:
I really like this general pattern (robot is rigorous about backward
compatibility, ODK may override)

PR #1249, opened 2025-03-07
Add support for predicate-object pairs to ROBOT template

jamesaoverton:
Replaces #1205, building on work by @pk-mitre.

Add support for property-object pairs in a single cell of a ROBOT template:

`AP` annotation property-object pairs: If the template string is `AP` then ROBOT will look in each cell of that column for predicate-object pairs, with a syntax similar to Turtle, such as `ex:p1 'A'` or `ex:p2 '1234'^^xsd:integer`. Using `AP SPLIT=|` we can specify a list of annotations: `ex:p1 'A'|ex:p2 "B"@en|ex:p3 '3'^^xsd:integer|ex:p4 ex:curie|ex:p5 <http://ex.com>`.


jamesaoverton:
@pk-mitre and @dlutz2 I've made some changes to make this more general. Can you please look at this and give me your feedback?

dlutz2:
@jamesaoverton These look good. Covering all object types very useful and we need the annotation value language tag, glad you added that.

docs/examples/multi_annotations.tsv
```diff
@@ -0,0 +1,2 @@
+Term	Term ID	Annotations
+ACME Communications INC	acme:s-1	"acme:p-1 " Value ZZ"|acme:p-2 "1234"^^xsd:integer|acme:p-3 'FOO	BAR,BAZ,BAT'"
```
matentzn, line 2: Just noting here that I didnt know `"acme:p-1 " Value ZZ"..."` would work without escaping the inner `"` symbols!

matentzn (approved):
Very cool feature! Implementation looks good as well.

PR #1253, opened 2025-05-10
Fix typo in `--clean-obo` option values.

gouttegd:

This PR fixes an unfortunate typo that has been introduced in #1236. The option to forcibly remove axioms that cannot be represented in OBO format is supposed to be `--clean-obo drop-untranslatable-axioms` (that’s what the documentation says), but the code that handles the `--clean-obo` values expects `drop-unstranslatable-axioms` -- note the added `s`.

Sorry for not having caught that sooner. It’d be nice if it could be fixed before the upcoming 1.10 release, so that the `drop-unstranslatable-axioms` variant is never actually used by anyone.


jamesaoverton (approved):
Good to catch that!

Issue #1254, opened 2025-05-15
Intermittent GitHub Action timeouts on ExtractOperation unit tests


jamesaoverton:
I've seen a couple cases where the automated tests fail after our 10 minute timeout. It always hangs in the same place: `[INFO] Running org.obolibrary.robot.ExtractOperationTest`:

https://github.com/ontodev/robot/actions/runs/14932625035/job/41952471664#step:5:12875

Re-running the GitHub Action has fixed it. It's just annoying. I'm not seeing this locally, so it will be very difficult to debug. I'm going to leave it for now.

My best guess is that we run out of free memory, in which case we could reduce the size of the test files. But I really don't know.

jamesaoverton:
It's not just Java 11. This one failed under Java 15 at the same spot: https://github.com/ontodev/robot/actions/runs/15072913419/job/42373478449

jamesaoverton:
And I don't think it's a memory issue anymore, because the largest OWL files for unit tests are just over 1MB. Hmm...

balhoff:
@jamesaoverton do you have an idea which ExtractOperation test? I have a guess that it might be the part that interacts with relation-graph, which makes use of an asynchronous queue. I could try updating relation-graph or seeing if any bugs were fixed in its dependency.

jamesaoverton:
@balhoff Sorry, I don't know which test is causing the problem. I haven't seen this happen locally, and the GitHub Action log doesn't get into that detail. We could add print statements to all the `ExtractOperationTest`s, and wait for it to fail, I guess. Or is there a way to see the `robot-core/target/surefire-reports/org.obolibrary.robot.ExtractOperationTest.txt` file inside the GitHub Action?

balhoff:
@jamesaoverton I released a relation-graph update which does include a newer version of the parallel processing framework it uses. Maybe we can merge #1255 and see how subsequent PRs go?

PR #1255, opened 2025-05-17
Update relation-graph version

balhoff:
This version of relation-graph only updates its dependencies, including the parallel processing framework (zio). I'm interested to see if that has an effect on #1254.


balhoff:
No test hanging at least in these runs.

Issue #1268, opened 2025-10-23
OBO Format does support prefixes now, but is skipped


matentzn:
See https://github.com/ontodev/robot/blob/c77d8628cf1197517668c425adeb5864cb919c22/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java#L232

You cant inject a prefix into a OBO format serialisation pipeline (where OBO format is the last step), but you can inject it into some other format (like ofn), save it, and then convert to OBO in a separate pipeline. 

I guess this could be an OWL API issue, falsely claiming OBO is not a `isPrefixOWLOntologyFormat()`.

PR #1270, opened 2025-10-27
Partially fix prefix injection when writing to OBO.

gouttegd:
Partially resolves [#1268]

When writing to OBO, prefixes added with `--add-prefix` are ignored. This is because until recently, the OBODocumentFormat was not a prefix-supporting format, and so there is nothing in the OWLAPI to automatically take care of providing the prefixes the document may contain into the OWL2OBO translator.

This PR fixes the issue at least for the case where the `--clean-obo` option is used. In that case, we are instantiating the OWL2OBO translator ourselves, so we can give it access to the prefixes from the document.

For the case where `--clean-obo` is not used, the fix needs to happen in the OWLAPI. In that case, we are relying on the ontology manager's save methods, and we have no way to reach the OWL2OBO translator that is, in fine, used by those methods when writing to OBO. (It would be possible to always bypass the manager when writing to OBO, regardless of the `--clean-obo` option, but that would be a more invasive fix).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add `--input-format` global option [#1038]
+
 ### Fixed
+- Inherit OWLDocumentFormat when performing SPARQL update [#1267]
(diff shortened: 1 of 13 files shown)


jamesaoverton:
Thanks!

PR #1281, opened 2026-04-20
Load gz files without catalog

### LINKED ISSUE #1280: Fix catalog file existence check when loading .gz ontology files
ROBOT fails with `FILE DOES NOT EXIST ERROR File does not exist: /work/mirror/catalog-v001.xml` when loading a `.gz` ontology file in a directory without a `catalog-v001.xml`. Non-gz files work fine in the same scenario.

## Context

In `IOHelper.java`, the method `loadOntology(File ontologyFile, File catalogFile, String inputFormat)` handles catalog file detection differently for `.gz` vs non-gz files.

For non-gz files (line ~482), the catalog is only used if `catalogFile != null && catalogFile.isFile()`. For `.gz` files (lines ~486-491), the condition only checks `catalogFile == null`, which is never true because `guessCatalogFile()` always returns a `File` object (pointing to `catalog-v001.xml` in the input file's directory). So the non-existent catalog path is always passed to `loadCompressedOntology`, which then throws. IMO loading a `.gz` file without a `catalog-v001.xml` present should work the same as loading a non-gz file without one: skip catalog-based IRI mapping and proceed normally.

I will open a PR to change the `.gz` branch to use the same guard as the non-gz branch:

```java
if (ontologyFile.getPath().endsWith(".gz")) {
  if (catalogFile != null && catalogFile.isFile()) {
    return loadCompressedOntology(ontologyFile, catalogFile.getAbsolutePath(), inputFormat);
  } else {
    return loadCompressedOntology(ontologyFile, null, inputFormat);
  }
}
```

`robot-core/src/main/java/org/obolibrary/robot/IOHelper.java`, around line 486.

matentzn:
Resolves #1280 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix problem with catalog file for compressed ontologies [#1281]
+
 ## [1.9.10] - 2026-02-18
 
 ### Fixed
--- a/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
+++ b/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java
@@ -288,7 +288,7 @@ public class IOHelper {
    * file for a catalog file.
    *
    * @param ontologyFile the
-   * @return the guessed catalog File; may not exist!
+   * @return the guessed catalog File or null if it does not exist
    */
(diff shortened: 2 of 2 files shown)


jamesaoverton:
Your fix is good, but I was annoyed that the debug logging message still referred to the non-existent catalog file. So I reworked `guessCatalogFile` to return null if the file does not exist, and removed now redundant checks for that case.

matentzn:
Awesome, thanks @jamesaoverton

PR #1285, opened 2026-06-22
Preserve catalog path in metadata for multi-ontology-input commands

### LINKED ISSUE #1250: catalog path is not put into command state when merge operation is first
This took me a while to figure out. In a chain of commands like this:

```
robot --catalog catalog-v001.xml convert -i go-edit-filtered-xrefs.ofn  merge --collapse-import-closure false -i imports/go_taxon_constraints.owl -i imports/reactome_xrefs_import.owl -i ../taxon_constraints/present_in_taxon.ofn -i imports/go-catalytic-activities-participants.owl remove --term 'IAO:0000116' --axioms annotation --signature true reason -r ELK -e asserted-only --exclude-duplicate-axioms true --exclude-tautologies structural relax reduce query --update ../sparql/insert-goterm-namespace.ru annotate -V http://purl.obolibrary.org/obo/go/releases/`date +%Y-%m-%d`/go.owl --annotation owl:versionInfo `date +%Y-%m-%d` verify --queries  ../sparql/equivalent-classes-violation.sparql  ../sparql/trailing-whitespace-violation.sparql  ../sparql/owldef-self-reference-violation.sparql  ../sparql/synonym-label-match-violation.sparql  ../sparql/replacedby-obsolete-violation.sparql  ../sparql/replacedby-namespace-violation.sparql  ../sparql/duplicate-exact-synonym-violation.sparql  ../sparql/duplicate-synonym-violation.sparql  ../sparql/non-IRI-value-violation.sparql  ../sparql/non-anyURI-value-violation.sparql  ../sparql/obsolete-definition-violation.sparql  ../sparql/definition-constraints-violation.sparql  ../sparql/one-to-one-xrefs-by-subject-violation.sparql  ../sparql/one-to-one-xrefs-by-value-violation.sparql  ../sparql/xref-syntax-violation.sparql ../sparql/multiple-class-links-violation.sparql ../sparql/obsolete-reference-violation.sparql ../sparql/missing_superclass-violation.sparql ../sparql/different-namespace-violation.sparql -O reports
```

When the `query` command runs an `update`, it has to parse the resulting ontology again. It looks in the `CommandState` for a catalog that may have been used at the beginning, so that it can properly parse the ontology. However, `merge` does not put the catalog path into the state. Commands that expect a single ontology use [updateInputOntology](https://github.com/ontodev/robot/blob/2bc146f5a9a521c061c6f55cc282ed4b5e58a687/robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java#L537) to get the current state and ontology; in this case if there is a catalog file the path is stored in the state.

But `merge` uses [getInputOntologies](https://github.com/ontodev/robot/blob/2bc146f5a9a521c061c6f55cc282ed4b5e58a687/robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java#L509), which returns a list of ontologies rather than an updated state.

I've been trying to decide on the best way to store the catalog path in this case, but wanted to go ahead and make the issue so it doesn't get lost. I'm only just now realizing that this command has been reloading imports off the web instead of locally partway through.

This problem may exist for other commands as well; I haven't checked.

balhoff:

Fixes #1250.


robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java
```diff
@@ -1114,28 +1202,68 @@ public static List<OWLOntology> getInputOntologies(IOHelper ioHelper, CommandLin
...
-    List<String> inputOntologyPaths = getOptionalValues(line, "input");
-    for (String inputOntologyPath : inputOntologyPaths) {
-      inputOntologies.add(ioHelper.loadOntology(inputOntologyPath, catalogPath, inputFormat));
-    }
-    // Check for input IRIs
-    List<String> inputOntologyIRIs = getOptionalValues(line, "input-iri");
-    for (String inputOntologyIRI : inputOntologyIRIs) {
-      inputOntologies.add(
-          ioHelper.loadOntology(IRI.create(inputOntologyIRI), catalogPath, inputFormat));
+    return getInputOntologiesWithMetadata(ioHelper, line, catalogPath).getOntologies();
```
matentzn, line 1205: Note this way you change the meaning of `catalogPath == null` from "dont use catalog" to "use default". I dont think this will pass the @jamesaoverton QC guard.

robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java
```diff
@@ -1114,28 +1202,68 @@ public static List<OWLOntology> getInputOntologies(IOHelper ioHelper, CommandLin
...
+    /** Path to the first local input ontology, if the first input was local. */
+    private String primaryOntologyPath;
+
+    /**
+     * Create a result container.
+     *
+     * @param catalogPath path to the explicit catalog, or null
+     */
+    InputOntologies(String catalogPath) {
+      this.catalogPath = catalogPath;
```
matentzn, line 1225: At least if I read the code correctly, you dont need the catalogPath in the object, and least not right now. The methods that are calling the class object already have the catalog available. just saying. :P

robot-command/src/main/java/org/obolibrary/robot/CommandLineHelper.java
```diff
@@ -1080,23 +1118,73 @@ private static File[] getFilesByPattern(String pattern) throws IllegalArgumentEx
...
     String pattern = getOptionalValue(line, "inputs");
     if (pattern != null) {
+      File catalogFile = null;
+      if (catalogPath != null) {
+        catalogFile = new File(catalogPath);
+      }
       for (File inputOntologyFile : getFilesByPattern(pattern)) {
-        inputOntologies.add(ioHelper.loadOntology(inputOntologyFile, true, inputFormat));
+        OWLOntology ontology;
+        if (catalogFile != null) {
```
matentzn, line 1182: Somehow this if else seems like it could be refactored to fit both 1160ff and this instance.

